PR #38, opened 2024-08-06
Updates links to public documentation

pranavm-nvidia:


tripy/README.md
```diff
@@ -26,24 +26,27 @@ If you want to build from source, please follow the instructions in [CONTRIBUTIN
 
 ## Quickstart
 
-In lazy mode evalulation, Tripy defers computation until it is actually needed:
+In eager mode, Tripy works just like you'd expect:
```
jhalakpatel, line 29: Is lazy evaluation now just an implementation detail and we claim we just support `eager` mode?

pranavm-nvidia: Yes, I think eager evaluation is the behavior we care about, i.e. you can just do `print(a + b)` without requiring a compilation step first. That we lazily evaluate doesn't affect that behavior.

parthchadha (approved):

PR #51, opened 2024-08-07
Update workspace folder for MLIR-TensorRT container

jhalakpatel:


mlir-tensorrt/.devcontainer/devcontainer.json
```diff
@@ -23,7 +23,7 @@
 		"gpu": "optional"
 	},
 	"remoteUser": "nvidia",
-	"workspaceMount": "source=${localWorkspaceFolder}/..,target=/workspaces/TensorRT-Incubator,type=bind,consistency=cached",
```
jhalakpatel, line 26: Looks like we can not as we want `git` access from within the container.
```
nvidia ➜ /workspaces/mlir-tensorrt $ git status .
fatal: not a git repository (or any parent up to mount point /workspaces)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

is there any benefit to having access to the `tripy` folder in the MLIR-TensorRT dev container workflow? If yes, we can close the PR.

jhalakpatel: I believe it would be straightforward for someone developing MLIR-TensorRT to test their changes with Tripy in Tripy container by installing MLIR-TensorRT wheels instead.

christopherbate: > Looks like we can not as we want git access from within the container.

right

> is there any benefit to having access to the tripy folder in the MLIR-TensorRT dev container workflow?

Yes, for TriPy development

> I believe it would be straightforward for someone developing MLIR-TensorRT to test their changes with Tripy in Tripy container by installing MLIR-TensorRT wheels instead.

That's a very awkward development workflow, I wouldn't recommend that.

jhalakpatel:
Looks like we need to have top-level workspace folder to be a git project. Closing this PR.

PR #61, opened 2024-08-07
Updates MLIR-TRT packages to depend on TensorRT 10

pranavm-nvidia:
- Updates MLIR-TRT and Tripy packages to depend on TRT 10.X

- Updates MLIR-TRT to include major version when loading TRT. The libraries that ship with the wheels always include the major version and do not provide a `libvninfer.so` symlink

- Updates MLIR-TRT to load the TRT Python package when the Python bindings are imported. This ensures that the correct version of the libraries will be picked up by the loader later on.

### DIFF
--- a/mlir-tensorrt/python/mlir_tensorrt_compiler/pyproject.toml
+++ b/mlir-tensorrt/python/mlir_tensorrt_compiler/pyproject.toml
@@ -26,5 +26,6 @@ dependencies = [
     "pybind11>=2.8.0, <=2.10.3",
     "PyYAML>= 5.3.1, <=6.0.1",
     "dataclasses>=0.6, <=0.8",
+    "tensorrt~=10.0",
 ]
 dynamic = ["version"]
--- a/mlir-tensorrt/python/mlir_tensorrt_runtime/pyproject.toml
+++ b/mlir-tensorrt/python/mlir_tensorrt_runtime/pyproject.toml
@@ -21,4 +21,7 @@ classifiers = [
     "Operating System :: POSIX :: Linux",
     "Private :: Do Not Upload"
 ]
+dependencies = [
+    "tensorrt~=10.0",
+]
 dynamic = ["version"]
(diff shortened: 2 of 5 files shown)


tripy/pyproject.toml
```diff
@@ -7,6 +7,7 @@ readme = "README.md"
 requires-python = ">= 3.9"
 license = {text = "Apache 2.0"}
 dependencies = [
+  "tensorrt~=10.0",
```
parthchadha, line 10: Why does it have to be a dependency of both `mlir-tensorrt` and `tripy`?

pranavm-nvidia: We are now using it directly in `plugin.py` to figure out the major version

pranavm-nvidia:
I just realized I won't be able to merge the Tripy changes until after MLIR-TRT packages have been pushed. I'll split this into two PRs.

mlir-tensorrt/python/mlir_tensorrt_compiler/pyproject.toml
```diff
@@ -26,5 +26,6 @@ dependencies = [
     "pybind11>=2.8.0, <=2.10.3",
     "PyYAML>= 5.3.1, <=6.0.1",
     "dataclasses>=0.6, <=0.8",
+    "tensorrt~=10.0",
```
christopherbate, line 29: This is not true. You can build the project to target TensorRT 8.6. We shouldn't bake `10.0` into the requirements like this.

mlir-tensorrt/python/mlir_tensorrt_runtime/pyproject.toml
```diff
@@ -21,4 +21,7 @@ classifiers = [
     "Operating System :: POSIX :: Linux",
     "Private :: Do Not Upload"
 ]
+dependencies = [
+    "tensorrt~=10.0",
```
christopherbate, line 25: Same here

mlir-tensorrt/python/mlir_tensorrt_compiler/pyproject.toml
pranavm-nvidia: @christopherbate should we specify any version range or leave it completely open? Maybe `>= 8.6`?

pranavm-nvidia:
Chris's PR subsumes mine, so closing this.

PR #62, opened 2024-08-07
Reduces container size, fixes Tripy dependencies, bumps version

pranavm-nvidia:
- Greatly reduces the size of the development container by using a smaller base image and cutting out many unnecessary packages.
    
- Adds TRT as a direct dependency of Tripy since in `plugin.py`, we need to load it to determine the major version of `libnvinfer_plugin`.
    
- Removes CHANGELOG as we can automatically generate release notes via GitHub releases.
    
- Removes JAX as a test dependency and the tests that use it. We were previously using it to test interoperability using DLPack, but we already test that  with torch/cupy/numpy. Adding JAX provides no additional coverage and unnecessarily bloats the container.

- Bumps version to 0.0.2

### DIFF
--- a/tripy/Dockerfile
+++ b/tripy/Dockerfile
@@ -1,6 +1,6 @@
-FROM nvcr.io/nvidia/cuda:12.2.2-devel-ubuntu22.04
+FROM ubuntu:22.04
 
-LABEL org.opencontainers.image.description Tripy development container
+LABEL org.opencontainers.image.description="Tripy development container"
 
 WORKDIR /tripy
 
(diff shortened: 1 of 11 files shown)


parthchadha (approved):

tripy/Dockerfile
```diff
@@ -22,14 +22,6 @@ RUN groupadd -r -f -g ${gid} trtuser && \
 
 # Copy your .lldbinit file into the home directory of the root user
 COPY .lldbinit /root/
-
-# Install the recommended version of TensorRT for development.
-RUN cd /usr/lib/ && \
-    wget https://developer.nvidia.com/downloads/compute/machine-learning/tensorrt/10.1.0/tars/TensorRT-10.1.0.27.Linux.x86_64-gnu.cuda-12.4.tar.gz && \
-    tar -xvzf TensorRT-10.1.0.27.Linux.x86_64-gnu.cuda-12.4.tar.gz && \
-    rm TensorRT-10.1.0.27.Linux.x86_64-gnu.cuda-12.4.tar.gz
-ENV LD_LIBRARY_PATH=/usr/lib/TensorRT-10.1.0.27/lib/:$LD_LIBRARY_PATH
```
pranavm-nvidia, line 31: Cannot merge this until MLIR-TRT packages have been pushed

tripy/pyproject.toml
```diff
@@ -7,6 +7,7 @@ readme = "README.md"
 requires-python = ">= 3.9"
 license = {text = "Apache 2.0"}
 dependencies = [
+  "tensorrt~=10.0",
```
pranavm-nvidia, line 10: Need to decide if we want to pin to 10.1 or allow any 10.X.

pranavm-nvidia: Resolving this - I think it's fine to keep it permissive for now. We can pin to a specific version later if we need to.

jhalakpatel (approved):

parthchadha (approved):

PR #66, opened 2024-08-08
Array does not allocate buffer for list data

yizhuoz004:
When we use list to initialize a Tensor, `Array` will not allocate memref. The list will be stored and directly translated to MLIR during evaluation.

+: No more casting when dtype is not supported by memref.
-: Add a separate condition `Array.has_memref`

On 2nd commit:
Array only accepts an object with memory allocated (such as external lib's array). When using a list to initialize a Tensor, the list will be directly passed to `Storage` op, and then lowered into MLIR via `ConstantOp`.

### DIFF
--- a/tripy/tripy/common/array.py
+++ b/tripy/tripy/common/array.py
@@ -161,6 +161,7 @@ class Array:
 
         self.device = device
 
+        self.has_memref = not isinstance(data, (Sequence, bool, int, float))
         if data is None:
             if dtype is None:
                 raise_error("Datatype must be provided when data is None.", [])
@@ -169,7 +170,7 @@ class Array:
             self.dtype = dtype
             self.shape = shape
         else:
-            if isinstance(data, (List, bool, int, float, tuple)):
+            if not self.has_memref:
                 # shape can still be inferred for empty data, but dtype cannot be
                 if has_no_contents(data) and dtype is None:
                     raise_error("Datatype must be provided for empty data (i.e., not containing any scalars).")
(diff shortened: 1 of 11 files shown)


jhalakpatel:
`The list will be stored and directly translated to MLIR during evaluation.`
You are adding conversion logic to memref during evaluation instead of during initialization. This might have some perf impact as `convert_list_to_array` will pack list data into a byte buffer.

jhalakpatel:
@yizhuoz004 How are we planning to handle a case where we create a list and request a type which could result in narrowing conversion since we no longer insert a cast operation?

tripy/tripy/common/array.py
```diff
@@ -169,7 +170,7 @@ def __init__(
             self.dtype = dtype
             self.shape = shape
         else:
-            if isinstance(data, (List, bool, int, float, tuple)):
+            if not self.has_memref:
```
pranavm-nvidia, line 173: Is there any way we can simplify all the branching here? Maybe don't create an `Array` at all if we're using a Python list?

tripy/tripy/flat_ir/ops/constant.py
```diff
@@ -44,34 +43,46 @@ def to_mlir(self, operands):
...
-        )
+            def to_attrs(data):
+                if isinstance(data, numbers.Number):
+                    return [mlir_utils.get_mlir_scalar_attr(out_dtype, data)]
+                attrs = []
+                for element in data:
+                    attrs.extend(to_attrs(element))
+                return attrs
+
+            attr = ir.DenseElementsAttr.get(attrs=to_attrs(self.data.data()), type=self.outputs[0].to_mlir())
```
yizhuoz004, line 86: @jhalakpatel This is the logic that converts a python list to MLIR constant:

- There is no cast because we are creating the MLIR attr of target dtype directly.
- `convert_list_to_array` is not called here, MLIR attr accpets python numbers directly, there should be no perf impact.

tripy/tripy/common/array.py
yizhuoz004: Yes I've explored that option too, I was stuck at the return type of `Tensor.eval()`, which is always an `Array`.  I will continue on this approach today.

PR #69, opened 2024-08-08
Add labeling workflow

pranavm-nvidia:


parthchadha (approved):

yizhuoz004 (approved):

label_config.yml
```diff
@@ -0,0 +1,7 @@
+tripy:
```
parthchadha, line 1: Why do we have two .yml files?

pranavm-nvidia: I was experimenting with something. Removed this file now.

PR #71, opened 2024-08-08
Added function `create_memref_view_from_dlpack`

markkraay:
This function greatly simplifies Tripy's Array implementation. We want to be able to handle memref creation from all types that implement the `__dlpack__()` interface rather than the limited set we currently support. This function should allow us to achieve this. The corresponding Tripy changes are #72.

PR #72, opened 2024-08-08
Use `create_memref_view_from_dlpack` within Tripy

markkraay:
Relies on #71. Simplifies our Array class by using `create_memref_view_from_dlpack`.

### DIFF
--- a/tripy/tests/integration/test_functional.py
+++ b/tripy/tests/integration/test_functional.py
@@ -166,11 +166,11 @@ class TestFunctional:
 
             tripy_linear = tp.Linear(2, 3)
             if kind == "gpu":
-                tripy_linear.weight = tp.Parameter(tp.Tensor(torch_linear.weight.to("cuda"), device=tp.device(kind)))
-                tripy_linear.bias = tp.Parameter(tp.Tensor(torch_linear.bias.to("cuda"), device=tp.device(kind)))
+                tripy_linear.weight = tp.Parameter(tp.Tensor(torch_linear.weight.detach().to("cuda"), device=tp.device(kind)))
+                tripy_linear.bias = tp.Parameter(tp.Tensor(torch_linear.bias.detach().to("cuda"), device=tp.device(kind)))
             else:
-                tripy_linear.weight = tp.Parameter(tp.Tensor(torch_linear.weight, device=tp.device(kind)))
-                tripy_linear.bias = tp.Parameter(tp.Tensor(torch_linear.bias, device=tp.device(kind)))
+                tripy_linear.weight = tp.Parameter(tp.Tensor(torch_linear.weight.detach(), device=tp.device(kind)))
+                tripy_linear.bias = tp.Parameter(tp.Tensor(torch_linear.bias.detach(), device=tp.device(kind)))
 
             tripy_out = tripy_linear(tp.Tensor(inp, device=tp.device(kind)))
             assert tp.allclose(tripy_out, tp.Tensor(torch_out))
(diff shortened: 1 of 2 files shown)


tripy/tests/integration/test_functional.py
```diff
@@ -166,11 +166,11 @@ def test_weights_loading_from_torch(self, kind):
 
             tripy_linear = tp.Linear(2, 3)
             if kind == "gpu":
-                tripy_linear.weight = tp.Parameter(tp.Tensor(torch_linear.weight.to("cuda"), device=tp.device(kind)))
-                tripy_linear.bias = tp.Parameter(tp.Tensor(torch_linear.bias.to("cuda"), device=tp.device(kind)))
+                tripy_linear.weight = tp.Parameter(tp.Tensor(torch_linear.weight.detach().to("cuda"), device=tp.device(kind)))
```
jhalakpatel, line 169: What happens when you remove `detach()`?

markkraay: Torch tensors will not allow invocation of the `__dlpack__` method if they have a `grad` tensor attached.

jhalakpatel: can we instead handle this internally? This could lead to bad UX.

markkraay: Are you suggesting we add a check within the `Array._memref` function to see if the `data`  is a torch tensor and then manually calling `detach()`?

tripy/tripy/common/array.py
```diff
@@ -273,70 +273,7 @@ def _memref(self, data):
...
-                        dtype=convert_tripy_dtype_to_runtime_dtype(self.dtype),
-                    )
-                else:
-                    return self.runtime_client.create_device_memref_view(
-                        ptr,
-                        shape=list(self.shape),
-                        dtype=convert_tripy_dtype_to_runtime_dtype(self.dtype),
-                        device=self.runtime_client.get_devices()[self.device.index],
-                    )
+                return self.runtime_client.create_memref_view_from_dlpack(data.__dlpack__())
```
jhalakpatel, line 275: We can only create a host memref view from the host array pointer and a device memref view from the device pointer.
I think we should pass `mlirtrt_device` to this method and check if the pointer type matches the expected device, if we not should throw an error. You use something similar to https://github.com/NVIDIA/TensorRT-Incubator/pull/73/files#diff-1768ba4123934bb7c78a1529f96460e3c2926a3228c036f1a8ffcdd3c982ab20R736 in your MLIR-TRT PR.

markkraay: From my understanding, the DLPack device would correspond to the device that the `data` pointer is on and therefore the device is already determined. Since the device is already determined, passing `mlirtrt_device` and performing an extra check seems unnecessary. Maybe I am misunderstanding something?

jhalakpatel: I agree. I was talking about a mismatch between user set device and location of the data pointer. I can take care of this in a separate PR.

jhalakpatel: Feel free to resolve the thread.

tripy/tests/integration/test_functional.py
jhalakpatel: Yeah, that would be better.

tripy/tripy/common/array.py
markkraay: Okay I understand now. Will close this.

yizhuoz004:
I'm closing this PR because #95 covers all the changes. Thanks.

PR #74, opened 2024-08-08
Add memory allocation tests

jhalakpatel:
Add memory allocation tests to ensure we do not implicitly copy input data.

This change is dependent on the corresponding MLIR-TensorRT PR: https://github.com/NVIDIA/TensorRT-Incubator/pull/73
Signed-off-by: Jhalak Patel [jhalakp@nvidia.com](mailto:jhalakp@nvidia.com)

### DIFF
--- a/tripy/tests/common/test_array.py
+++ b/tripy/tests/common/test_array.py
@@ -176,3 +176,31 @@ class TestArray:
     def test_array_unsupported_python_sequence_type(self, dtype):
         with pytest.raises(AssertionError):
             arr = Array([0], shape=None, dtype=dtype, device=tp.device("cpu"))
+
+    @pytest.mark.parametrize("src, dst", [("cpu", "gpu"), ("gpu", "cpu")])
+    def test_invalid_memory_copy(self, src, dst):
+        with tests.helper.raises(tp.TripyException, f"Cannot allocate tensor that is currently on: {src}:0 on requested device: {dst}:0"):
+            module = np if src == "cpu" else cp
+            data = module.ones((1000,), dtype=np.int32)
+            a = Array(data, None, tp.int32, device=tp.device(dst))
+
+    def test_memory_allocation(self):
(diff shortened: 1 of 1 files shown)


tripy/tests/common/test_array.py
```diff
@@ -176,3 +176,31 @@ def test_array_supported_python_sequence_type(self, dtype):
...
+
+        # (torch) Allocate on gpu -> gpu view
+        torch_data = torch.tensor(data).to(torch.device("cuda"))
+        a = Array(torch_data, None, tp.int32, tp.device("gpu"))
+        assert torch_data.data_ptr() == a.memref_value.ptr
+
+        # (cupy) Allocate on gpu -> gpu view
+        cp_data = cp.ones((1000,), dtype=np.int32)
+        a = Array(cp_data, None, tp.int32, tp.device("gpu"))
+        assert cp_data.data.ptr == a.memref_value.ptr
```
pranavm-nvidia, line 206: Are we getting any different coverage with these? If we are, could we split them into separate tests?

yizhuoz004: We have tests for the same purpose: `tripy/tests/integration/test_functional.py::test_tensor_round_tripping`. Only difference is that it's at the Tensor level.

If we are going to drop `Array`, then these tests probably are not needed.

jhalakpatel: Agreed. We can either remove them completely or add a version similar to the one I added here: https://github.com/NVIDIA/TensorRT-Incubator/pull/73/files#diff-77cb958600670760fcd82a8ba97caf1a3ab9b9f14ff4f1b06513ea0a61e85019R243

parthchadha: @jhalakpatel can we close this PR if the tests are no longer needed?

jhalakpatel:
Closing this MR since we no longer have an Array class.

PR #80, opened 2024-08-09
Only execute shape inference when output is dynamic at runtime

yizhuoz004:
Changes the executor to:
1. Only execute shape inference when output shape is not static
2. Remove most of `Array` instances, this is in preparation for dropping `Array` class.


tripy/tripy/backend/mlir/executor.py
```diff
@@ -126,12 +128,11 @@ def _get_output_tensor_info(self, outputs_runtime_shape, output_devices):
         return outputs_tensor_info
 
     def get_output_tensor_runtime_info(self, inputs, output_devices=List[device]):
-        inputs_shape_memref = self._get_inputs_shape_memref(
-            inputs
-        )  # Can we use executable signature inputs to retrieve this information?
```
yizhuoz004, line 131: A: We can't, the information in the signature is only for compilation, but we need runtime information here.

parthchadha (approved):

pranavm-nvidia (approved):

PR #81, opened 2024-08-09
Add allocated CPU and GPU memory reporting

jhalakpatel:


mlir-tensorrt/executor/include/mlir-executor/Runtime/API/API.h
```diff
@@ -773,6 +773,9 @@ class AllocTracker {
   /// Return true if the tracker's map contains `ptr`.
   bool contains(uintptr_t ptr) const;
 
+  /// Report total CPU and GPU memory allocated by runtime client.
+  std::pair<int64_t, int64_t> reportAllocatedMemory() const;
```
christopherbate, line 777: There actually could  are more  types than just these two, so I'd prefer if we separate it into a struct or array. Array could be indexed by all the potential values of `PointerType`.

mlir-tensorrt/executor/lib/Runtime/API/API.cpp
```diff
@@ -429,6 +429,24 @@ PointerInfo AllocTracker::lookupOrDefault(uintptr_t ptr) const {
   return map.at(ptr);
 }
 
+std::pair<int64_t, int64_t> AllocTracker::reportAllocatedMemory() const {
+    int64_t totalCpuMemory = 0;
```
christopherbate, line 433: We should use `uint64_t` here

mlir-tensorrt/python/bindings/Runtime/RuntimePyBind.cpp
```diff
@@ -775,7 +775,18 @@ PYBIND11_MODULE(_api, m) {
...
+          py::arg("stream") = py::none())
+      .def(
+          "report_allocated_memory",
+          [](PyRuntimeClient &self) {
+            int64_t totalGpuMemory;
+            int64_t totalCpuMemory;
+            MTRT_Status s = mtrtReportAllocatedMemory(self, &totalCpuMemory, &totalGpuMemory);
+            THROW_IF_MTRT_ERROR(s);
+            py::object namedtuple = py::module::import("collections").attr("namedtuple");
+            py::object MemoryUsage = namedtuple("MemoryUsage", "cpu_memory gpu_memory");
```
christopherbate, line 787: You'll need to update the stubs so users can see this type information in the IDE.

mlir-tensorrt/compiler/include/mlir-tensorrt-c/Runtime/Runtime.h
```diff
@@ -221,10 +221,15 @@ mtrtRuntimeClientGetNumDevices(MTRT_RuntimeClient client, int32_t *numDevices);
 MLIR_CAPI_EXPORTED MTRT_Status mtrtRuntimeClientGetDevice(
     MTRT_RuntimeClient client, int32_t index, MTRT_Device *device);
 
-/// Retrieve the runtiem client that manages the specified memref.
+/// Retrieve the runtime client that manages the specified memref.
 MLIR_CAPI_EXPORTED MTRT_RuntimeClient
 mtrtMemRefGetClient(MTRT_MemRefValue memref);
 
+/// Retrieve the runtime client allocated cpu and gpu memory.
```
christopherbate, line 228: This isn't quite accurate. You're reporting the CPU/GPU memory that is being tracked by the RuntimeClient. It can track buffers that are externally allocated.

mlir-tensorrt/compiler/include/mlir-tensorrt-c/Runtime/Runtime.h
```diff
@@ -221,10 +221,15 @@ mtrtRuntimeClientGetNumDevices(MTRT_RuntimeClient client, int32_t *numDevices);
...
+MTRT_Status mtrtReportAllocatedMemory(MTRT_RuntimeClient client,
+                                  int64_t *totalCpuMemory,
+                                  int64_t *totalGpuMemory);
```
christopherbate, line 231: Let's use `uint64_t` or `size_t` here.

christopherbate:
It looks fine, but without knowledge of the use case it's hard to review. This is only giving you the amount of memory tracked by the RuntimeClient at any given point in time. I hope you wouldn't be calling this often (since you're adding it up on the fly). If you need specific memory stats we could build something a bit more robust

mlir-tensorrt/executor/lib/Runtime/API/API.cpp
```diff
@@ -429,6 +429,24 @@ PointerInfo AllocTracker::lookupOrDefault(uintptr_t ptr) const {
...
+    int64_t totalGpuMemory = 0;
+
+    for (const auto &entry : map) {
+        const PointerInfo &info = entry.second;
+        if (info.isExternallyManaged())
```
jhalakpatel, line 438: @christopherbate Is this sufficient for tracking only internally managed/allocated pointers?

pranavm-nvidia:
Closing stale PRs

PR #37, opened 2024-08-06
Added compile fixture

markkraay:
This MR adds a compile fixture to our integration test suite so that we can test ops both lazily and using the compiler API. Addresses #256.

PR #87, opened 2024-08-12
Added array_equal function and tests

markkraay:
While working on #37, noticed that we were using `tp.allclose` in contexts where we should probably be using something similar to `np.array_equal` instead. This MR will:
1. Add `tp.array_equal`
2. Correct the test cases with improper usage of `tp.allclose` (we should probably only be using `tp.allclose` when working with float tensors)
3. Change the test cases which are using the `np.array_equal(cp.from_dlpack(_).get()` trick to use `tp.array_equal`

### DIFF
--- a/tripy/tripy/frontend/ops/array_equal.py
+++ b/tripy/tripy/frontend/ops/array_equal.py
@@ -36,7 +36,7 @@ def array_equal(a: "tripy.Tensor", b: "tripy.Tensor") -> bool:
 
         # doc: print-locals a, b, is_equal
 
-        a = tp.ones((2,2))
+        a = tp.ones((2,2), dtype=tp.int32)
         b = tp.Tensor([
             [1, 1],
             [1, 1]
@@ -45,6 +45,13 @@ def array_equal(a: "tripy.Tensor", b: "tripy.Tensor") -> bool:
         assert is_equal
     """
     from tripy.frontend.trace.ops.reduce import all
+    from tripy.frontend.trace.ops.cast import cast
 
-    # `tp.Shape` overloads `__eq__`, therefore we can directly compare shapes here.
-    return a.shape == b.shape and a.dtype == b.dtype and bool(all(a == b))
+    # `tp.Shape` overloads `__ne__`, therefore we can directly compare shapes here.
+    if a.shape != b.shape:
+        return False
+
+    if a.dtype != b.dtype:
+        b = cast(b, a.dtype)
+
+    return bool(all(a == b))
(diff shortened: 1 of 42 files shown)


tripy/tripy/frontend/ops/array_equal.py
```diff
@@ -0,0 +1,46 @@
...
+    .. code-block:: python
+        :linenos:
+        :caption: Example
+
+        a = tp.ones((2,2))
+        b = tp.Tensor([
+            [1, 1],
+            [1, 1]
+        ])
+        assert tp.array_equal(a, b)
```
pranavm-nvidia, line 42: `assert`s are stripped from docs. Probably want:
```suggestion
        # doc: print-locals a b is_equal
        is_equal = tp.array_equal(a, b)
        assert is_equal
```

tripy/tripy/frontend/ops/array_equal.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    from tripy.frontend.trace.ops.reduce import all
+
+    return a.shape == b.shape and a.dtype == b.dtype and bool(all(a == b))
```
pranavm-nvidia, line 46: This works because shape tensors overload `==` right? Let's add a comment here explaining that for paranoid readers like me :)

markkraay:
@pranavm-nvidia Should we auto-convert input tensors to `tp.Tensor` for this function? Then we could avoid wrapping the arguments in `tp.Tensor` and handle this implicitly.

pranavm-nvidia:
> @pranavm-nvidia Should we auto-convert input tensors to `tp.Tensor` for this function? Then we could avoid wrapping the arguments in `tp.Tensor` and handle this implicitly.

I'm not entirely sure. You could make this argument for pretty much all Tripy APIs. Let's discuss offline.

tripy/tripy/frontend/ops/array_equal.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    from tripy.frontend.trace.ops.reduce import all
+    from tripy.frontend.trace.ops.cast import cast
+
+    # `tp.Shape` overloads `__ne__`, therefore we can directly compare shapes here.
+    if a.shape != b.shape:
+        return False
+
+    if a.dtype != b.dtype:
+        b = cast(b, a.dtype)
```
pranavm-nvidia, line 55: Shouldn't we return False here?

pranavm-nvidia:
Superseded by https://github.com/NVIDIA/TensorRT-Incubator/pull/375

PR #88, opened 2024-08-12
Enable creating ConstantOp from python list/number

yizhuoz004:
This MR enables creating `ConstantOp` from python list/number without creating `Array` instances. It can avoid most `cast` ops because of unsupported dtypes. `ConstantOp` will lower python list/number directly to MLIR.


tripy/tripy/backend/mlir/utils.py
```diff
@@ -80,6 +80,11 @@ def get_mlir_dtype(dtype: "tripy.dtype"):
     }[dtype.name]
 
 
+def get_mlir_scalar_attr(dtype: "tripy.dtype", value):
+    attr_func = ir.FloatAttr.get if "float" in dtype.name else ir.IntegerAttr.get
```
parthchadha, line 84: Add assertion that dtype is Integer or Float.

tripy/tripy/flat_ir/ops/constant.py
```diff
@@ -31,10 +32,11 @@
 @dataclass(repr=False)
 class ConstantOp(BaseFlatIROp):
 
-    data: Array
+    data: Union[Array, Sequence]
```
pranavm-nvidia, line 35: Do we need to be able to give it a shape?

pranavm-nvidia (approved):

tripy/tripy/backend/mlir/utils.py
pranavm-nvidia: Also we should check this by `issubclass(dtype, tp.floating)` instead of string comparison.

tripy/tripy/flat_ir/ops/constant.py
yizhuoz004: Not really, any `shape` argument will already be transformed into a tensor before passing it to `ConstantOp`.

tripy/tripy/backend/mlir/utils.py
yizhuoz004: There is no restriction on `dtype` here. The logic is any float tripy dtype translates to `FloatAttr`, and anything else translates to `IntegerAttr` (including bool).
```
tp.float32 -> mlir.f32
tp.float16-> mlir.f16
tp.bfloat16 -> mlir.bf16
...
tp.int32 -> mlir.i32
tp.bool -> mlir.i1
```

yizhuoz004: Switched to use `tp.floating`.

pranavm-nvidia: Could you add a comment explaining that? Also wondering whether we should make this more future-proof against other data types we could conceivably add (e.g. string?)

yizhuoz004: Added a comment. Basically MLIR dtypes only has 2 categories: `FloatType` or `IntegerType`:

For a new normal datatype (e.g. float64, fp4), we create it under `_make_float_datatype` or `_make_int_datatype`, automatically assigning it with a float/int parent class.

For other new datatypes that cannot be represented by MLIR (if we really want to add), lots of functions in the code base will break because we assume dtypes are representable in MLIR. Then a refactor is needed.

PR #92, opened 2024-08-13
[Tripy] Add `__len__` for `tp.Shape` and infer length statically when possible

slyubomirsky:
For some cases, it is useful to know the length of a `tp.Shape` without executing the model. This PR adds a method `infer_len` that allows operators to specify how to statically infer the length of `Shape` outputs when possible (it is always optional). Test cases are added.

### DIFF
--- a/tripy/tripy/frontend/trace/ops/expand.py
+++ b/tripy/tripy/frontend/trace/ops/expand.py
@@ -50,7 +50,7 @@ class Expand(BaseTraceOp):
 
     def infer_rank(self):
         if self.output_rank is None:
-            out_shape = op_utils.get_op_input_shape(self.inputs[1])
+            out_shape = op_utils.get_trace_shape(self.inputs[1])
             assert len(out_shape) == 1
             assert out_shape[0] >= 0, f"incorrect shape computation {out_shape}"
             self.output_rank = out_shape[0]
--- a/tripy/tripy/frontend/trace/ops/expand.py
+++ b/tripy/tripy/frontend/trace/ops/expand.py
@@ -16,7 +16,7 @@
 #
 
 from dataclasses import dataclass
-from typing import Sequence, Union
+from typing import Optional, Sequence, Union
 
 from tripy import export
 from tripy.utils import Result
@@ -29,6 +29,7 @@ from tripy.frontend.trace.ops.base import BaseTraceOp
 @dataclass(repr=False)
 class Expand(BaseTraceOp):
     output_rank: int
+    output_len: Optional[int]
 
     def infer_dtypes(self):
         self.outputs[0].dtype = self.inputs[0].dtype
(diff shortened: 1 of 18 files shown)


tripy/tripy/frontend/trace/ops/expand.py
```diff
@@ -29,6 +29,7 @@
 @dataclass(repr=False)
 class Expand(BaseTraceOp):
     output_rank: int
+    output_len: Optional[int]
```
pranavm-nvidia, line 32: Does this mean anything for non-shape tensors? Is it set in that case? Could you document the behavior in a comment here?

tripy/tripy/frontend/trace/ops/slice.py
```diff
@@ -37,6 +37,11 @@ def infer_rank(self):
         # How can we compute the output rank in the case when start, size, stride tensors are dynamic?
         self.outputs[0].rank = self.inputs[0].rank
 
+    def infer_len(self):
+        # Skipping inference here since it depends on the concrete _values_ of the inputs
+        # rather than their shapes. This can be revisited if necessary
+        return [None]
```
pranavm-nvidia, line 43: Not sure we can have dynamic slice on a shape tensor. Even if we could, the shape tensor wouldn't be usable to create an execution tensor since dynamic rank is not supported. Maybe we can assume that the slice bounds are always constants for shape tensors - this will probably be a bit complicated as we would need constant folding to make it work reliably.

tripy/tripy/frontend/trace/ops/utils.py
```diff
@@ -113,6 +113,29 @@ def never_return_shape(self, inputs):
         return Result.ok([])
 
 
+##
+## Inferring shape lengths (helpers)
+##
+
+
+def get_op_input_shape(input: "TraceTensor") -> Sequence[int]:
```
pranavm-nvidia, line 121: Why not just `get_shape`?

tripy/tripy/frontend/trace/ops/expand.py
slyubomirsky: No, it should not be used otherwise. I will add a comment

tripy/tripy/frontend/trace/ops/slice.py
slyubomirsky: `tp.Shape`s in Tripy are just a wrapper on ordinary tensors; they are not implemented as TRT shape tensors. Hence, any operation possible on other tensors is possible on them.

That said, we could do inference in this case by having special handling for constants as in some of the other cases in this PR.

tripy/tripy/frontend/trace/ops/utils.py
slyubomirsky: Changed it to `get_trace_shape` to avoid confusion with other functions with names of the form `get_shape*`

slyubomirsky:
I can confirm that the CI passes.

tripy/tripy/frontend/trace/ops/slice.py
slyubomirsky: Added in a workaround for slice, though it's not the prettiest (lots of corner cases, but that is what makes Python slicing semantics user-friendly).

tripy/tripy/frontend/trace/ops/where.py
```diff
@@ -53,6 +53,10 @@ def infer_shape_output_idxs(self, inputs):
                 ]
             )
 
+    def infer_len(self):
+        # broadcast to the largest of the input lengths
+        return [max(map(lambda inp: op_utils.get_trace_shape(inp)[0], self.inputs))]
```
parthchadha, line 58: @slyubomirsky looks like we are calling `get_trace_shape` in almost all `infer_len` implementations, isn't this very expensive? (at-least with the current implementation of `get_trace_shape` where we are recursing back to inputs in almost every case).

slyubomirsky: It can be expensive but we do need to know the input shapes in these cases. In some cases (e.g., constants), the shape is set directly so `get_trace_shape` would not have to call into the `ShapeContext`. I do not see any way to avoid it besides doing shape inference in Tripy (which we had reasons for avoiding).

parthchadha:
@slyubomirsky can you rebase your branch? This will help run CI on github.

parthchadha (approved):

tripy/tripy/frontend/trace/ops/slice.py
slyubomirsky: Resolving since the new implementation handles the case of a constant slice and it's indeed very likely to be the most common case.

PR #96, opened 2024-08-14
Allow containers of homogenous superclass in state dict

akhilg-nv:
Updates homogenous container check to allow subclasses that all derive from the same superclass to be "mixed" in the same container. This effectively allows lists/dicts of modules or params to be registered in the state dict. 

L0 test: 1959 passed, 47 skipped, 1 deselected, 1 warning

### DIFF
--- a/tripy/tripy/frontend/module/module.py
+++ b/tripy/tripy/frontend/module/module.py
@@ -40,8 +40,8 @@ def _check_param_compatible(original_param, new_param, param_name):
         )
 
 
-def _is_homogeneous_container(container: Sequence, types: List[T]):
-    return any(all(isinstance(op, typ) for op in container) for typ in types)
+def _is_homogeneous_container(container: Sequence, typ: T):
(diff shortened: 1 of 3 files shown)


tripy/tripy/frontend/module/module.py
```diff
@@ -106,7 +106,7 @@ def __setattr__(self, name: str, value: Any) -> None:
 
         if isinstance(value, List) or isinstance(value, Dict):
             container = value if isinstance(value, List) else value.values()
-            if _contains_types(container, [Parameter, Module]) and not _is_homogeneous_container(container):
+            if _contains_types(container, [Parameter, Module]) and not _is_homogeneous_container(container, [Module, Parameter]):
```
parthchadha, line 109: @akhilg-nv lets add a unit test for this case.

akhilg-nv: done, please check.

parthchadha (approved):

PR #98, opened 2024-08-14
Update Convolution & GroupNorm for dynamic shapes

akhilg-nv:
Avoids expensive shape of shape evaluation.

L0 Test: 1959 passed, 47 skipped, 1 deselected, 1 warning

### DIFF
--- a/tripy/tripy/frontend/module/groupnorm.py
+++ b/tripy/tripy/frontend/module/groupnorm.py
@@ -79,9 +79,8 @@ class GroupNorm(Module):
             torch_tensor = torch.from_dlpack(input) # doc: omit
             torch_gn = torch.nn.GroupNorm(2, 4).to(torch.device("cuda")) # doc: omit
             torch_out = cp.from_dlpack(torch_gn(torch_tensor).detach()).get() # doc: omit
-            atol_ = 3e-2 # doc: omit
             assert np_out.shape == torch_out.shape # doc: omit
-            assert np.allclose(np_out, torch_out, atol=atol_) # doc: omit
+            assert np.allclose(np_out, torch_out) # doc: omit
         """
         super().__init__()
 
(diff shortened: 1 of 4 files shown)


tripy/tripy/frontend/module/groupnorm.py
```diff
@@ -70,15 +70,17 @@ def __init__(self, num_groups: int, num_channels: int, dtype: datatype.dtype = d
             group_norm.weight = tp.ones_like(group_norm.weight)
             group_norm.bias = tp.zeros_like(group_norm.bias)
 
-            input = tp.iota((1, 4, 2, 2))
+            input = tp.iota((1, 4, 2, 2), dim=1)
```
parthchadha, line 73: We should be adding unit tests for this case rather than changing the docstring example.

tripy/tripy/frontend/module/module.py
```diff
@@ -276,13 +276,13 @@ def _iterate_members_of_type(self, typ: T) -> Iterator[Tuple[str, T]]:
         for name, value in vars(self).items():
             if isinstance(value, typ):
                 yield name, value
-            elif isinstance(value, List) and _contains_types(value, [typ]) and _is_homogeneous_container(value):
+            elif isinstance(value, List) and _contains_types(value, [typ]) and _is_homogeneous_container(value, [Module, Parameter]):
```
pranavm-nvidia, line 279: Shouldn't this be:
```suggestion
            elif isinstance(value, List) and _contains_types(value, [typ]) and _is_homogeneous_container(value, [typ]):
```
? 
Same below

tripy/tripy/frontend/module/groupnorm.py
akhilg-nv: The reason for this change is because `tp.iota((1, 4, 2, 2))` is simply equivalent to `tp.zeros((1, 4, 2, 2)),` which is not very useful for either demoing how GroupNorm works and docstring tests.

parthchadha (approved):

PR #99, opened 2024-08-14
Disable canonicalize test for TensorRT dialect

shelkesagar29:
This PR disables `canonicalize.mlir` test for TensorRT dialect to avoid CI failures until fix is pushed.


jhalakpatel (approved):

christopherbate:
superseded #103

PR #100, opened 2024-08-14
Add Stable Diffusion demo

akhilg-nv:
Cleanup for Tripy inference, torch inference pipeline, and accuracy test are still WIP. README is currently minimal as well.


tripy/examples/diffusion/example.py
```diff
@@ -0,0 +1,222 @@
...
+    run_start_time = time.perf_counter()
+
+    # Run through CLIP to get context
+    tokenizer = ClipTokenizer()
+    prompt = tp.Tensor([tokenizer.encode(args.prompt)])
+    print(f"Got tokenized prompt.")
+    unconditional_prompt = tp.Tensor([tokenizer.encode("")])
+    print(f"Got unconditional tokenized prompt.")
+
+    print("Compiling CLIP model...")
```
parthchadha, line 28: @akhilg-nv we should have model compilation (for all the models in diffusion) be separate from execution. Then we can also introduce serialization of compiled models to disk so that can you just run standalone inference. Also, it makes the code easier to understand.

tripy/examples/diffusion/model.py
```diff
@@ -0,0 +1,775 @@
...
+    for b in range(2**8):
+        if b not in bs:
+            bs.append(b)
+            cs.append(2**8 + n)
+            n += 1
+    cs = [chr(n) for n in cs]
+    return dict(zip(bs, cs))
+
+
+class ClipTokenizer:
```
parthchadha, line 649: `ClipTokenizer` is not using `Tripy`, why don't we just use HF ClipTokenizer?

tripy/examples/diffusion/model.py
```diff
@@ -0,0 +1,775 @@
...
+
+
+class CLIPTextTransformer(tp.Module):
+    def __init__(self):
+        self.embeddings = CLIPTextEmbeddings()
+        self.encoder = CLIPEncoder()
+        self.final_layer_norm = tp.LayerNorm(768)
+
+    def __call__(self, input_ids):
+        max_seq_length = 77 # input_ids.shape[1]
```
parthchadha, line 562: Can you file an issue for this? Also, add a comment why 77?

tripy/examples/diffusion/model.py
```diff
@@ -0,0 +1,775 @@
...
+    def __call__(self, hidden_states, causal_attention_mask):
+        for l in self.layers:
+            hidden_states = l(hidden_states, causal_attention_mask)
+        return hidden_states
+
+
+class CLIPTextEmbeddings(tp.Module):
+    def __init__(self):
+        self.token_embedding = tp.Embedding(49408, 768)
+        self.position_embedding = tp.Embedding(77, 768)
```
parthchadha, line 549: Define 77 somewhere and tag it with the issue (to remove it)

tripy/examples/diffusion/model.py
```diff
@@ -0,0 +1,775 @@
...
+
+    def __call__(self, hidden_states, causal_attention_mask):
+        for l in self.layers:
+            hidden_states = l(hidden_states, causal_attention_mask)
+        return hidden_states
+
+
+class CLIPTextEmbeddings(tp.Module):
+    def __init__(self):
+        self.token_embedding = tp.Embedding(49408, 768)
```
parthchadha, line 548: All numbers should ideally be defined at the top of this file in config class. (ex: `config.hidden_size = 768`)

tripy/examples/diffusion/model.py
```diff
@@ -0,0 +1,775 @@
...
+
+    def __call__(self, x):
+        return x
+
+
+class FeedForward(tp.Module):
+    def __init__(self, dim, mult=4):
+        self.net = [
+            GEGLU(dim, dim * mult),
+            Dummy(),  # needed for weights loading code to work
```
parthchadha, line 265: I am not sure if I understand why this is required.

tripy/examples/diffusion/test_acc.py
```diff
@@ -0,0 +1,22 @@
+import tripy as tp
```
pranavm-nvidia, line 1: Can we make the test live under `tests/` rather than with the example? I think the example should produce an image, not an accuracy comparison.

tripy/tripy/frontend/module/groupnorm.py
```diff
@@ -70,15 +70,17 @@ def __init__(self, num_groups: int, num_channels: int, dtype: datatype.dtype = d
             group_norm.weight = tp.ones_like(group_norm.weight)
             group_norm.bias = tp.zeros_like(group_norm.bias)
 
-            input = tp.iota((1, 4, 2, 2))
+            input = tp.iota((1, 4, 2, 2), dim=1)
```
pranavm-nvidia, line 73: Looks like this is dependent on your other PR. Can you move this to draft mode until that's merged?

tripy/examples/diffusion/model.py
```diff
@@ -0,0 +1,775 @@
+# https://arxiv.org/pdf/2112.10752.pdf
+# https://github.com/ekagra-ranjan/huggingface-blog/blob/main/stable_diffusion.md
```
pranavm-nvidia, line 2: Could you provide a little bit more context on the links?

tripy/examples/diffusion/model.py
```diff
@@ -0,0 +1,775 @@
...
+) -> tp.Tensor:
+    """
+    Computes scaled dot-product attention.
+    `self` is the query tensor, `key` is the key tensor, and `value` is the value tensor.
+
+    - Described: https://paperswithcode.com/method/scaled
+    - Paper: https://arxiv.org/abs/1706.03762v7
+    """
+
+    if is_causal:  # this path is not called in demoDiffusion
```
pranavm-nvidia, line 32: Can we drop this if it's not called?

tripy/examples/diffusion/model.py
```diff
@@ -0,0 +1,775 @@
...
+        # TODO: #228: WAR to prevent computing output rank in infer_rank for reshape
+        target_shape.trace_tensor.shape = (2,)
+        attn_mask = tp.cast(tp.tril(tp.ones(target_shape)), tp.bool)
+    if attn_mask is not None and attn_mask.dtype == tp.bool:
+        attn_mask = tp.where((attn_mask == 0), tp.ones_like(attn_mask) * -float("inf"), tp.zeros_like(attn_mask))
+    qk = query @ tp.transpose(key, -2, -1) / math.sqrt(embedding_dim)
+    return tp.cast(tp.softmax((qk + attn_mask) if attn_mask is not None else qk, -1), query.dtype) @ value
+
+
+def sequential(input: tp.Tensor, ll: List[Callable[[tp.Tensor], tp.Tensor]]):
```
pranavm-nvidia, line 43: I think it's worth making this part of the API, similar to `torch.nn.Sequential`.

tripy/examples/diffusion/model.py
```diff
@@ -0,0 +1,775 @@
...
+    """
+    Applies a sequence of functions to `self` chaining the output of each function to the input of the next.
+    """
+    return reduce(lambda x, f: f(x), ll, input)
+
+
+# convenience for dynamic reshapes
+one_shape = tp.Shape(tp.ones((1,), dtype=tp.int32))
```
pranavm-nvidia, line 51: Why can't we just use `[1]`?

tripy/examples/diffusion/model.py
```diff
@@ -0,0 +1,775 @@
...
+        self.to_k = tp.Linear(in_channels, in_channels)
+        self.to_v = tp.Linear(in_channels, in_channels)
+        self.to_out = [tp.Linear(in_channels, in_channels)]
+        self.in_channels = in_channels
+
+    # adapted from AttnBlock in ldm repo
+    def __call__(self, x):
+        h_ = self.group_norm(x)
+
+        b, c, h, w = h_.shape[0], h_.shape[1], h_.shape[2], h_.shape[3]
```
pranavm-nvidia, line 67: Since we can determine the length of the shape after @slyubomirsky's PR, could we use `__iter__` to do the unpacking? Then this would be:
```py
b, c, h, w = h_.shape
```

tripy/examples/diffusion/model.py
pranavm-nvidia: Also curious. I'm assuming it's to get the list offset right, but what was here in the HF model? Can we just change the weight loading code to account for this?

tripy/examples/diffusion/model.py
```diff
@@ -0,0 +1,775 @@
...
+        self.final_layer_norm = tp.LayerNorm(768)
+
+    def __call__(self, input_ids):
+        max_seq_length = 77 # input_ids.shape[1] 
+        x = self.embeddings(input_ids, tp.reshape(tp.iota((max_seq_length,), dtype=tp.int32), (1, -1)))
+        x = self.encoder(x, tp.triu(tp.full((1, 1, 77, 77), float("-inf")), 1))
+        return self.final_layer_norm(x)
+
+
+# Clip tokenizer, taken from https://github.com/openai/CLIP/blob/main/clip/simple_tokenizer.py (MIT license)
```
pranavm-nvidia, line 568: Couldn't we import this as a dependency instead of inlining the code?

tripy/examples/diffusion/model.py
```diff
@@ -0,0 +1,775 @@
...
+
+        residual = hidden_states
+        hidden_states = self.layer_norm2(hidden_states)
+        hidden_states = self.mlp(hidden_states)
+        hidden_states = residual + hidden_states
+
+        return hidden_states
+
+
+class CLIPEncoder(tp.Module):
```
pranavm-nvidia, line 536: Could we split up each model into its own file? I think that'll make this a lot easier to read.

tripy/examples/diffusion/example.py
```diff
@@ -0,0 +1,222 @@
...
+
+    # save image
+    im = Image.fromarray(cp.from_dlpack(x).get().astype(np.uint8, copy=False))
+    print(f"saving {args.out}")
+    im.save(args.out)
+    # Open image.
+    if not args.noshow:
+        im.show()
+
+def hf_diffusion(args):
```
pranavm-nvidia, line 121: Why do we want this?

tripy/examples/diffusion/model.py
akhilg-nv: It's to account for a Dropout layer in the HF model and to get the list offset correct, yeah. I can clarify the comment with more details.

tripy/examples/diffusion/example.py
akhilg-nv: This allows a pure torch inference mode for comparison with the example script - I also plan to use it for the e2e accuracy test that compares the two output images.

tripy/examples/diffusion/model.py
akhilg-nv: I have left it for now in case we want to add it as part of the API somewhere

tripy/examples/diffusion/model.py
pranavm-nvidia: This is being done here: https://github.com/NVIDIA/TensorRT-Incubator/pull/321

tripy/examples/diffusion/README.md
```diff
@@ -0,0 +1,40 @@
+# Implementing Stable Diffusion With Tripy
+
+## Introduction
+
+This example demonstrates how to implement a Stable Diffusion model using Tripy APIs.
+
+It's broken up into three components:
```
pranavm-nvidia, line 7: We can shorten this a little bit:
```suggestion
This example implements a Stable Diffusion model using Tripy.
There are 3 components:
```

tripy/examples/diffusion/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+1. `model.py` defines the model using `tripy.Module` and associated APIs. `clip_model.py`, `unet_model.py`, `vae_model.py` implement specific components of the diffusion model.
```
pranavm-nvidia, line 9: Can we organize these files under a `model/` subdirectory?

tripy/examples/diffusion/README.md
```diff
@@ -0,0 +1,40 @@
...
+    python3 -m pip install -r requirements.txt
+    ```
+
+2. Run the example:
+
+    ```bash
+    python3 example.py --seed 420 --steps 50 --prompt "a beautiful photograph of Mt. Fuji during cherry blossom" --fp16 --engine-dir fp16_engines --verbose
+    ```
+
+3. (Optional) Compare with torch reference to verify accuracy:
```
pranavm-nvidia, line 29: nit: In other READMEs we use this format for optional steps:
```suggestion
3. **[Optional]** Compare with torch reference to verify accuracy:
```

tripy/examples/diffusion/clip_model.py
```diff
@@ -0,0 +1,130 @@
...
+class CLIPAttention(tp.Module):
+    def __init__(self, config: CLIPConfig):
+        self.embed_dim = config.embedding_size
+        self.num_heads = config.num_heads
+        self.head_dim = self.embed_dim // self.num_heads
+        self.k_proj = tp.Linear(self.embed_dim, self.embed_dim, dtype=config.dtype)
+        self.v_proj = tp.Linear(self.embed_dim, self.embed_dim, dtype=config.dtype)
+        self.q_proj = tp.Linear(self.embed_dim, self.embed_dim, dtype=config.dtype)
+        self.out_proj = tp.Linear(self.embed_dim, self.embed_dim, dtype=config.dtype)
+        self.dtype = config.dtype
```
pranavm-nvidia, line 56: Do we need to store the dtype here?

tripy/examples/diffusion/clip_model.py
```diff
@@ -0,0 +1,130 @@
...
+
+class CLIPTextTransformer(tp.Module):
+    def __init__(self, config: CLIPConfig):
+        self.embeddings = CLIPTextEmbeddings(config)
+        self.encoder = CLIPEncoder(config)
+        self.final_layer_norm = tp.LayerNorm(config.embedding_size, dtype=config.dtype)
+        self.max_seq_len = config.max_seq_len
+
+    def __call__(self, input_ids):
+        x = self.embeddings(input_ids, tp.reshape(tp.iota((input_ids.shape[1],), dtype=tp.int32), (1, -1)))
```
pranavm-nvidia, line 128: Might want to just use `tp.unsqueeze` here.

tripy/examples/diffusion/clip_model.py
```diff
@@ -0,0 +1,130 @@
...
+        x = self.encoder(x, tp.triu(tp.full((1, 1, self.max_seq_len, self.max_seq_len), float("-inf")), 1))
```
pranavm-nvidia, line 129: I think the `tp.triu` can be generated in the constructor? Ideally it would be constant folded by the compiler, but it would be interesting to see if it buys us any perf.

tripy/examples/diffusion/README.md
```diff
@@ -0,0 +1,40 @@
...
+    ```bash
+    python3 compare_images.py --seed 420 --steps 50 --prompt "a beautiful photograph of Mt. Fuji during cherry blossom" --fp16 --engine-dir fp16_engines
```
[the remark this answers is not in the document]
pranavm-nvidia: That would also simplify the `compare_images.py` script a lot.

tripy/examples/diffusion/requirements.txt
```diff
@@ -0,0 +1,12 @@
+-f https://nvidia.github.io/TensorRT-Incubator/packages.html
+nvtripy
```
pranavm-nvidia, line 2: Probably should have a minimum version here - 0.1.2?

tripy/examples/diffusion/requirements.txt
```diff
@@ -0,0 +1,12 @@
...
+torch>=2.3.1
+torchvision>=0.18.1
+numpy>=1.24.4
+tqdm>=4.66.1
+accelerate==1.8.1
+# iopath>=0.1.10
```
pranavm-nvidia, line 8:
```suggestion
```

tripy/examples/diffusion/unet_model.py
```diff
@@ -0,0 +1,327 @@
...
+        x, gate = tp.split(proj, 2, proj.rank - 1)  # TODO: allow negative dim in split
+        return x * tp.gelu(gate)
+
+
+class Dummy(tp.Module):
+    def __init__(self):
+        pass
+
+    def __call__(self, x):
+        return x
```
pranavm-nvidia, line 105: We can use `lambda x: x` in place of this

tripy/examples/diffusion/unet_model.py
```diff
@@ -0,0 +1,327 @@
...
+            CrossAttnUpBlock2D(config, up_channels[4:7], down_channels[1], use_upsampler=False),
+        ]
+        self.conv_norm_out = tp.GroupNorm(32, config.model_channels, dtype=config.dtype)
+        self.conv_act = tp.silu
+        self.conv_out = tp.Conv(
+            config.model_channels, config.io_channels, (3, 3), padding=((1, 1), (1, 1)), dtype=config.dtype
+        )
+
+    def __call__(self, x, timesteps=None, context=None):
+        # TODO: real time embedding
```
pranavm-nvidia, line 302: Is this TODO still relevant?

tripy/examples/diffusion/weight_loader.py
```diff
@@ -0,0 +1,51 @@
...
+    tripy_keys = tripy_state_dict.keys()
+
+    hf_state_dict = hf_model.state_dict()
+    hf_keys = hf_state_dict.keys()
+
+    assert_msg = f"Mismatched keys: {hf_keys} != {tripy_keys}"
+    if debug and len(hf_keys) != len(tripy_keys):
+        print("\nERROR: unused weights in HF_state_dict:\n", sorted(list(hf_keys - tripy_keys)))
+        print("\nERROR: unused weights in tripy_state_dict:\n", sorted(list(tripy_keys - hf_keys)))
+        assert_msg = f"Mismatched keys between HF and provided model, see above."
```
pranavm-nvidia, line 32: These checks are done by `load_state_dict`, so we can probably drop them here

tripy/examples/diffusion/clip_model.py
akhilg-nv: I'm unsure how we would do it otherwise, this is how we propagate dtype information which can be changed via setting this attribute when instantiating the config.

tripy/examples/diffusion/clip_model.py
pranavm-nvidia: Makes sense.

tripy/examples/diffusion/clip_model.py
pranavm-nvidia: I mean that `self.dtype` doesn't appear to be used anywhere (AFAICT)

tripy/examples/diffusion/example.py
```diff
@@ -0,0 +1,285 @@
...
+    parser.add_argument("--steps", type=int, default=30, help="Number of denoising steps in diffusion")
+    parser.add_argument("--prompt", type=str, default=default_prompt, help="Phrase to render")
+    parser.add_argument("--out", type=str, default=None, help="Output filepath")
+    parser.add_argument("--fp32", action="store_true", help="Run the model in fp32 precision")
+    parser.add_argument("--seed", type=int, help="Set the random latent seed")
+    parser.add_argument("--guidance", type=float, default=7.5, help="Prompt strength")
+    parser.add_argument(
+        "--hf-token", type=str, default="", help="HuggingFace API access token for downloading model checkpoints"
+    )
+    parser.add_argument("--engine-dir", type=str, default="engines", help="Output directory for TensorRT engines")
```
pranavm-nvidia, line 272: These would actually be Tripy executables rather than engines right?

pranavm-nvidia: We should make the distinction since the former has extra metadata and so can't be directly deserialized by TRT.

tripy/examples/diffusion/clip_model.py
akhilg-nv: oh yes, was looking at the wrong module. Fixed.

tripy/examples/diffusion/example.py
pranavm-nvidia: Oh I wasn't even aware of that. Yes, we should use "executable" globally ideally. I can make a separate PR for that.

tripy/examples/diffusion/compare_images.py
```diff
@@ -0,0 +1,142 @@
...
+import glob
+
+
+def load_reference_image(image_path, verbose=False):
+    """Load reference image from file path."""
+    if not os.path.exists(image_path):
+        raise FileNotFoundError(f"Reference image not found: {image_path}")
+
+    if verbose:
+        print(f"[I] Loading reference image from {image_path}")
```
pranavm-nvidia, line 32: Doesn't have to be done in this PR, but I don't think these prints are so verbose that they need to be hidden behind a flag. IMO we could probably just drop the verbose option and make all of these unconditional.

tripy/examples/diffusion/example.py
```diff
@@ -0,0 +1,285 @@
...
+            f"seed{args.seed if args.seed else 'rand'}-"
+            f"{int(time.time())}.png"
+        )
+        filename = os.path.join("output", filename)
+
+    # Save image
+    print(f"[I] Saving image to {filename}")
+    if not os.path.isdir(os.path.dirname(filename)):
+        print(f"[I] Creating '{os.path.dirname(filename)}' directory.")
+        os.makedirs(os.path.dirname(filename))
```
pranavm-nvidia, line 138: Alternatively:
```suggestion
    os.makedirs(os.path.dirname(filename), exist_ok=True)
```

tripy/examples/diffusion/models/unet_model.py
```diff
@@ -0,0 +1,318 @@
...
+#
+
+import math
+from typing import List, Tuple
+
+import nvtripy as tp
+from dataclasses import dataclass
+
+from examples.diffusion.helper import scaled_dot_product_attention
+from examples.diffusion.models.vae_model import Upsample, Downsample
```
pranavm-nvidia, line 25: Wondering if we should put shared functions under a `models/blocks.py` or something like that? SDPA could go there as well

tripy/examples/diffusion/example.py
```diff
@@ -0,0 +1,285 @@
...
+        iterator = tqdm(list(range(num_timesteps))[::-1])
+    else:
+        iterator = list(range(num_timesteps))[::-1]
+
+    for index in iterator:
+        latent = model(
+            unconditional_context,
+            context,
+            latent,
+            tp.Tensor(timesteps[index : index + 1]),
```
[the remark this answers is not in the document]
akhilg-nv: Yes, iirc this had to do with how the tensor was initially created in torch (which was done to avoid eager mode compilation for creating a tripy tensor that had to be moved to gpu)

tripy/examples/diffusion/example.py
```diff
@@ -0,0 +1,285 @@
...
+        description="Run Stable Diffusion", formatter_class=argparse.ArgumentDefaultsHelpFormatter
+    )
+    parser.add_argument("--steps", type=int, default=30, help="Number of denoising steps in diffusion")
+    parser.add_argument("--prompt", type=str, default=default_prompt, help="Phrase to render")
+    parser.add_argument("--out", type=str, default=None, help="Output filepath")
+    parser.add_argument("--fp32", action="store_true", help="Run the model in fp32 precision")
+    parser.add_argument("--seed", type=int, help="Set the random latent seed")
+    parser.add_argument("--guidance", type=float, default=7.5, help="Prompt strength")
+    parser.add_argument(
+        "--hf-token", type=str, default="", help="HuggingFace API access token for downloading model checkpoints"
```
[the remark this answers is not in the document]
akhilg-nv: It's not strictly necessary for the particular HF model we are using right now, but in the past it was needed. Having the option to supply it covers future changes.

tripy/examples/diffusion/models/unet_model.py
akhilg-nv: I've refactored `helper.py` to `models/utils.py` and moved these blocks there

tripy/examples/diffusion/example.py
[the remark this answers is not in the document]
pranavm-nvidia: One way to fix those would be to do the slicing within the compiled model and pass the tensor + index as inputs.

tripy/examples/diffusion/models/clip_model.py
```diff
@@ -0,0 +1,129 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import nvtripy as tp
+
+from dataclasses import dataclass
+
+from examples.diffusion.models.utils import scaled_dot_product_attention
```
pranavm-nvidia, line 22: Will this import work if this example is run from within the `diffusion` directory? Might want a relative import here instead.

pranavm-nvidia: Same thing elsewhere.

akhilg-nv: Yes, I found that it still works even when run outside of diffusion directory (this is where I normally run it). I think using relative imports would require adding .init files to create a proper package?

tripy/examples/diffusion/example.py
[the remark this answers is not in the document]
pranavm-nvidia: @akhilg-nv the root cause is that we're taking a view of a tensor at non-aligned addresses. i.e. even thought the tensor may be correctly aligned, the `ith` element of that tensor will not be for certain values of `i`. The right way IMO is to make TRT aware of the whole thing including the loop.

tripy/examples/diffusion/models/clip_model.py
pranavm-nvidia: I meant the opposite - does it work when run from _inside_ the `diffusion` directory? Because how would it know where `examples` is?

pranavm-nvidia: Ah I see, I misinterpreted your first comment. The only nitpick I have with using `examples.diffusion` is that it makes the example less portable. It would be nice if just `from diffusion.models.utils ...` worked as well.

tripy/examples/diffusion/example.py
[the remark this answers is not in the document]
akhilg-nv: After discussing offline, we concluded to keep this change as is and remove the warning. Resolving.

PR #107, opened 2024-08-14
Report informative error when input shapes are out of bounds.

parthchadha:
Catch (and replace with more informative) the error thrown by backend when input shapes don't match expected bounds

### DIFF
--- a/tripy/tripy/backend/compiler_api.py
+++ b/tripy/tripy/backend/compiler_api.py
@@ -222,6 +222,19 @@ class Executable:
                                 tensor,
                             ],
                         )
+            elif "InternalError: failed to set input shape" in str(err) or "Runtime shape mismatch" in str(err):
+                expected_input_shapes = [info.shape_bounds for info in self.get_input_info()]
+                for tensor, expected_bounds, arg_name in zip(input_tensors, expected_input_shapes, self._arg_names):
+                    shape = tensor.shape.data().data()
+                    for i in range(len(shape)):
+                        if shape[i] < expected_bounds[i][0] or shape[i] > expected_bounds[i][1]:
+                            raise_error(
+                                f"Unexpected tensor shape.",
+                                [
+                                    f"For parameter `{arg_name}`, expected the tensor shape `{tensor.shape}` to be within bounds for all dimensions. However, dimension {i} has a shape of {shape[i]}, which is not within the expected bounds of {expected_bounds[i]}. Note: The provided argument was: ",
+                                    tensor,
+                                ],
+                            )
             raise
 
         # TODO (#192): avoid get_stack_info in runtime
--- a/tripy/tests/backend/test_compiler_api.py
+++ b/tripy/tests/backend/test_compiler_api.py
@@ -260,6 +260,16 @@ class TestCompile:
             )
             compiled_add(a, a)
 
+    def test_incorrect_shape_rejected(self):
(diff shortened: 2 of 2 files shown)


tripy/tripy/backend/compiler_api.py
```diff
@@ -222,6 +222,19 @@ def add(a, b):
                                 tensor,
                             ],
                         )
+            elif "InternalError: failed to set input shape" in str(err) or "Runtime shape mismatch" in str(err):
```
pranavm-nvidia, line 225: Is there a better approach we can take here? Checking strings may not be scalable. Can MLIR-TRT emit a more informative error to begin with?

parthchadha: It does print out information from which you can interpret what went wrong but Tripy can report better consistent errors.
Ex:
```
Unexpected tensor shape.
    For parameter `x`, expected the tensor shape `shape(32, 2)` to be within bounds for all dimensions. However, dimension 0 has a shape of 32, which is not within the expected bounds of (1, 16). Note: The provided argument was: 

    --> /tripy/tripy/frontend/trace/ops/fill.py:121 in full()
          |
      121 |     return full_impl(shape, value, dtype, output_rank)
          | 

    --> /tripy/tripy/frontend/ops/tensor_initializers.py:57 in ones()
          |
       57 |     return full(shape, 1, dtype)
          |            ^^^^^^^^^^^^^^^^^^^^^ --- required from here

    --> /tripy/tester.py:18 in <module>()
          |
       18 | inp = tp.ones((32, 2), dtype=tp.float32)
          |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ --- required from here
```
vs 
```
IExecutionContext::setInputShape: Error Code 3: API Usage Error (Parameter check failed, condition: satisfyProfile. Set dimension [32,2] for tensor arg0 does not satisfy any optimization profiles. Valid range for profile 0: [1,2]..[16,2].)
Traceback (most recent call last):
  File "/tripy/tester.py", line 19, in <module>
    print(fast_geglu(inp))
  File "/tripy/tripy/backend/compiler_api.py", line 210, in __call__
    executor_outputs = self._executor.execute(self._output_devices, input_tensors)
  File "/tripy/tripy/utils/utils.py", line 73, in wrapper
    result = func(*args, **kwargs)
  File "/tripy/tripy/backend/mlir/executor.py", line 181, in execute
    self.session.execute_function("main", in_args=in_args, out_args=out_args, stream=self.stream)
mlir_tensorrt.runtime._mlir_libs._api.MTRTException: InternalError: InternalError: failed to run function "main": [string "__check_for_function("_load_i64");..."]:35: InternalError: failed to set input shape
stack traceback:
        [C]: in function 'base._trtrt_enqueue'
        [string "__check_for_function("_load_i64");..."]:35: in function 'base.main'
```

yizhuoz004: We could probably do this check explicilty and do not need to rely on MLIR-TRT. See `Executable.get_input_info()`, it returns the `min, max` bounds for each input.

pranavm-nvidia (approved):

tripy/tripy/backend/compiler_api.py
pranavm-nvidia: That's a good idea and will be robust against changes in the error message emitted by MLIR-TRT.

PR #112, opened 2024-08-15
Update installation instructions to include pip install

parthchadha:


tripy/README.md
```diff
@@ -16,9 +16,12 @@ user experience without compromising performance. Some of the features of Tripy
 
 ## Installation
 
-Due to how our package is currently set up, installation is more complicated than a simple `pip install`. 
-This should change in the near future, but for now, you can use Tripy by building the development container
-as specified in [CONTRIBUTING.md](./CONTRIBUTING.md).
+```bash
+python3 -m pip install --no-index -f https://nvidia.github.io/TensorRT-Incubator/packages.html tripy --no-deps
+python3 -m pip install -f https://nvidia.github.io/TensorRT-Incubator/packages.html tripy
```
yizhuoz004, line 21: May I ask what's the difference between the 2 commands? Are they both required?

yizhuoz004 (approved):

tripy/README.md
parthchadha: The first one forces to just install `tripy` and no dependencies from the given link and second one installs the dependencies of `tripy`. There is a name conflict with tripy package in `pypi` and therefore we need to force install (1st step) our `tripy`.

PR #119, opened 2024-08-16
Make `tp.Layernorm` 1:1 with torch

markkraay:
- Make  the `normalized_shape` argument 1:1 with Torch. Previously, our API only supported taking a single integer for the normalization dimension. Now, a list of integers can be provided.
- Add integration tests for `tp.LayerNorm` and `tp.GroupNorm`

### DIFF
--- a/tripy/tripy/frontend/module/layernorm.py
+++ b/tripy/tripy/frontend/module/layernorm.py
@@ -112,7 +112,9 @@ class LayerNorm(Module):
         D = len(self.normalized_shape)
 
         if x.shape[-D:] != self.normalized_shape:
-            raise_error(f"The input's last {D} dimensions must have a shape of {self.normalized_shape} and received {x.shape[-D:].as_tensor().data()}")
+            raise_error("Unexpected input shape",
+                [f"The input's last {D} dimensions must have a shape of {self.normalized_shape} and received {x.shape[-D:].data()}"]
+            )
 
         reduce_dims = tuple(-i for i in range(D, 0, -1))
         mean_val = mean(x, dim=reduce_dims, keepdim=True)
--- a/tripy/tripy/frontend/module/layernorm.py
+++ b/tripy/tripy/frontend/module/layernorm.py
@@ -109,14 +109,14 @@ class LayerNorm(Module):
             A tensor of the same shape as the input.
         """
         from tripy.frontend.trace.ops.reduce import mean, var
+        from tripy.frontend.shape import Shape
         from tripy.frontend.trace.ops.unary_elementwise import rsqrt
         from tripy.common.exception import raise_error
-        import tripy as tp
 
         # The mean and the variance are computed over the last D dimensions
         D = len(self.normalized_shape)
 
-        if x.shape[-D:] != tp.Shape(self.normalized_shape):
+        if x.shape[-D:] != Shape(self.normalized_shape):
             raise_error(f"The input's last {D} dimensions must have a shape of {self.normalized_shape} and received {x.shape[-D:].as_tensor().data()}")
 
         reduce_dims = tuple(-i for i in range(D, 0, -1))
--- a/tripy/tripy/frontend/module/layernorm.py
+++ b/tripy/tripy/frontend/module/layernorm.py
@@ -16,6 +16,7 @@
 #
 
 from dataclasses import dataclass
+from typing import Union, Tuple
 
 from tripy import export, utils
 from tripy.common import datatype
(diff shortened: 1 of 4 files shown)


tripy/tripy/frontend/module/layernorm.py
```diff
@@ -47,7 +54,10 @@ class LayerNorm(Module):
     eps: float
     """A value added to the denominator to prevent division by zero."""
 
-    def __init__(self, normalized_shape: int, dtype: datatype.dtype = datatype.float32, eps: float = 1e-5) -> None:
+    correction: int
+    """Difference between the sample size and the degrees of freedom."""
+
+    def __init__(self, normalized_shape: Union[int, Tuple[int]], dtype: datatype.dtype = datatype.float32, eps: float = 1e-5, correction: int = 0) -> None:
```
akhilg-nv, line 60: [Torch layernorm](https://pytorch.org/docs/stable/generated/torch.nn.LayerNorm.html) doesn't expose correction in its API, is there a specific use case in mind for this?

tripy/tripy/frontend/module/layernorm.py
```diff
@@ -47,7 +54,10 @@ class LayerNorm(Module):
     eps: float
     """A value added to the denominator to prevent division by zero."""
 
-    def __init__(self, normalized_shape: int, dtype: datatype.dtype = datatype.float32, eps: float = 1e-5) -> None:
+    correction: int
+    """Difference between the sample size and the degrees of freedom."""
```
akhilg-nv, line 58: Let's specify what it defaults to (correction=0 which is the biased estimator for variance).

akhilg-nv (commented):
Could we also add tests for GroupNorm to the integration tests?

tripy/tripy/frontend/module/layernorm.py
markkraay: Good point. We should probably remove this then. Are there any cases you can think of where this value is not = 0?

akhilg-nv: I'm not aware of any such case where it would be modified. I think we should add both eps and correction to the constant fields - if a user finds a use case for changing the correction we can always change it later.

tripy/tripy/frontend/module/layernorm.py
```diff
@@ -47,7 +54,10 @@ class LayerNorm(Module):
...
+
+    def __init__(self, normalized_shape: Union[int, Tuple[int]], dtype: datatype.dtype = datatype.float32, eps: float = 1e-5, correction: int = 0) -> None:
         """
         Args:
             normalized_shape: The size of the feature dimension of the input over which normalization is performed.
```
akhilg-nv, line 63: Let's make sure to update the description for normalized shape here as well. No need to add correction since we won't expose it in the API.

tripy/tripy/frontend/module/layernorm.py
markkraay: I added normalized_shape but not `eps`. I think it makes sense for `eps` to be changed on the fly. This follows torch.

tripy/tripy/frontend/module/layernorm.py
markkraay: Removed correction.

tripy/tripy/frontend/module/layernorm.py
```diff
@@ -25,19 +26,25 @@
...
+@utils.constant_fields(["dtype", "normalized_shape"])
 class LayerNorm(Module):
     r"""
     Applies layer normalization over the input tensor:
 
     :math:`\text{LayerNorm}(x) = \Large \frac{x - \bar{x}}{ \sqrt{\sigma^2 + \epsilon}} \normalsize * \gamma + \beta`
 
     where :math:`\bar{x}` is the mean and :math:`\sigma^2` is the variance.
+
+    The mean the standard deviation are calculated over the last :math:`D`
```
parthchadha, line 38:
```suggestion
    The mean and standard deviation are calculated over the last :math:`D`
```

parthchadha (approved):

tripy/tripy/frontend/module/layernorm.py
```diff
@@ -92,9 +104,18 @@ def __call__(self, x: "tripy.Tensor") -> "tripy.Tensor":
...
         """
         from tripy.frontend.trace.ops.reduce import mean, var
+        from tripy.frontend.shape import Shape
         from tripy.frontend.trace.ops.unary_elementwise import rsqrt
+        from tripy.common.exception import raise_error
+
+        # The mean and the variance are computed over the last D dimensions
+        D = len(self.normalized_shape)
+
+        if x.shape[-D:] != Shape(self.normalized_shape):
```
pmarathe25, line 114: I thought we're able to compare shape tensors directly against lists? Can we avoid the explicit constructor here?

tripy/tripy/frontend/module/layernorm.py
```diff
@@ -92,9 +104,18 @@ def __call__(self, x: "tripy.Tensor") -> "tripy.Tensor":
...
+            raise_error(f"The input's last {D} dimensions must have a shape of {self.normalized_shape} and received {x.shape[-D:].as_tensor().data()}")
```
pmarathe25, line 115:
```suggestion
            raise_error(f"The input's last {D} dimensions must have a shape of {self.normalized_shape} but received {x.shape[-D:].as_tensor().data()}")
```

pmarathe25: Can we make the summary line something like "Unexpected input shape" and move this to the details section? Also, do we need to change the shape tensor to a regular tensor here?

PR #123, opened 2024-08-19
Fix L1 yml path issue

yizhuoz004:


.github/workflows/tripy-l1.yml
```diff
@@ -29,17 +29,18 @@ jobs:
       url: ${{ steps.deployment.outputs.page_url }}
     container:
       image: ghcr.io/nvidia/tensorrt-incubator/tripy:latest
-      options: --gpus all -v ${{ github.workspace }}/tripy:/tripy
```
yizhuoz004, line 32: The `-v` option is not working here for some unknown reason, adding `cd tripy/` explicitly to solve this problem. The L1 cannot run against a dev branch because it has deployment jobs to github pages, we have to merge it first.

parthchadha (approved):

PR #132, opened 2024-08-20
Add `Shape.__ne__()`

markkraay:
- Adds `Shape.__ne__()` method so that shapes can be conveniently checked for inequality.
- Fixes logic error where `Shape.__eq__()` would ignore checking rank. This would lead to an error when comparing shape's with different ranks.
- Adds shape inequality test cases for shapes with different and same ranks.

### DIFF
--- a/tripy/tripy/frontend/shape.py
+++ b/tripy/tripy/frontend/shape.py
@@ -160,7 +160,15 @@ class Shape(Tensor):
     def __eq__(self, other):
         from tripy.frontend.trace.ops.reduce import all
 
+        # `.shape` will contain a single int. To avoid endlessly calling
+        # Shape.__eq__, we can compare the int's directly using `.data()`
+        if self.shape.data() != other.shape.data():
+            return False 
+
         return bool(all(self.as_tensor() == other.as_tensor()))
+
+    def __ne__(self, other):
+        return not (self == other)
 
     # __len__ for shapes gives the number of dims in the shape, i.e., the first dimension of the shape's shape
     def __len__(self):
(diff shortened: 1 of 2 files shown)


tripy/tripy/frontend/shape.py
```diff
@@ -160,4 +160,12 @@ def __radd__(self, other):
     def __eq__(self, other):
         from tripy.frontend.trace.ops.reduce import all
 
+        # `.shape` will contain a single int. To avoid endlessly calling
+        # Shape.__eq__, we can compare the int's directly using `.data()`
+        if self.shape.data() != other.shape.data():
```
parthchadha, line 165: Not sure why is the return on line 168 required? When will that return False?

markkraay: 168 would return False for something like `tp.Shape([1,2]) == tp.Shape([1,3])`. The check on 165 will check the actual rank. So `tp.Shape([1]) != tp.Shape([1,2])`. This case was missed in my previous MR. If we don't check the rank, Tripy would broadcast the inputs to the same shape, possibly causing `tp.Shape([1]) == tp.Shape([1,1])`.

parthchadha: Oh I didn't read that `self.shape` is shape of shape tensor..

parthchadha (approved):

tripy/tripy/frontend/shape.py
markkraay: Yes. Is there a way to do this that doesn't trigger lowering?

slyubomirsky: You could also write this with `__len__` (just merged my change), though it could in some cases trigger lowering.

slyubomirsky (approved):

PR #135, opened 2024-08-21
Add MLIR-TensorRT CI workflow

shelkesagar29:
This MR adds CI workflow for MLIR-TensorRT project which runs PR created against main branch (except draft PR). This workflow performs format checking and runs LIT tests.

### DIFF
--- a/.github/workflows/mlir-tensorrt-ci.yml
+++ b/.github/workflows/mlir-tensorrt-ci.yml
@@ -1,0 +1,130 @@
+name: MLIR-TensorRT CI
+
+on:
+  pull_request:
+    branches:
+      - main
+    types: [synchronize, opened, reopened, ready_for_review]
+    paths: ["mlir-tensorrt/**"]
+
+env:
+  DEFAULT_IMAGE: ghcr.io/nvidia/tensorrt-incubator/mlir-tensorrt:cuda12.5-ubuntu-llvm17
+  REGISTRY: ghcr.io
+
+jobs:
+  mlir-tensorrt-tests:
+    if: github.event.pull_request.draft == false
+    # `ubuntu-latest` is a CPU runner.
+    # If selected, tests requiring GPU are not run.
+    runs-on: ubuntu-latest
+
+    steps:
+      # Free some disk space, otherwise we get OOM error.
+      - name: Free disk space
+        run: |
+          sudo rm -rf \
+            /usr/share/dotnet "$AGENT_TOOLSDIRECTORY" /usr/local/lib/android /opt/ghc \
+            /usr/local/share/powershell /usr/share/swift /usr/local/.ghcup \
+            /usr/lib/jvm
+
+          sudo apt-get purge microsoft-edge-stable || true
+          sudo apt-get purge google-cloud-cli || true
+          sudo apt-get purge dotnet-sdk-* || true
+          sudo apt-get purge google-chrome-stable || true
+
+          sudo apt-get autoremove -y
+          sudo apt-get autoclean -y
+
+      # Value of `github.workspace` is /home/runner/work/{repo_name}/{repo-name}
+      # i.e. /home/runner/work/TensorRT-Incubator/TensorRT-Incubator in our case.
+      # After this action, repo is cloned inside above path.
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 5
+
+      # Run initial format check
+      - name: Run python format and clang check
+        uses: addnab/docker-run-action@v3
+        with:
+          image: ${{ env.DEFAULT_IMAGE }}
+          options: -v ${{ github.workspace }}:/tensorrt-incubator
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+          # This step does two things
+          # 1. Check if Python files follow black format
+          # 2. Check if C++ files follow clang format
+          # NOTE: We are placed at the root directory ('/') inside the container.
+          run: |
+            cd tensorrt-incubator
+            git config --global --add safe.directory /tensorrt-incubator
+            cat > run_format_check.sh <<EOF
+            #!/bin/bash
+            set -e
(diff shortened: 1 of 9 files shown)


yizhuoz004 (approved):

.github/workflows/mlir-tensorrt-ci.yml
```diff
@@ -0,0 +1,227 @@
...
+
+env:
+  DEFAULT_IMAGE: ghcr.io/nvidia/tensorrt-incubator/mlir-tensorrt:cuda12.5-ubuntu-llvm17
+  ADHOC_TEST_IMAGE: adhoc-test-image:cuda12.5-ubuntu-llvm17
+  REGISTRY: ghcr.io
+
+jobs:
+  mlir-tensorrt-tests:
+    if: github.event.pull_request.draft == false
+    runs-on: tripy-self-hosted
```
christopherbate, line 18: I thought the idea was to use GitHub default runner for this pipeline?

.github/workflows/mlir-tensorrt-ci.yml
```diff
@@ -0,0 +1,227 @@
...
+on:
+  pull_request:
+    branches:
+      - main
+    #types: [synchronize, opened, reopened, ready_for_review]
+    #paths: ['mlir-tensorrt/**']
+
+env:
+  DEFAULT_IMAGE: ghcr.io/nvidia/tensorrt-incubator/mlir-tensorrt:cuda12.5-ubuntu-llvm17
+  ADHOC_TEST_IMAGE: adhoc-test-image:cuda12.5-ubuntu-llvm17
```
christopherbate, line 12: What is 'ADHOC_TEST_IMAGE' for?

.github/workflows/mlir-tensorrt-ci.yml
```diff
@@ -0,0 +1,227 @@
+name: MLIR-TensorRT CI
+
+on:
+  pull_request:
+    branches:
+      - main
+    #types: [synchronize, opened, reopened, ready_for_review]
```
christopherbate, line 7: Why are these commented out? I believe we do want to filter based on files changed, correct?

.github/workflows/mlir-tensorrt-ci.yml
```diff
@@ -0,0 +1,227 @@
...
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+          # This step does two things
+          # 1. Check if Python files follow black format
+          # 2. Check if C++ files follow clang format
+          run: |
+            cd tensorrt-incubator
+            git config --global --add safe.directory /tensorrt-incubator
+            cat > run_clang_style_check.py <<EOF
+            #!/usr/bin/env python3
```
christopherbate, line 84: I believe this entire script is no longer necessary. I think you can just run the command `git clang-format HEAD~1 --diff`

.github/workflows/mlir-tensorrt-ci.yml
```diff
@@ -0,0 +1,227 @@
...
+        uses: addnab/docker-run-action@v3
+        with:
+          image: ${{ env.selected_test_image }}
+          options: -v ${{ github.workspace }}/mlir-tensorrt:/mlir-tensorrt -v ${{ github.workspace }}/ccache:/ccache -v ${{ github.workspace }}/.ccache.cpm:/.ccache.cpm
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+          run: |
+            export CCACHE_BASEDIR="$PWD"
+            export CCACHE_DIR="$PWD/ccache"
```
christopherbate, line 199: Does this actually match `${{ github.workspace}}/ccache`?

.github/workflows/mlir-tensorrt-ci.yml
shelkesagar29: It was just for test. We can't use GPU's there since driver version is less for CUDA>=12.5 we need.

.github/workflows/mlir-tensorrt-ci.yml
shelkesagar29: This is tag for the docker image that will be created on fly, if any files that affect docker image are changed. In our case, they are the following.
```
- 'mlir-tensorrt/build_tools/docker/Dockerfile'
- 'mlir-tensorrt/python/requirements.txt'
- 'mlir-tensorrt/python/requirements-dev.txt'
```
If above files are changed and we are still testing with `DEFAULT_IMAGE`, there is chance that tests will fail (for example, if new python dependency is added). CI to generate a new `DEFAULT_IMAGE` if changes are in the above file runs after PR is merged in main.

If none of the above files is changed, we use `DEFAULT_IMAGE`.

.github/workflows/mlir-tensorrt-ci.yml
shelkesagar29: It does work since I mount these inside docker before.

https://github.com/NVIDIA/TensorRT-Incubator/actions/runs/10514361515/job/29132010233#step:12:3579

GitHub cache action takes care of lading and saving.
Without cache CI runs take ~1:15 minutes. Mainly because of LLVM build.
With cache CI run takes ~8 minutes e2e.

.github/workflows/mlir-tensorrt-ci.yml
shelkesagar29: Yes. This should do the job. Thanks for pointing out

.github/workflows/mlir-tensorrt-ci.yml
christopherbate: I would avoid rebuilding the entire container just for `requirements.txt` changes. Those you can `pip install` on the fly before running the build.

christopherbate (approved):

.github/workflows/mlir-tensorrt-ci.yml
shelkesagar29: Done.

PR #138, opened 2024-08-21
Add support for stream in Tripy and make execution async by default

parthchadha:


tripy/tripy/backend/compiler_api.py
```diff
@@ -33,6 +33,97 @@
 from tripy.utils import json as json_utils
 
 
+@export.public_api(document_under="compiler")
+class Stream:
+    """
+    Represents CUDA streams that can be used to manage concurrent operations.
```
pmarathe25, line 39:
```suggestion
    Represents a CUDA stream that can be used to manage concurrent operations.
```

tripy/tripy/backend/compiler_api.py
```diff
@@ -33,6 +33,97 @@
...
+
+    This class is a wrapper around the underlying stream object, allowing management of CUDA streams.
+
+    Args:
+        create_new (bool): If True, creates a new stream. If False, uses the default stream.
```
pmarathe25, line 44: Why don't we just provide a `default_stream` variable as part of the API?

tripy/tripy/backend/compiler_api.py
```diff
@@ -33,6 +33,97 @@
...
+
+    .. code-block:: python
+        :linenos:
+        :caption: Create new streams
+
+        streamA = tp.Stream()
```
pmarathe25, line 50: Use snake_case:
```suggestion
        stream_a = tp.Stream()
```

tripy/tripy/backend/compiler_api.py
```diff
@@ -33,6 +33,97 @@
...
+        :caption: Use streams in execution
+
+        linear = tp.Linear(2, 3)
+        compiler = tp.Compiler(linear)
+
+        compiled_linear = compiler.compile(tp.InputInfo((2, 2), dtype=tp.float32))
+
+        a = tp.ones((2, 2), dtype=tp.float32)
+
+        compiled_linear.stream = tp.Stream()
```
pmarathe25, line 67: Can streams not be used in eager mode? I think using a context manager to set the current stream and making the behavior consistent in both eager/compiled mode would be preferable

parthchadha: Updated to make Stream work with context. @jhalakpatel  @yizhuoz004 can you please re-review.

tripy/tripy/backend/compiler_api.py
```diff
@@ -31,6 +32,136 @@
...
+
+@export.public_api(document_under="compiler")
+class Stream:
+    """
+    Represents a CUDA stream that can be used to manage concurrent operations.
+
+    This class is a wrapper around the underlying stream object, allowing management of CUDA streams.
+
+    Args:
+        default_stream (bool): If True, uses the default stream. If False, creates a new stream.
```
pmarathe25, line 46: I meant that there should be a `tp.default_stream` or maybe `tp.Stream.default()` otherwise the semantics of this constructor are confusing.

tripy/tripy/backend/compiler_api.py
```diff
@@ -31,6 +32,136 @@
...
+    @classmethod
+    def get_active_stream(cls):
+        return cls._active_stream
+
+    def cleanup(self):
+        if self in Stream._all_streams:
+            Stream._all_streams.remove(self)
+            del self._stream
+
+    def __del__(self):
```
pmarathe25, line 127: `__del__` is not reliable

tripy/tripy/backend/compiler_api.py
```diff
@@ -31,6 +32,136 @@
...
+    .. code-block:: python
+        :linenos:
+        :caption: Create new streams
+
+        stream_a = tp.Stream()
+        stream_b = tp.Stream()
+        default_stream = tp.Stream(default_stream=True)
+        assert stream_a != stream_b
+        assert default_stream != stream_b
+        stream_a.cleanup()
```
pmarathe25, line 57: Do we have to manually clean up streams? Can't the context manager handle that?

tripy/tripy/backend/compiler_api.py
```diff
@@ -31,6 +32,136 @@
...
+        with tp.Stream():
+            a = tp.ones((2, 2), dtype=tp.float32)
+            out = compiled_linear(a)
+
+        assert cp.array_equal(cp.from_dlpack(out), cp.from_dlpack(linear(a)))
+    """
+
+    _active_stream = None
+    _default_stream = None
+    _all_streams = set()
```
pmarathe25, line 77: Why do we need to keep track of all streams? Wouldn't it greatly simplify this class if we didn't do so?

tripy/tripy/backend/compiler_api.py
parthchadha: This is only in the case where `Stream` is created without context manager. I can also just remove this case and we can rely only on context manager for Stream management.

tripy/tripy/backend/compiler_api.py
parthchadha: `tp.Stream.default_stream()` exists. Let me know if you `default` over `default_stream`?

tripy/tripy/backend/compiler_api.py
parthchadha: It was for syncing all streams but I have removed it from current iteration (we can reintroduce later if required).

tripy/tripy/backend/compiler_api.py
pmarathe25: Let's remove the constructor argument then? No need to have two ways to access the default stream

parthchadha: Fixed in latest iteration.

tripy/tripy/backend/mlir/executor.py
```diff
@@ -178,8 +179,9 @@ def execute(self, output_devices=List[device], inputs: List["Tensor"] = []) -> L
             out_args.append(memref)
 
         # Execute and populate device pointers.
-        self.session.execute_function("main", in_args=in_args, out_args=out_args, stream=self.stream)
-        self.stream.sync()
+        self.session.execute_function(
+            "main", in_args=in_args, out_args=out_args, stream=Stream.get_active_stream()._stream
```
pmarathe25, line 183: Not sure this would work as expected - if we have a long sequence of dependent operations where they are divided across multiple streams, they will all execute on the stream of only the final operation that is evaluated. Somehow we need to associate the stream with the operation at trace time and potentially construct multiple executables.

tripy/tripy/backend/compiler_api.py
```diff
@@ -31,6 +32,91 @@
...
+        self._stream = MLIRRuntimeClient().create_stream()
+        Stream._active_stream = self
+
+    @classmethod
+    def default_stream(cls):
+        """Get the default stream, create it if it doesn't exist."""
+        if cls._default_stream is None:
+            cls._default_stream = cls.__new__(cls)
+            cls._default_stream._stream = MLIRRuntimeClient().create_stream()
+            cls._active_stream = cls._default_stream
```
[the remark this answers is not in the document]
parthchadha: That's right, by default when you ask for default stream it means you want that for current context.

tripy/tripy/backend/mlir/executor.py
parthchadha: Usually, multi-stream is well handled within TRT compiler and we shouldn't worry about assigning streams for independent work in the frontend. (unless the user asks for it which means they are aware of what work they want to enqueue on a stream).

pmarathe25: How does the user ask for it though? If we did:
```py
with tp.Stream() as stream1:
    # do something
    out = ...

with tp.Stream() as stream2:
    out = out + 1
    print(out)
```

Then everything would execute on stream2, even though I expected only the addition on stream2 and everything else on stream1.

tripy/tripy/backend/compiler_api.py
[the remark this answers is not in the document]
pmarathe25: Isn't that what the context manager is for?

tripy/tripy/backend/compiler_api.py
```diff
@@ -31,6 +32,91 @@
...
+
+        if not (hasattr(self, "_stream") and hasattr(other, "_stream")):
+            return False
+
+        return self._stream == other._stream
+
+    def __repr__(self):
+        return f"<Stream(id={id(self)}, default={self == Stream._default_stream})>"
+
+    def __hash__(self):
```
pmarathe25, line 106: Do we need to define hash? If yes, id will not be reliable since different streams could have the same id over the lifetime of the program

tripy/tripy/backend/compiler_api.py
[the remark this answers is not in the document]
parthchadha: Ah yeah, this was added when the previous non-context workflow was also supported. Will remove.

tripy/tripy/backend/mlir/executor.py
parthchadha: We can force the evaluation of `out` in `__exit__` of stream1? Though, I was imagining the use of stream only in case of compiled graphs since in the eager mode streams are kind of useless as TRT would not respect the stream for building the engine.

tripy/tripy/backend/compiler_api.py
parthchadha: Not required anymore. Removed from latest iteration.

tripy/tripy/backend/mlir/executor.py
parthchadha: Let's merge for now since it works with the limited context of `compiled` execution.

yizhuoz004 (approved):

PR #140, opened 2024-08-22
Integrate MLIR-TensorRT internal changes

christopherbate:
This change is a compilation of the following commits:

-- [executor] Enable wider support for loads/stores of aggregates types

  Previously, we only supported load/store of aggregate  types in a very
  limited scope. This change expands that support so that we support load/store
  of arbitrary aggregates. We introduce a pass `executor-decompose-aggregate-loads-and-stores`
  that specifically decomposes these loads/stores into load/stores of the
  individual elements. Since this decomposition results in the creation of
  `executor.offset` operations, we need to sequence the new pass correctly with
  the `executor-expand-ops` pass (which lowers `executor.alloca` and `executor.offset`)
  as well as with the lowering of executor operations into opaque
  `executor.call` ops. To make all this work correctly, this change also
  factors out the latter transformation into a dedicated pass
  `executor-lower-to-runtime-builtins`.

-- NFC: fix some Python typing annotations

-- [executor] add Executor and runtime support for `complex<f32>` and `complex<f64>` types

  This changes:

  - adds support to the Executor dialect for `complex<f32>` and `complex<f64>` types
  - adds support to the runtime API interface for the corresponding `c32` and `c64` types

-- [executor] Properly serialize absent function signatures

  When a `FunctionMetadataAttr` attribute is not provided on a `func.func`
  during translation/serialization to the Executable format, we should pass
  a 0-offset to the signature field when creating a `rt::impl::Function`
  flatbuffer table.

  This was caught when working on complex32/complex64 support. An additional
  validation check is added immediately after finalizing the executable
  buffer in the `mlir::translateToRuntimeExecutable` function.

-- [executor] Fix i4 multiplication runtime error and i4 tests

  Fixes an issue where Lua user type metatable information wasn't correctly
  set for the `nv_int4` type. The 'arithmetic.mlir' i4 tests also were not
  effectively testing the runtime i4 functions because the compiler was
  constant-folding most of the operations being tested. To fix the constant
  folding issue, we just need to pass some arguments to the test functions
  instead of inlining them into the test function bodies.

-- Add bufferization integration test pipeline

  This change adds a simple test bufferization pipeline for the Executor project
  and uses that pipeline to construct new integration tests. The test verifies
  host i4 operations and makes a small bug fix.

-- NFC: move some unit tests from top-level 'test' under the 'executor' sub-project

Signed-off-by: Christopher Bate <cbate@nvidia.com>


shelkesagar29 (approved):
LGTM

PR #148, opened 2024-08-26
Fix docs building

yizhuoz004:


parthchadha (approved):

.github/workflows/tripy-l0.yml
```diff
@@ -46,6 +46,12 @@ jobs:
         image: ${{ env.l0_image }}
         options: --gpus all -v ${{ github.workspace }}/tripy:/tripy
         run: |
+          cat > build_docs_and_test.sh <<EOF
+          #!/bin/bash
+
           python3 docs/generate_rsts.py
-          sphinx-build build/doc_sources build/docs -c docs/ -j 6 -W
+          sphinx-build build/doc_sources build/docs -c docs/ -j 1 -W
           pytest --cov=tripy/ --cov-config=.coveragerc tests/ -v -m "l0 or not l1" --durations=15
```
pranavm-nvidia, line 54: Is there a cleaner way to run these as separate steps? Or some way to indicate that the workflow should fail if any step fails?

yizhuoz004: We are using `addnab/docker-run-action@v3` to run a variable container, the best solution I've seen so far is wrapping the commands into a bash script.

yizhuoz004: In L1 however, we use a static container with `container` tag, so that we can define steps in a job.

PR #149, opened 2024-08-27
[Tripy] Add iterator implementation for `tp.Shape`

slyubomirsky:
Since we are treating `tp.Shape` as a collection, it may also make sense to iterate over them. This PR adds a very simple iterator implementation to `tp.Shape`. Another way we could implement it would be to override `__getitem__` to throw an exception if it gives an out-of-bounds access and then use the default iterator, but this is likely simpler.

### DIFF
--- a/tripy/tests/frontend/test_shape.py
+++ b/tripy/tests/frontend/test_shape.py
@@ -268,6 +268,11 @@ class TestShape:
         s1 = tp.Shape(values)
         assert len(s1[slice_value]) == len(values[slice_value])
 
+    def test_iteration(self, values):
+        s = tp.Shape(values)
+        for i, v in enumerate(s):
+            assert cp.from_dlpack(v).get() == values[i]
+
     def test_reduce(self, values):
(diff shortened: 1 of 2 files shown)


parthchadha (approved):

pranavm-nvidia:
@slyubomirsky wouldn't it be much easier to throw an exception from `__getitem__`?

akhilg-nv (approved):

slyubomirsky:
I guess I can call into the default implementation as well, I'll try it that way.

slyubomirsky:
It seems that the `__getitem__` approach encounters an unexpected complication: `__getitem__` is used for iteration only if there is no implementation of `__iter__`. However, `Shape` would inherit `__iter__` from `Tensor` (the one that throws an exception). I don't see any way to delete the method (`del Shape.__iter__` doesn't work because it's inherited) or force the default implementation. One suggestion I found is to override `__getattr__` which seems like more complexity than we might want.

pranavm-nvidia (approved):

PR #150, opened 2024-08-27
Add tp.flatten

parthchadha:


yizhuoz004:
What is the behavior of flattening a scalar? I noticed torch has a note: `Flattening a zero-dimensional tensor will return a one-dimensional view.` https://pytorch.org/docs/stable/generated/torch.flatten.html

parthchadha:
> What is the behavior of flattening a scalar? I noticed torch has a note: `Flattening a zero-dimensional tensor will return a one-dimensional view.` https://pytorch.org/docs/stable/generated/torch.flatten.html

```py
a = tp.Tensor(1)
a.flatten()
```
```
Invalid dimensions: start_dim=0, end_dim=-1, rank=0.
```

I am not sure if we need to follow torch's convention here.

yizhuoz004:
> > What is the behavior of flattening a scalar? I noticed torch has a note: `Flattening a zero-dimensional tensor will return a one-dimensional view.` https://pytorch.org/docs/stable/generated/torch.flatten.html
> 
> ```python
> a = tp.Tensor(1)
> a.flatten()
> ```
> 
> ```
> Invalid dimensions: start_dim=0, end_dim=-1, rank=0.
> ```
> 
> I am not sure if we need to follow torch's convention here.

We don't have to as long as we give an informative error, flattening a scalar is meaningless IMO.

PR #151, opened 2024-08-27
[Tripy][Docs] Recommend the use of `--pull always` to update the image in the contributing doc

slyubomirsky:
`docker run` by default will only pull an image if it is entirely missing, so the command given in `CONTRIBUTING.md` will not update the image if it had been pulled before. This change adds the `--pull always` setting in the example command so that it would check for an update before running.

### DIFF
--- a/tripy/CONTRIBUTING.md
+++ b/tripy/CONTRIBUTING.md
@@ -31,6 +31,8 @@ Thank you for considering contributing to Tripy!
         docker run --gpus all -it -v $(pwd):/tripy/ --rm ghcr.io/nvidia/tensorrt-incubator/tripy
         ```
 
+        If you have pulled the image in the past and need to update it, you should add `--pull=always` to the above command or separately call `docker pull ghcr.io/nvidia/tensorrt-incubator/tripy`.
+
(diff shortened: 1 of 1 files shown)


parthchadha (approved):

tripy/CONTRIBUTING.md
```diff
@@ -31,6 +31,8 @@ Thank you for considering contributing to Tripy!
         docker run --gpus all -it -v $(pwd):/tripy/ --rm ghcr.io/nvidia/tensorrt-incubator/tripy
         ```
 
+        If you have pulled the image in the past and need to update it, you should add `--pull=always` to the above command or separately call `docker pull ghcr.io/nvidia/tensorrt-incubator/tripy`.
```
pranavm-nvidia, line 34: Let's just make this the default and add `--pull=always` to the code block above? If an older container was required, we'd be building it locally in most cases.

slyubomirsky: Good idea. Also the equals sign is wrong, so I'll take that out.

pranavm-nvidia (approved):

PR #167, opened 2024-08-30
[Tripy] Override `__mul__` for `tp.Shape`s to work like `__mul__` for Python lists

slyubomirsky:
Since we are treating `tp.Shape` more like a `Sequence` than a `Tensor`, we can treat `*` similarly to `+` and have a separate `multiply` to do elementwise multiplication while treating `*` as tiling.

This PR also improves the doc comment for `Shape`, so that the generated doc page clearly explains how `+` and `*` work differently compared to `Tensor`.

### DIFF
--- a/tripy/tripy/frontend/shape.py
+++ b/tripy/tripy/frontend/shape.py
@@ -30,7 +30,7 @@ class Shape(Tensor):
     A Shape is a tensor used to represent a tensor shape.
     Shapes are vectors (rank 1) of non-negative integers (using int32 as the datatype).
 
-    Note that Shapes are intended to used in many cases like Python lists, hence `+` acts as concatenation
+    Note that Shapes are intended to be used in many cases like Python lists; hence `+` acts as concatenation
     on Shapes rather than elementwise addition and `*` acts as tiling rather than elementwise multiplication;
     the methods `add` and `multiply` can be used for elementwise addition and multiplication, respectively.
     Additionally, `len` can be used to get the length of a shape.
--- a/tripy/tripy/frontend/shape.py
+++ b/tripy/tripy/frontend/shape.py
@@ -29,6 +29,11 @@ class Shape(Tensor):
     """
     A Shape is a tensor used to represent a tensor shape.
     Shapes are vectors (rank 1) of non-negative integers (using int32 as the datatype).
+
+    Note that Shapes are intended to used in many cases like Python lists, hence `+` acts as concatenation
+    on Shapes rather than elementwise addition and `*` acts as tiling rather than elementwise multiplication;
+    the methods `add` and `multiply` can be used for elementwise addition and multiplication, respectively.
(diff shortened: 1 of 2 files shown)


tripy/tripy/frontend/shape.py
```diff
@@ -149,6 +177,38 @@ def __radd__(self, other):
             other = Shape(other)
         return concatenate([other, self], 0)
 
+    # multiplication for shapes is tiling, not elementwise multiplication
+
+    def __mul__(self, other):
```
pranavm-nvidia, line 182: Are there calls to this that were relying on the old behavior of `__mul__`?

slyubomirsky: I don't believe so, other than those in the tests updated here. I've searched for all appearances of ` * ` in the Tripy codebase and they don't seem to be used with shapes. The CI will tell us in either case.

As with our handling of `+`, this violates the Liskov substitution principle, so that's a hazard of inheritance here 😅

tripy/tripy/frontend/shape.py
```diff
@@ -29,6 +29,11 @@ class Shape(Tensor):
     """
     A Shape is a tensor used to represent a tensor shape.
     Shapes are vectors (rank 1) of non-negative integers (using int32 as the datatype).
+
+    Note that Shapes are intended to used in many cases like Python lists, hence `+` acts as concatenation
```
pranavm-nvidia, line 33:
```suggestion
    Note that Shapes are intended to be used in many cases like Python lists; hence `+` acts as concatenation
```

tripy/tripy/frontend/shape.py
pranavm-nvidia, line 182: Do we also need to override the reverse ops?

tripy/tripy/frontend/shape.py
```diff
@@ -149,6 +177,38 @@ def __radd__(self, other):
...
+        # Only defined with a scalar argument
+        if not isinstance(other, Tensor):
+            other = Tensor(other, dtype=int32)
+        if other.rank >= 1:
+            raise_error(
+                "Attempting to multiply a Tripy Shape by a tensor of rank >= 1, which is undefined", details=[other]
+            )
+        # note: in Python, if a list is multiplied by a negative number, this is the same as multiplying by 0,
+        # so we should clamp the argument
+        other = where(other >= 0, other, Tensor(0))
```
pranavm-nvidia, line 201: Can we use `max(other, 0)` here?

pranavm-nvidia (approved):

tripy/tripy/frontend/shape.py
slyubomirsky: Slipped my mind, that's simpler.

pranavm-nvidia (approved):

parthchadha (approved):

PR #169, opened 2024-08-30
Add test case to check if all ops are being verified

Mgluhovskoi:
Added one test case that checks if all ops have their data types verified. I also added a list where functions can be excluded if needed.

### DIFF
--- a/tripy/tests/spec_verification/test_dtype_constraints.py
+++ b/tripy/tests/spec_verification/test_dtype_constraints.py
@@ -22,7 +22,7 @@ from tripy.common.datatype import DATA_TYPES
 import itertools
 import pytest
 from tests.spec_verification.object_builders import create_obj
-from tripy.constraints import TYPE_VERIFICATION, RETURN_VALUE
+from tripy.constraints import TYPE_VERIFICATION, RETURN_VALUE, FUNC_W_DOC_VERIF
 import tripy as tp
 from contextlib import ExitStack
 
@@ -181,3 +181,21 @@ def test_neg_dtype_constraints(test_data):
         api_call_locals, namespace = _run_dtype_constraints_subtest(test_data)
         if isinstance(api_call_locals[RETURN_VALUE], tp.Tensor):
             assert api_call_locals[RETURN_VALUE].dtype == namespace[return_dtype]
+
+
+operations = [obj.__qualname__ for _, obj in inspect.getmembers(tp) if inspect.isfunction(obj)]
+# add any function that you do not want to be verified:
+func_exceptions = ["plugin", "dequantize"]
+
+
+# Check if there are any operations that are not included (Currently does not test any __<op>__ functions)
+@pytest.mark.parametrize("func_qualname", operations, ids=lambda val: f"is_{val}_verified")
+def test_all_ops_verified(func_qualname):
+    if func_qualname in FUNC_W_DOC_VERIF:
+        return
+    elif func_qualname.startswith("_") or func_qualname in func_exceptions:
+        return
+    else:
+        pytest.fail(
+            f"function {func_qualname}'s data types have not been verified. Please add data type verification by following the guide within tripy/tests/spec_verification or exclude it from this test."
+        )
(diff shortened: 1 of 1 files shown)


Mgluhovskoi:
I was able to implement a test case that checks if most of the operations are being verified or not, but I am not sure how to get a list of all __<op>__s. I tried to use inspect.getmembers, but they don't show up here. Any recommendations?

tripy/tests/spec_verification/test_dtype_constraints.py
```diff
@@ -181,3 +181,21 @@ def test_neg_dtype_constraints(test_data):
         api_call_locals, namespace = _run_dtype_constraints_subtest(test_data)
         if isinstance(api_call_locals[RETURN_VALUE], tp.Tensor):
             assert api_call_locals[RETURN_VALUE].dtype == namespace[return_dtype]
+
+
+operations = [obj.__qualname__ for _, obj in inspect.getmembers(tp) if inspect.isfunction(obj)]
```
pranavm-nvidia, line 186: Could you use `tests.helper.get_all_tripy_interfaces` here?

tripy/tests/spec_verification/test_dtype_constraints.py
```diff
@@ -181,3 +181,21 @@ def test_neg_dtype_constraints(test_data):
...
+@pytest.mark.parametrize("func_qualname", operations, ids=lambda val: f"is_{val}_verified")
+def test_all_ops_verified(func_qualname):
+    if func_qualname in FUNC_W_DOC_VERIF:
+        return
+    elif func_qualname.startswith("_") or func_qualname in func_exceptions:
+        return
+    else:
+        pytest.fail(
+            f"function {func_qualname}'s data types have not been verified. Please add data type verification by following the guide within tripy/tests/spec_verification or exclude it from this test."
+        )
```
pranavm-nvidia, line 201:
```suggestion
    if not func_qualname.startswith("_") and func_qualname not in func_exceptions:
        assert func_qualname in FUNC_W_DOC_VERIF, f"function {func_qualname}'s data types have not been verified. Please add data type verification by following the guide within tripy/tests/spec_verification or exclude it from this test."
```

tripy/tests/spec_verification/test_dtype_constraints.py
Mgluhovskoi: It seems like `tests.helper.get_all_tripy_interfaces` doesn't work anymore

pranavm-nvidia: Could you elaborate? We use it for our docstring testing, so it should still be working.

Mgluhovskoi: I keep getting the following error when calling it:

`test_dtype_constraints.py:29: in <module>
    print(get_all_tripy_interfaces())
../../tripy/common/datatype.py:48: in __repr__
    return cls.name
E   AttributeError: type object 'BaseDtype' has no attribute 'name`

Mgluhovskoi: Oh this is an issue do to me trying to print the result.

pranavm-nvidia: I'll fix that in https://github.com/NVIDIA/TensorRT-Incubator/pull/170

tripy/tests/spec_verification/test_dtype_constraints.py
```diff
@@ -181,3 +181,63 @@ def test_neg_dtype_constraints(test_data):
         api_call_locals, namespace = _run_dtype_constraints_subtest(test_data)
         if isinstance(api_call_locals[RETURN_VALUE], tp.Tensor):
             assert api_call_locals[RETURN_VALUE].dtype == namespace[return_dtype]
+
+
+def get_all_possible_verif_ops():
```
pranavm-nvidia, line 186: Another approach could be to check all public APIs, and then for classes, check their methods. We can then filter those on whether any of the type annotations include `tp.Tensor`.

pranavm-nvidia:
Adding this in https://github.com/NVIDIA/TensorRT-Incubator/pull/274

PR #170, opened 2024-08-30
Adds `repeat` API, various other fixes

pranavm-nvidia:
- Updates unsqueeze test to check that shape is correct

- Updates docstrings in some functions to print outputs, disables `bool` in allclose

- Miscellaneous bug fixes in various operations
    
- Adds a `tp.repeat` API. The API is currently constrained to accepting only a single integer value for
     `repeats` since this greatly simplifies the implementation.
    
- Introduces a `process_dim` decorator which handles negative dimensions in API functions.
    
- Various minor fixes to dtype constraint checking. Also merges the positive/negative
     dtype constraints test into one so it's easier to run all the tests for a given op.

### DIFF
--- a/tripy/tripy/frontend/ops/repeat.py
+++ b/tripy/tripy/frontend/ops/repeat.py
@@ -1,0 +1,78 @@
+from tripy import constraints, export
+from tripy.common.exception import raise_error
+from tripy.frontend import utils as frontend_utils
+
+
+@export.public_api(document_under="operations/functions")
+@constraints.dtype_info(
+    dtype_variables={
+        "T1": ["float32", "float16", "bfloat16", "float8", "int8", "int32", "int64", "bool"],
+    },
+    dtype_constraints={"input": "T1", constraints.RETURN_VALUE: "T1"},
+)
+@frontend_utils.process_dim
+def repeat(input: "tripy.Tensor", repeats: int, dim: int) -> "tripy.Tensor":
+    """
+    Repeats each element of a tensor after itself along the specified dimension.
+
+    Args:
+        input: The input tensor.
+        repeats: The number of times to repeat each element.
+        dim: The dimension along which to repeat values.
+
+    Returns:
+        The new tensor.
+
+    .. code-block:: python
+        :linenos:
+        :caption: 1D tensor
+
+        inp = tp.arange(4, dtype=tp.int32)
+        out0 = tp.repeat(inp, 2, dim=0)
+
+        np_inp = np.from_dlpack(tp.copy(inp, device=tp.device("cpu"))) # doc: omit
+        ref_out0 = np.repeat(np_inp, 2, 0) # doc: omit
+        assert np.array_equal(ref_out0, np.from_dlpack(tp.copy(out0, device=tp.device("cpu"))))
+
+
+    .. code-block:: python
+        :linenos:
+        :caption: 2D tensor
+
+        inp = tp.reshape(tp.arange(4, dtype=tp.int32), (2, 2))
+        out0 = tp.repeat(inp, 2, dim=0)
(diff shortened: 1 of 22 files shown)


tripy/tripy/frontend/ops/repeat.py
```diff
@@ -0,0 +1,79 @@
...
+    #
+    # [1, 2] -> [[1],  -> [[1, 1],  -> [1, 1, 2, 2]
+    #            [2],]     [2, 2],]
+    #
+    out = unsqueeze(input, dim + 1)
+    out = expand(out, input.shape[: dim + 1] + [repeats] + input.shape[dim + 1 :])
+
+    repeat_mask = [1] * input.rank
+    repeat_mask[dim] = repeats
+    # TODO: Change to `.multiply()`
```
pranavm-nvidia, line 76: Need to change this once https://github.com/NVIDIA/TensorRT-Incubator/pull/167 merges

tripy/tripy/utils/utils.py
```diff
@@ -468,3 +468,29 @@ def merge_function_arguments(func, *args, **kwargs):
     all_args = get_positional_arg_names(func, *args)
     all_args.extend(kwargs.items())
     return all_args
+
+
+def get_arg_by_name(name, func, *args, **kwargs):
```
parthchadha, line 473: Can you add some docstrings for both `get_arg_by_name` and `modify_arg` (makes it easier to reuse these utils).

tripy/tripy/frontend/utils.py
```diff
@@ -297,3 +297,35 @@ def topological_sort(ops: List[Union[BaseTraceOp, BaseFlatIROp]]) -> List[Union[
...
+# NOTE: This is currently an extremely specialized decorator that expects an `input` tensor
+# argument and a `dim` argument for the axis.
+# In the future, we can generalize this and use it more broadly.
+def process_dim(func):
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        input = utils.get_arg_by_name("input", func, *args, **kwargs)
+
+        def process_dim(dim: int) -> int:
+            new_dim = dim
```
parthchadha, line 314: Maybe add an error if dim is not python integer?

pranavm-nvidia: We already have type checking in the frontend as a consequence of our function registry:
```
Could not find an implementation for function: 'repeat'.
    Note: Argument types were: [Tensor, int, dim=Tensor].
    Candidate overloads were:

    --> /tripy/tripy/frontend/utils.py:6 in repeat()
          |
        6 | @export.public_api(document_under="operations/functions")
        7 | @constraints.dtype_info(
        8 |     dtype_variables={
        9 |     ...
          | 

    Not a valid overload because: For parameter: 'dim', expected an instance of type: 'int' but got argument of type: 'Tensor'.
```

tripy/tripy/frontend/ops/repeat.py
slyubomirsky: It's in

tripy/tripy/frontend/ops/repeat.py
```diff
@@ -0,0 +1,79 @@
...
+    new_shape = input.shape * repeat_mask
```
slyubomirsky, line 77: Is `new_shape` the same as `input.shape[: dim + 1] + [repeats] + input.shape[dim + 1 :]` above? We can just assign that to a variable and reuse it in that case.

tripy/tripy/utils/utils.py
pranavm-nvidia: Added to `modify_arg`. I think `get_arg_by_name` is fairly self-explanatory.

PR #173, opened 2024-09-03
Re-introduce tp.Stream

parthchadha:
Changes in this MR:

1. Adds `tp.Stream` interface. (supports streams with priority 0 as of now). Allows getting default stream.
2. Tripy execution will be async by default and user is expected to call `stream.synchronize()`.
3. In case the user does not set stream for execution, Tripy will use default stream. The user is still expected to call synchronize and can call synchronize on the class which invokes sync on default stream. (ie. `tp.synchronize()`.

### DIFF
--- a/tripy/tests/backend/test_stream.py
+++ b/tripy/tests/backend/test_stream.py
@@ -1,0 +1,61 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import pytest
+import gc, sys
+import tripy as tp
+import cupy as cp
+
+
+def test_default_stream_creation():
+    default_stream1 = tp.Stream.get_default_stream()
+    default_stream2 = tp.Stream.get_default_stream()
+    assert default_stream1 == default_stream2
+
+
+def test_new_streams():
+    stream1 = tp.Stream()
+    stream2 = tp.Stream()
+    assert stream1 != stream2
+    assert stream1 != tp.Stream().get_default_stream()
+
+
+def test_enqueue_work_on_stream():
+    linear = tp.Linear(25, 30)
+    compiler = tp.Compiler(linear)
+
+    compiled_linear = compiler.compile(tp.InputInfo((2, 25), dtype=tp.float32))
+
+    a = tp.ones((2, 25), dtype=tp.float32)
+
+    stream = tp.Stream()
+    compiled_linear.stream = stream
+    out = compiled_linear(a)
+    # stream sync below is not required since from_dlpack method will eval() the tensor which will call stream sync anyway.
+    compiled_linear.stream.synchronize()
+    assert cp.array_equal(cp.from_dlpack(out), cp.from_dlpack(linear(a)))
--- a/tripy/tests/backend/test_stream.py
+++ b/tripy/tests/backend/test_stream.py
@@ -12,20 +12,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-# SPDX-License-Identifier: Apache-2.0
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-# http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
 import pytest
 import gc, sys
 import tripy as tp
@@ -33,8 +19,8 @@ import cupy as cp
 
 
 def test_default_stream_creation():
-    default_stream1 = tp.Stream.get_default_stream()
-    default_stream2 = tp.Stream.get_default_stream()
+    default_stream1 = tp.default_stream()
+    default_stream2 = tp.default_stream()
     assert default_stream1 == default_stream2
 
 
@@ -42,7 +28,7 @@ def test_new_streams():
     stream1 = tp.Stream()
     stream2 = tp.Stream()
     assert stream1 != stream2
-    assert stream1 != tp.Stream().get_default_stream()
+    assert stream1 != tp.default_stream()
 
 
 def test_enqueue_work_on_stream():
@@ -52,6 +38,10 @@ def test_enqueue_work_on_stream():
     compiled_linear = compiler.compile(tp.InputInfo((2, 25), dtype=tp.float32))
 
     a = tp.ones((2, 25), dtype=tp.float32)
+
+    out = compiled_linear(a)
+    tp.default_stream().synchronize()
+    assert cp.array_equal(cp.from_dlpack(out), cp.from_dlpack(linear(a)))
 
     stream = tp.Stream()
     compiled_linear.stream = stream
--- a/tripy/tripy/backend/compiler_api.py
+++ b/tripy/tripy/backend/compiler_api.py
@@ -15,7 +15,6 @@
 # limitations under the License.
 #
 
-import atexit
 import base64
 import inspect
 import numbers
@@ -32,6 +31,49 @@ from tripy.common.exception import raise_error
 from tripy.common.shape_bounds import ShapeBounds
 from tripy.frontend import Tensor, Trace
 from tripy.utils import json as json_utils
+from tripy.common.device import device as Device
+
+
+# Global variable to store instances
+_default_stream_instances = {}
+
+
+@export.public_api(document_under="compiler")
+def default_stream(device: Device = Device("gpu")) -> "tripy.Stream":
+    """
+    Provides access to the default CUDA stream for a given device.
+    This function implements singleton pattern to ensure a single default stream instance per device.
+
+    Args:
+        device: The device for which to get the default stream.
+
+    Returns:
+        The default Stream object for the specified device.
+
+    Raises:
+        ValueError: If the device is not of type 'gpu' or if the device index is not 0.
+
+    .. code-block:: python
+        :linenos:
+        :caption: Creation of default_stream
+
+        default = default_stream()  # Returns the default Stream object for the current device
+
+    Note:
+        Calling default_stream() with the same device always returns the same Stream instance for that device.
+    """
+    global _default_stream_instances
+
+    if device.kind != "gpu":
+        raise_error(f"default_stream creation requires device to be of type gpu, got device={device}.")
+
+    if device.index != 0:
+        raise_error(f"Tripy stream only works with device index 0, got device={device}")
+
+    if device.index not in _default_stream_instances:
+        _default_stream_instances[device.index] = Stream()
+
+    return _default_stream_instances[device.index]
 
 
 @export.public_api(document_under="compiler")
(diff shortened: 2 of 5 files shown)


tripy/tripy/backend/compiler_api.py
```diff
@@ -33,6 +34,108 @@
...
+
+    @classmethod
+    def get_default_stream(cls):
+        """Get the default stream, create it if it doesn't exist."""
+        if cls._default_stream is None:
+            cls._default_stream = cls()
+        return cls._default_stream
+
+    @classmethod
+    def synchronize(cls):
```
parthchadha, line 95: Fix this, can't have the same method name for class and instance.

tripy/tests/backend/test_stream.py
```diff
@@ -0,0 +1,51 @@
...
+def test_enqueue_work_on_stream():
+    linear = tp.Linear(25, 30)
+    compiler = tp.Compiler(linear)
+
+    compiled_linear = compiler.compile(tp.InputInfo((2, 25), dtype=tp.float32))
+
+    a = tp.ones((2, 25), dtype=tp.float32)
+
+    out = compiled_linear(a)
+    tp.Stream.synchronize()
```
pranavm-nvidia, line 43: I think requiring explicit synchronization for an implicit stream will be extremely confusing/error prone. Can we run in synchronous mode by default?

tripy/tripy/backend/compiler_api.py
```diff
@@ -33,6 +34,108 @@
...
+
+    def __hash__(self):
+        return hash(id(self))
+
+    @classmethod
+    def cleanup_default_stream(cls):
+        if cls._default_stream:
+            # TODO: expose methods from mlir-tensorrt python bindings to destroy stream
+            cls._default_stream = None
+        cls._active_stream = None
```
pranavm-nvidia, line 129: I don't think this is needed

tripy/tripy/backend/compiler_api.py
```diff
@@ -33,6 +34,108 @@
...
+            cls._default_stream != None
+        ), f"Expect default stream to be created before calling tp.Stream.synchronize()"
+        cls._default_stream.synchronize()
+
+    def synchronize(self):
+        """Synchronize the stream, blocking until all operations in this stream are complete."""
+        self._active_cuda_stream.sync()
+
+    def __del__(self):
+        del self._active_cuda_stream
```
pranavm-nvidia, line 107: Should not be needed
```suggestion
```

tripy/tripy/backend/compiler_api.py
parthchadha: Do you mean the method is not required or setting the active_stream to None?

tripy/tripy/backend/compiler_api.py
```diff
@@ -33,6 +34,108 @@
...
+
+    @classmethod
+    def synchronize(cls):
+        """Synchronize the default stream, blocking until all operations in this stream are complete."""
+        assert (
+            cls._default_stream != None
+        ), f"Expect default stream to be created before calling tp.Stream.synchronize()"
+        cls._default_stream.synchronize()
+
+    def synchronize(self):
```
pranavm-nvidia, line 102:
```suggestion
    @staticmethod
    def synchronize(self):
```

tripy/tripy/backend/compiler_api.py
pranavm-nvidia: Method is not required, as this will happen on program exit anyway.

parthchadha: We might need this to correctly destroy stream object created in `mlir-tensorrt`.

tripy/tests/backend/test_stream.py
parthchadha: Tripy will be async by default and the method for syncing default stream is now renamed to `tp.synchronize()`

tripy/tripy/backend/compiler_api.py
```diff
@@ -33,6 +33,110 @@
...
 
 
+@export.public_api(document_under="compiler")
+def synchronize():
+    """
+    Synchronizes the default stream.
+
+    .. code-block:: python
+        :linenos:
+        :caption: Use streams in execution
```
pranavm-nvidia, line 43: Since there's only one:
```suggestion
        :caption: Example
```

tripy/tripy/backend/compiler_api.py
```diff
@@ -33,6 +33,110 @@
...
+        compiled_linear = compiler.compile(tp.InputInfo((2, 2), dtype=tp.float32))
+
+        a = tp.ones((2, 2), dtype=tp.float32)
+
+        compiled_linear.stream = tp.Stream()
+        out = compiled_linear(a)
+        assert cp.array_equal(cp.from_dlpack(out), cp.from_dlpack(linear(a)))
+    """
+
+    _default_stream = None  # class variable
```
pranavm-nvidia, line 95: Why isn't this just a global?

tripy/tripy/backend/compiler_api.py
```diff
@@ -33,6 +33,110 @@
...
+    Represents a CUDA stream that can be used to manage concurrent operations.
+
+    This class is a wrapper around the underlying stream object, allowing management of CUDA streams.
+
+    Args:
+        priority (int): Assign priority for the new stream. Lower number signifies higher priority.
+
+    .. code-block:: python
+        :linenos:
+        :caption: Create new streams
```
pranavm-nvidia, line 71:
```suggestion
        :caption: Creating new streams
```

tripy/tripy/backend/compiler_api.py
```diff
@@ -33,6 +33,110 @@
...
+
+        streamA = tp.Stream()
+        streamB = tp.Stream()
+        defaultStream = tp.Stream.get_default_stream()
+        assert streamA != streamB
+        assert defaultStream != streamB
+
+    .. code-block:: python
+        :linenos:
+        :caption: Use streams in execution
```
pranavm-nvidia, line 81:
```suggestion
        :caption: Using streams in execution
```

tripy/tripy/backend/compiler_api.py
```diff
@@ -33,6 +33,110 @@
...
+        a = tp.ones((2, 2), dtype=tp.float32)
+
+        compiled_linear.stream = tp.Stream()
+        out = compiled_linear(a)
+        assert cp.array_equal(cp.from_dlpack(out), cp.from_dlpack(linear(a)))
+    """
+
+    _default_stream = None  # class variable
+
+    _active_cuda_stream = None
```
pranavm-nvidia, line 97: Should this be in the constructor instead?

tripy/tripy/backend/compiler_api.py
parthchadha: Isn't it much cleaner for it to be part of the Stream class? What's the advantage of making it global variable?

pranavm-nvidia: Global is easier to reason about IMO. A default stream is just a `Stream` that's created and available by default in Tripy, i.e. a global of type `Stream`. Making `Stream` contain a `Stream` as a class variable doesn't really align with that mental model. 

Besides, since you have:
```py
# Create default stream.
Stream.get_default_stream()
```
in the global scope, the default stream effectively is a global anyway. 

Even in PyTorch, the [default stream](https://pytorch.org/docs/stable/generated/torch.cuda.default_stream.html) is separated from the `Stream` class.

tripy/tripy/backend/compiler_api.py
```diff
@@ -33,6 +33,130 @@
...
+        if priority != 0:
+            raise_error(
+                "Incorrect stream priority",
+                [f"Stream priority can only be 0 until #172 is fixed, got priority={priority}."],
+            )
+        from tripy.backend.mlir.utils import MLIRRuntimeClient
+
+        self._active_cuda_stream = MLIRRuntimeClient().create_stream()
+
+    def synchronize(self):
```
yizhuoz004, line 140: Needs return type. Please check docs can be built successfully.

tripy/tripy/backend/compiler_api.py
```diff
@@ -33,6 +33,130 @@
...
+        if device is None:
+            device = 0
+
+        if device != 0:
+            raise_error(f"Tripy stream only works with device index 0, got device={device}")
+
+        if device not in cls._instances:
+            instance = super().__new__(cls)
+            instance.stream = Stream()
+            cls._instances[device] = instance.stream
```
pranavm-nvidia, line 66: Why don't we just do:
```py
cls._instances[device] = Stream()
```
?

tripy/tripy/backend/compiler_api.py
```diff
@@ -33,6 +33,130 @@
...
+        default = default_stream()  # Returns the default Stream object for the current device
+        default_gpu1 = default_stream(1)  # Returns the default Stream object for GPU 1
+
+    Note:
+        Calling default_stream() with the same device always returns the same Stream instance for that device.
+    """
+
+    _instances = {}
+
+    def __new__(cls, device=None):
```
pranavm-nvidia, line 56: Should we use `tp.device` for consistency? We can require it to be a GPU device.

tripy/tripy/backend/compiler_api.py
parthchadha: Fixed.

tripy/tripy/backend/compiler_api.py
parthchadha: Fixed.

tripy/tripy/backend/compiler_api.py
```diff
@@ -31,6 +31,130 @@
...
+
+    def __new__(cls, device: Device = Device("gpu")):
+        if device.kind == "cpu":
+            raise_error(f"default_stream creation requires device to be of type gpu, got device={device}.")
+
+        if device.index != 0:
+            raise_error(f"Tripy stream only works with device index 0, got device={device}")
+
+        if device.index not in cls._instances:
+            instance = super().__new__(cls)
```
pranavm-nvidia, line 65: What's the purpose of `instance` here?

tripy/tripy/backend/compiler_api.py
```diff
@@ -31,6 +31,130 @@
 from tripy.common.shape_bounds import ShapeBounds
 from tripy.frontend import Tensor, Trace
 from tripy.utils import json as json_utils
+from tripy.common.device import device as Device
+
+
+# Create default stream.
+@export.public_api(document_under="compiler")
+class default_stream:
```
pranavm-nvidia, line 39: This might actually be better as a function. Right now it's a class whose constructor returns an object of a different type than an instance of the class.

parthchadha: Moved to a function

tripy/tripy/backend/compiler_api.py
```diff
@@ -31,6 +31,169 @@
...
+        a = tp.ones((2, 2), dtype=tp.float32)
+
+        compiled_linear.stream = default_stream
+        out = compiled_linear(a)
+        default_stream.synchronize()
+        assert cp.array_equal(cp.from_dlpack(out), cp.from_dlpack(linear(a)))
+
+    Usage:
+        default = default_stream()  # Returns the default Stream object for the current device
+        default_gpu1 = default_stream(tp.device("gpu", 1))  # Returns the default Stream object for GPU 1
```
pranavm-nvidia, line 75: I think this should just be the example. We don't necessarily need to show how to assign a stream to an executable here, since that's already covered in `Stream`.

tripy/tripy/backend/compiler_api.py
```diff
@@ -31,6 +31,169 @@
...
+
+
+@export.public_api(document_under="compiler")
+def default_stream(device: Device = Device("gpu")) -> "Stream":
+    """
+    Provides access to the default CUDA stream for a given device.
+    This function implements singleton pattern to ensure a single default stream instance per device.
+
+    Args:
+        device (Device, optional): The device for which to get the default stream. Defaults to Device("gpu").
```
pranavm-nvidia, line 48: Type annotations and defaults are automatically documented:
```suggestion
        device: The device for which to get the default stream. 
```

tripy/tripy/backend/compiler_api.py
```diff
@@ -31,6 +31,169 @@
...
+
+    Returns:
+        Stream: The default Stream object for the specified device.
```
pranavm-nvidia, line 51:
```suggestion
        The default Stream object for the specified device.
```

tripy/tripy/backend/compiler_api.py
```diff
@@ -31,6 +31,169 @@
...
 from tripy.utils import json as json_utils
+from tripy.common.device import device as Device
+
+
+# Global variable to store instances
+_default_stream_instances = {}
+
+
+@export.public_api(document_under="compiler")
+def default_stream(device: Device = Device("gpu")) -> "Stream":
```
pranavm-nvidia, line 42:
```suggestion
def default_stream(device: Device = Device("gpu")) -> "tripy.Stream":
```

tripy/tripy/backend/compiler_api.py
```diff
@@ -31,6 +31,169 @@
...
+        default = default_stream()  # Returns the default Stream object for the current device
+        default_gpu1 = default_stream(tp.device("gpu", 1))  # Returns the default Stream object for GPU 1
+
+    Note:
+        Calling default_stream() with the same device always returns the same Stream instance for that device.
+    """
+    global _default_stream_instances
+
+    if device.kind != "gpu":
+        raise ValueError(f"default_stream creation requires device to be of type gpu, got device={device}.")
```
pranavm-nvidia, line 83: Why don't we use `raise_error` here and below?

tripy/tripy/backend/compiler_api.py
```diff
@@ -31,6 +31,169 @@
...
+
+        linear = tp.Linear(2, 3)
+        compiler = tp.Compiler(linear)
+
+        compiled_linear = compiler.compile(tp.InputInfo((2, 2), dtype=tp.float32))
+
+        a = tp.ones((2, 2), dtype=tp.float32)
+
+        out = compiled_linear(a)
+        tp.synchronize()
```
pranavm-nvidia, line 111: Maybe we should show a perf benchmark here since that's probably the only case where you'd manually synchronize?

tripy/tripy/backend/compiler_api.py
```diff
@@ -31,6 +31,169 @@
...
+        compiler = tp.Compiler(linear)
+
+        compiled_linear = compiler.compile(tp.InputInfo((2, 2), dtype=tp.float32))
+
+        a = tp.ones((2, 2), dtype=tp.float32)
+
+        compiled_linear.stream = tp.Stream()
+        out = compiled_linear(a)
+        assert cp.array_equal(cp.from_dlpack(out), cp.from_dlpack(linear(a)))
+    """
```
pranavm-nvidia, line 151: This part of the docstring should be under the constructor to make it render correctly. Also, we can drop the type annotations

tripy/tripy/backend/compiler_api.py
```diff
@@ -31,6 +31,169 @@
...
+    if device.index not in _default_stream_instances:
+        _default_stream_instances[device.index] = Stream()
+
+    return _default_stream_instances[device.index]
+
+
+@export.public_api(document_under="compiler")
+def synchronize() -> None:
+    """
+    Synchronizes the default stream.
```
pranavm-nvidia, line 97: I think this should synchronize all streams (that's the behavior in torch)? Otherwise it's not really adding any value - we can just do `default_stream().synchronize()`. 
Also, this won't work for multi-device. Maybe just remove it for now?

tripy/tripy/backend/compiler_api.py
```diff
@@ -31,6 +31,169 @@
...
+
+            linear = tp.Linear(2, 3)
+            compiler = tp.Compiler(linear)
+
+            compiled_linear = compiler.compile(tp.InputInfo((2, 2), dtype=tp.float32))
+
+            a = tp.ones((2, 2), dtype=tp.float32)
+            compiled_linear.stream = tp.Stream()
+            out = compiled_linear(a)
+            compiled_linear.stream.synchronize()
```
pranavm-nvidia, line 178: Maybe show how to benchmark here? Otherwise the example is basically the same as what you have in the constructor.

tripy/tripy/backend/compiler_api.py
```diff
@@ -31,6 +31,169 @@
...
+        if not (hasattr(self, "_active_cuda_stream") and hasattr(other, "_active_cuda_stream")):
+            return False
+
+        return self._active_cuda_stream == other._active_cuda_stream
+
+    def __repr__(self):
+        return f"<Stream(id={id(self)})>"
+
+    def __hash__(self):
+        return hash(id(self))
```
pranavm-nvidia, line 196: Not sure you need any of these

pranavm-nvidia (approved):

tripy/tripy/backend/compiler_api.py
parthchadha: `__hash__` might not be required right now but I think its fine to keep it. `__str__` and `__eq__` are required.

tripy/tripy/backend/compiler_api.py
parthchadha: Removing for now but I think there is value in it (will file issue to bring it back).

tripy/tripy/backend/compiler_api.py
parthchadha: Ok, this as an example was a little too short :)

PR #182, opened 2024-09-05
Fix debug opts

yizhuoz004:


tripy/tripy/backend/mlir/compiler.py
```diff
@@ -54,18 +54,20 @@ def __init__(self, trt_builder_opt_level=0) -> None:
...
+        ]
         if config.enable_mlir_debug or config.enable_tensorrt_debug:
-            opts.set_debug_options(
-                config.enable_mlir_debug,
-                config.mlir_debug_types if config.enable_mlir_debug else [],
-                config.mlir_debug_tree_path if config.enable_mlir_debug else None,
-                config.tensorrt_debug_path if config.enable_tensorrt_debug else None,
-            )
-        return opts
+            opts.append("--debug=true")
```
parthchadha, line 63: Why are we always enabling debug flag? What does debug=True mean?

yizhuoz004: Not always, only if either `TRIPY_MLIR_DEBUG_ENABLED` or `TRIPY_TRT_DEBUG_ENABLED` is set.

parthchadha: Ah, missed the unchanged if condition. What does debug=True do in `mlir-tensorrt`? There is already debug-only flag that should have indicated `debug=True`.

yizhuoz004: See this: https://github.com/NVIDIA/TensorRT-Incubator/blob/main/mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/Options.h#L53
I want used https://github.com/NVIDIA/TensorRT-Incubator/blob/main/mlir-tensorrt/test/python/mlir_tensorrt_compiler/compiler_api/test_compiler_debug_dump.py as a reference.

parthchadha (approved):

Issue #168, opened 2024-08-30
Update `convert_inputs_to_tensors` to require Python numbers


pranavm-nvidia:
`convert_inputs_to_tensors` currently passes any non-tensor arguments directly to the `Tensor` constructor, which means that it ends up enabling more than just Python numbers (which was its original intention) in our APIs. The issue is that this makes the API inconsistent, with some APIs working fine with DLPack tensors and others not. 

We should update `convert_inputs_to_tensors` to only accept Python numbers to avoid this. 

This change will also make it slightly easier to fix https://github.com/NVIDIA/TensorRT-Incubator/issues/154 since we can just look at the Python type to determine whether or not to allow automatic type conversions.

pranavm-nvidia:
Implemented by #185

PR #185, opened 2024-09-05
[Tripy] Do not implicitly convert external input formats in `convert_inputs_to_tensors` decorator

slyubomirsky:
Addresses issue #168 and does not permit implicit conversion of non-Tripy inputs other than Python numbers (e.g., Numpy arrays).

### DIFF
--- a/tripy/examples/nanogpt/weight_loader.py
+++ b/tripy/examples/nanogpt/weight_loader.py
@@ -1,4 +1,3 @@
-
 #
 # SPDX-FileCopyrightText: Copyright (c) 1993-2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
@@ -50,7 +49,7 @@ def load_weights_from_hf(model, model_type, dtype):
                 weight = hf_state_dict[key].t().contiguous()
         if "ln" not in key:
             weight = weight.to(torch_dtype)
-        param = tp.Parameter(weight)
+        param = tp.Parameter(tp.Tensor(weight))
         tripy_state_dict[key] = param
 
     model.load_from_state_dict(tripy_state_dict)
@@ -110,7 +109,7 @@ def load_quant_weights_from_hf(model, model_type, dtype, quant_mode):
 
         if "ln" not in key:
             weight = weight.to(torch_dtype)
-        param = tp.Parameter(weight.contiguous())
+        param = tp.Parameter(tp.Tensor(weight.contiguous()))
         tripy_state_dict[key] = param
 
     model.load_from_state_dict(tripy_state_dict)
--- a/tripy/examples/nanogpt/weight_loader.py
+++ b/tripy/examples/nanogpt/weight_loader.py
@@ -49,7 +49,7 @@ def load_weights_from_hf(model, model_type, dtype):
                 weight = hf_state_dict[key].t().contiguous()
         if "ln" not in key:
             weight = weight.to(torch_dtype)
-        param = tp.Parameter(tp.Tensor(weight))
+        param = tp.Parameter(weight)
         tripy_state_dict[key] = param
 
     model.load_from_state_dict(tripy_state_dict)
@@ -109,7 +109,7 @@ def load_quant_weights_from_hf(model, model_type, dtype, quant_mode):
 
         if "ln" not in key:
             weight = weight.to(torch_dtype)
-        param = tp.Parameter(tp.Tensor(weight.contiguous()))
+        param = tp.Parameter(weight.contiguous())
         tripy_state_dict[key] = param
 
     model.load_from_state_dict(tripy_state_dict)
--- a/tripy/tests/backend/mlir/test_compiler.py
+++ b/tripy/tests/backend/mlir/test_compiler.py
@@ -53,9 +53,9 @@ class TestErrorMapping:
         with FlatIRTensor.context(["This is the first level of context"]):
             with FlatIRTensor.context(["This is the second level of context"]):
                 # We need to emit an error from one of the internally created `FlatIRTensor`s to see the context
-                a = tp.ones(1)
-                b = tp.ones(1)
-                trace = Trace([a + b])
+                a = tp.ones((1,))
+                b = tp.ones((1,))
+                trace = Trace([a[0] + b[0]])
                 flat_ir = trace.to_flat_ir()
                 producer = flat_ir.outputs[0].producer.inputs[0]
                 flat_ir_inputs = ",".join(map(lambda i: i.name, producer.producer.inputs))
--- a/tripy/tests/backend/mlir/test_compiler.py
+++ b/tripy/tests/backend/mlir/test_compiler.py
@@ -55,7 +55,7 @@ class TestErrorMapping:
                 # We need to emit an error from one of the internally created `FlatIRTensor`s to see the context
                 a = tp.ones((1,))
                 b = tp.ones((1,))
-                trace = Trace([a[0] + b[0]])
+                trace = Trace([a + b])
                 flat_ir = trace.to_flat_ir()
                 producer = flat_ir.outputs[0].producer.inputs[0]
                 flat_ir_inputs = ",".join(map(lambda i: i.name, producer.producer.inputs))
--- a/tripy/tests/frontend/test_utils.py
+++ b/tripy/tests/frontend/test_utils.py
@@ -16,6 +16,7 @@
 #
 
 import cupy as cp
+import numpy as np
 
 import tripy as tp
 from tripy.frontend.utils import convert_inputs_to_tensors
@@ -291,3 +292,17 @@ class TestConvertInputsToTensors:
             match=r"At least one of the arguments: \('a', 'b', 'c'\) must be a \`tripy.Tensor\`.",
         ):
             x, y, z = sync_arg_types(3.0, 3, 4)
+
+    def test_invalid_argument_type(self):
+        with helper.raises(
+            tp.TripyException,
+            match=r"convert_inputs_to_tensors decorator supports conversion only for Python numbers or sequences thereof",
+        ):
+            _ = func(np.array([1, 2, 3]))
+
+    def test_invalid_argument_type_in_sequence(self):
+        with helper.raises(
+            tp.TripyException,
+            match=r"convert_inputs_to_tensors decorator supports conversion only for Python numbers or sequences thereof",
+        ):
+            _ = func([np.array([1, 2, 3]), np.array([4, 5, 6])])
--- a/tripy/tests/frontend/test_utils.py
+++ b/tripy/tests/frontend/test_utils.py
@@ -293,14 +293,28 @@ class TestConvertInputsToTensors:
         ):
             x, y, z = sync_arg_types(3.0, 3, 4)
 
+    def test_seq_arg_invalid(self):
+        with helper.raises(
+            tp.TripyException,
+            match=r"Encountered non-number of type str in sequence: hello",
+        ):
+            _ = func([1, 2, "hello"])
+
+    def test_nested_seq_inconsistent_len(self):
+        with helper.raises(
+            tp.TripyException,
+            match=r"Expected a sequence of length 3 but got length 4: \[7, 8, 9, 10\]",
+        ):
+            _ = func([[1, 2, 3], [4, 5, 6], [7, 8, 9, 10]])
+
+    def test_nested_seq_inconsistent_types(self):
+        with helper.raises(
+            tp.TripyException,
+            match=r"Expected a sequence but got str: hi",
+        ):
+            _ = func([[1, 2, 3], [4, 5, 6], "hi"])
+
     def test_invalid_argument_type_not_converted(self):
         a = np.array([1, 2, 3])
         b = func(np.array([1, 2, 3]))
         assert (a == b).all()
-
-    def test_invalid_argument_type_in_sequence_not_converted(self):
-        a = np.array([1, 2, 3])
-        b = np.array([4, 5, 6])
-        c, d = func([np.array([1, 2, 3]), np.array([4, 5, 6])])
-        assert (c == a).all()
-        assert (d == b).all()
--- a/tripy/tests/frontend/test_utils.py
+++ b/tripy/tests/frontend/test_utils.py
@@ -293,16 +293,14 @@ class TestConvertInputsToTensors:
         ):
             x, y, z = sync_arg_types(3.0, 3, 4)
 
-    def test_invalid_argument_type(self):
-        with helper.raises(
-            tp.TripyException,
-            match=r"convert_inputs_to_tensors decorator supports conversion only for Python numbers or sequences thereof",
-        ):
-            _ = func(np.array([1, 2, 3]))
+    def test_invalid_argument_type_not_converted(self):
+        a = np.array([1, 2, 3])
+        b = func(np.array([1, 2, 3]))
+        assert (a == b).all()
 
-    def test_invalid_argument_type_in_sequence(self):
-        with helper.raises(
-            tp.TripyException,
-            match=r"convert_inputs_to_tensors decorator supports conversion only for Python numbers or sequences thereof",
-        ):
-            _ = func([np.array([1, 2, 3]), np.array([4, 5, 6])])
+    def test_invalid_argument_type_in_sequence_not_converted(self):
+        a = np.array([1, 2, 3])
+        b = np.array([4, 5, 6])
+        c, d = func([np.array([1, 2, 3]), np.array([4, 5, 6])])
+        assert (c == a).all()
+        assert (d == b).all()
--- a/tripy/tests/integration/test_functional.py
+++ b/tripy/tests/integration/test_functional.py
@@ -130,36 +130,3 @@ class TestCopyFunctional:
         out = tp.copy(a, tp.device("gpu"))
         assert out.tolist() == [1, 2]
         assert out.device.kind == "gpu"
-
-
-class TestConversionToTripyType:
-    @pytest.mark.parametrize(
-        "reverse_direction",
-        [False, True],
-    )
-    @pytest.mark.parametrize(
-        "input0",
-        [cp.ones((2, 3), dtype=cp.float32), cp.ones((3,), dtype=np.float32)],
-    )
-    @pytest.mark.parametrize(
-        "input1",
-        [
-            [
-                4.0,
-            ],
-            (5.0,),
-            cp.array([4.0], dtype=cp.float32),
-            cp.ones((1, 3), dtype=cp.float32),
-            torch.Tensor([[4.0]]),
-        ],
-    )
-    def test_element_wise_prod(self, reverse_direction, input0, input1):
-        a = tp.Tensor(input0)
-        if isinstance(input1, torch.Tensor):
-            input1 = input1.to("cuda")
-        if reverse_direction:
-            out = tp.Tensor(input1) * a
-            input0, input1 = input1, input0
-        else:
-            out = a * tp.Tensor(input1)
-        assert cp.array_equal(cp.from_dlpack(out), cp.array(input0) * cp.array(input1))
--- a/tripy/tests/integration/test_functional.py
+++ b/tripy/tests/integration/test_functional.py
@@ -158,8 +158,8 @@ class TestConversionToTripyType:
         if isinstance(input1, torch.Tensor):
             input1 = input1.to("cuda")
         if reverse_direction:
-            out = input1 * a
+            out = tp.Tensor(input1) * a
             input0, input1 = input1, input0
         else:
-            out = a * input1
+            out = a * tp.Tensor(input1)
         assert cp.array_equal(cp.from_dlpack(out), cp.array(input0) * cp.array(input1))
--- a/tripy/tests/test_function_registry.py
+++ b/tripy/tests/test_function_registry.py
@@ -318,7 +318,7 @@ class TestFunctionRegistry:
 
     def test_tensor_literal(self, registry):
         @registry("test")
-        def func(n: "tripy.TensorLiteral.sig"):
+        def func(n: "tripy.types.tensor_literal"):
             return n
 
         assert registry["test"](1) == 1
@@ -472,7 +472,7 @@ class TestFunctionRegistry:
 
     def test_error_tensor_literal_not_sequence(self, registry):
         @registry("test")
-        def func(n: "tripy.TensorLiteral.sig"):
+        def func(n: "tripy.types.tensor_literal"):
             return n
 
         with helper.raises(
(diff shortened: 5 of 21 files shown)


tripy/examples/nanogpt/weight_loader.py
```diff
@@ -50,7 +49,7 @@ def load_weights_from_hf(model, model_type, dtype):
                 weight = hf_state_dict[key].t().contiguous()
         if "ln" not in key:
             weight = weight.to(torch_dtype)
-        param = tp.Parameter(weight)
+        param = tp.Parameter(tp.Tensor(weight))
```
slyubomirsky, line 52: We might want to make an exception for `tp.Parameter`, since it might be convenient. Perhaps we could have a switch in the decorator for permitting external input types (false by default)?

tripy/tripy/frontend/utils.py
```diff
@@ -193,6 +194,20 @@ def find_sync_target_dtype(arg_name):
                     return None
 
                 def convert_nontensor_arg(arg, list_index=None):
+                    def is_valid_sequence(seq_arg: Sequence) -> bool:
```
slyubomirsky, line 197: We could validate the entire sequence if that's preferable, but it would mean a lot more iteration. On the other hand, explicit Python sequences passed in are likely to be small, so it may be okay to iterate over all of it.

tripy/examples/nanogpt/weight_loader.py
pranavm-nvidia: Yeah, a switch makes sense.

tripy/tests/frontend/test_utils.py
```diff
@@ -291,3 +292,17 @@ def test_sync_arg_type_invalid(self):
             match=r"At least one of the arguments: \('a', 'b', 'c'\) must be a \`tripy.Tensor\`.",
         ):
             x, y, z = sync_arg_types(3.0, 3, 4)
+
+    def test_invalid_argument_type(self):
+        with helper.raises(
+            tp.TripyException,
+            match=r"convert_inputs_to_tensors decorator supports conversion only for Python numbers or sequences thereof",
```
pranavm-nvidia, line 299: The user shouldn't need to know anything about `convert_inputs_to_tensors`. Can we just pass the argument through to the underlying API? Because of the function registry, each API does type checking, so it should fail with a more obvious error message

tripy/examples/nanogpt/weight_loader.py
pranavm-nvidia: Actually, instead of a switch, can we just make the `Parameter` constructor create the tensor if the input is not a tensor? Otherwise we're over-complicating the decorator.

tripy/tripy/frontend/utils.py
```diff
@@ -193,6 +194,20 @@ def find_sync_target_dtype(arg_name):
...
+                        if len(seq_arg) == 0:
+                            return True
+                        if isinstance(seq_arg[0], Sequence):
+                            return is_valid_sequence(seq_arg[0])
+                        return isinstance(seq_arg[0], (int, float))
+
+                    if not isinstance(arg, (int, float)) and not (isinstance(arg, Sequence) and is_valid_sequence(arg)):
+                        raise_error(
```
pranavm-nvidia, line 205: I think a better way to do this might be to just `return arg`. Then the function registry's type checking will take over and generate potentially nicer error messages (we might need to update the type annotations to `Union[Tensor, numbers.Number]` and then also make the registry handle `Union` types (basically, in `matches_type` we would just do `any(matches_type(...))` for the contained types

tripy/tests/frontend/test_utils.py
slyubomirsky: Good point, I'll change the message

tripy/tripy/frontend/utils.py
slyubomirsky: We would need to update the registry to have better error messages for a case like this. Here is a failure I get if I try this approach:

```python
import tripy as tp
import numpy as np

a = tp.Tensor([1, 2, 3])
b = np.array([1, 2, 3])
print(a + b)
```

Error:
```
  File "/tripy/tripy/function_registry.py", line 281, in wrapper
    return self.find_overload(key, args, kwargs)(*args, **kwargs)
  File "/tripy/tripy/function_registry.py", line 185, in __call__
    return self.func(*args, **kwargs)
  File "/tripy/tripy/function_registry.py", line 281, in wrapper
    return self.find_overload(key, args, kwargs)(*args, **kwargs)
  File "/tripy/tripy/function_registry.py", line 185, in __call__
    return self.func(*args, **kwargs)
  File "/tripy/tripy/frontend/utils.py", line 273, in wrapper
    return func(*new_args, **new_kwargs)
  File "/tripy/tripy/frontend/trace/ops/binary_elementwise.py", line 208, in __add__
    return BinaryElementwise.build([self, other], BinaryElementwise.Kind.SUM)
  File "/tripy/tripy/frontend/trace/ops/base.py", line 77, in build
    inp_trace_tensors = [inp.trace_tensor for inp in inputs]
  File "/tripy/tripy/frontend/trace/ops/base.py", line 77, in <listcomp>
    inp_trace_tensors = [inp.trace_tensor for inp in inputs]
AttributeError: 'numpy.ndarray' object has no attribute 'trace_tensor'
```

slyubomirsky: I think that's only for the case of properties like `__add__` because this error message seems sensible:

```python
import tripy as tp

b = np.array([1, 2, 3])
print(tp.reshape(b, (1, 3)))
```

Error:
```
Could not find an implementation for function: 'reshape'.
    Note: Argument types were: [ndarray, tuple].
    Candidate overloads were:

    --> /tripy/tripy/frontend/trace/ops/reshape.py:75 in reshape()
          |
       75 | def reshape(input: "tripy.Tensor", shape: Union["tripy.Shape", Sequence[Union[int, "tripy.Tensor"]]]) -> "tripy.Tensor":
       76 |     ...
          | 

    Not a valid overload because: For parameter: 'input', expected an instance of type: 'Tensor' but got argument of type: 'ndarray'.
```

slyubomirsky: Correctly checking convertible tensor literals would require a bit of extra machinery, since you could have a sequence of any depth. This is possible to handle, thankfully: https://stackoverflow.com/questions/53845024/defining-a-recursive-type-hint-in-python

tripy/examples/nanogpt/weight_loader.py
slyubomirsky: I went with the latter approach.

tripy/tests/frontend/test_utils.py
slyubomirsky: No longer applicable, see below

tripy/tripy/frontend/utils.py
slyubomirsky: I added in support for `Union` and `Sequence` in the checker (the generated messages are very clunky, but we could make further changes) and put in a WAR for tensor literals (general support for recursive annotations is probably a rabbithole we don't want to go down)

pranavm-nvidia: Do you have an example of the generated error messages?

tripy/tripy/common/utils.py
```diff
@@ -21,6 +21,19 @@
 
 from tripy.common.exception import raise_error
 import tripy.common.datatype
+from tripy import export
+
+
+# This is a WAR to avoid implementing general support for recursive type annotations in the registry
+@export.public_api()
+class TensorLiteral:
```
pranavm-nvidia, line 29: Don't you already have recursive type checking though? Seems like it should just work - what's missing?

tripy/tripy/frontend/module/parameter.py
```diff
@@ -31,7 +31,9 @@ class Parameter(Tensor):
     """
 
     @frontend_utils.convert_inputs_to_tensors()
-    def __init__(self, tensor: "tripy.Tensor") -> None:
+    def __init__(
+        self, tensor: Union["tripy.Tensor", "np.ndarray", "cp.ndarray", "torch.Tensor", "jnp.ndarray"]
```
pranavm-nvidia, line 35: Maybe we can make the type annotation `Any`? It's hard to extensively list all types that support the DLPack interface

tripy/tripy/common/utils.py
slyubomirsky: It is not autorecursive. Here is what we would need to check:

```python
tensor_literal = Union[numbers.Number, Sequence["tensor_literal"]]
```

This would be needed to check sequences of arbitrary depth, since you could have a tensor literal of any rank. The `"tensor_literal"` is parsed as a `typing.ForwardRef`. Checking these in the general case can be very, very tricky.

tripy/tripy/frontend/module/parameter.py
slyubomirsky: I copied it over from the `Tensor` constructor but `Any` would be justified here, yes.

tripy/tripy/frontend/utils.py
pranavm-nvidia: I actually think it's more clear to just show the full annotation:
```
Not a valid overload because: For parameter: 'n', expected an instance of type: 'Sequence[Union[int, float]]' but got: <...>.
```

tripy/tripy/common/utils.py
pranavm-nvidia: I see. Could we put this into a new `types.py` file? We can also document it under a `types` section in API docs.

tripy/tripy/frontend/utils.py
slyubomirsky: I had the more detailed report because it will oftentimes give something less detailed like:

```
For parameter 'n', expected an instance of type typing.Sequence[typing.Union[int, float]] but got list
```

That's just how `type` works. However, this suggests that doing a little bit of extra work in the sequence case could pay off, so I'll try that instead 🙂

slyubomirsky: Improving the pretty-printing in the type matcher got rid of the need for weird recursive messages like these

tripy/tripy/common/utils.py
slyubomirsky: It's been split off into another file.

tripy/tripy/frontend/module/parameter.py
slyubomirsky: Replaced with `Any`.

tripy/tripy/frontend/utils.py
slyubomirsky: I've used the approach from #189 (`__forward_arg__`) to handle the recursive case directly, so no need for a WAR now 🙂The signature is a little clunky because it needed to be inside a class to be exported and documented (`tripy.TensorLiteral.sig`).

tripy/tests/backend/mlir/test_compiler.py
```diff
@@ -53,9 +53,9 @@ def test_reason_context(self):
         with FlatIRTensor.context(["This is the first level of context"]):
             with FlatIRTensor.context(["This is the second level of context"]):
                 # We need to emit an error from one of the internally created `FlatIRTensor`s to see the context
-                a = tp.ones(1)
-                b = tp.ones(1)
```
slyubomirsky, line 57: Is this intended to work? The signatures for `full`, `ones`, etc., indicate not, but `np.ones`, `torch.ones`, etc., would treat this as equivalent to `ones((1,))`.

slyubomirsky:
Note that the regexes in `test_function_registry.py` were passing spuriously. I think having lines like `    |` were resulting in matching individual whitespace, so those needed to be escaped.

tripy/tests/backend/mlir/test_compiler.py
pranavm-nvidia: Not intended to work. `np.ones` and `torch.ones` accept variadic positional args which is why they would treat this as `(1,)`, while we do not currently.

tripy/tests/backend/mlir/test_compiler.py
```diff
@@ -53,9 +53,9 @@ def test_reason_context(self):
...
-                trace = Trace([a + b])
+                a = tp.ones((1,))
+                b = tp.ones((1,))
+                trace = Trace([a[0] + b[0]])
```
pranavm-nvidia, line 58: I think the intention was to add the entire tensor? Not sure if it actually matters either way for this test.

tripy/tests/integration/test_functional.py
```diff
@@ -131,6 +131,7 @@ def test_multiple_copy_2(self):
         assert out.tolist() == [1, 2]
         assert out.device.kind == "gpu"
 
+
 class TestConversionToTripyType:
```
pranavm-nvidia, line 135: This entire test seems quite useless now. I think it's covered by more directed tests. Can we just remove it?

tripy/tests/test_function_registry.py
```diff
@@ -61,31 +61,30 @@ def test_overloading_string_annotations(self, int_float_registry):
         assert int_float_registry["transform"](1.0) == 0.0
 
     def test_error_on_missing_overload(self, int_float_registry):
+        # Note presence of ANSI color codes. Also note that the last | has a space after it
```
pranavm-nvidia, line 64: Why was this working before without the ANSI color codes? 
We have a `ANSI_ESCAPE` regex in `helper.py`. Maybe we can update `helper.raises` to strip out color codes?

tripy/tests/test_function_registry.py
```diff
@@ -270,3 +266,238 @@ def func(**kwargs: Dict[str, Any]):
...
+        assert registry["test"](2.0) == 2.0
+        assert registry["test"]([1, 2, 3]) == [1, 2, 3]
+        assert registry["test"]([[0.0, 1.0, 2.0], [3.0, 4.0, 5.0]]) == [[0.0, 1.0, 2.0], [3.0, 4.0, 5.0]]
+
+    def test_error_sequence(self, registry):
+        @registry("test")
+        def func(n: Sequence[int]) -> int:
+            return sum(n)
+
+        with pytest.raises(
```
pranavm-nvidia, line 335: Prefer `helper.raises` over `pytest.raises` since the former will include the exception text in verbose output and can also check for stack information for tensors.

tripy/tests/test_function_registry.py
```diff
@@ -270,3 +266,238 @@ def func(**kwargs: Dict[str, Any]):
...
+            TripyException,
+            match=dedent(
+                rf"""
+            Could not find an implementation for function: 'test'\.
+                Note: Argument types were: \[list\].
```
pranavm-nvidia, line 340: Maybe we can drop the line that shows argument types? We provide more detail later in the error anyway.

tripy/tripy/frontend/utils.py
```diff
@@ -193,6 +194,17 @@ def find_sync_target_dtype(arg_name):
                     return None
 
                 def convert_nontensor_arg(arg, list_index=None):
+                    def is_valid_sequence(seq_arg: Sequence) -> bool:
+                        if len(seq_arg) == 0:
+                            return True
+                        if isinstance(seq_arg[0], Sequence):
+                            return is_valid_sequence(seq_arg[0])
+                        return isinstance(seq_arg[0], (int, float))
```
pranavm-nvidia, line 202: We can check `numbers.Number` instead of limiting it to `int` and `float`.

tripy/tripy/frontend/utils.py
pranavm-nvidia: If we're doing type checking we might as well be thorough. On the other hand, mixed sequences seem unlikely.

tripy/tripy/function_registry.py
```diff
@@ -98,7 +98,26 @@ def _get_annotations(self):
...
+        def render_arg_type(arg: Any) -> str:
+            # it is more useful to report more detailed types for sequences/tuples in error messages
+            from typing import List, Tuple
+
+            if isinstance(arg, List):
+                if len(arg) == 0:
+                    return "List"
+                return f"List[{render_arg_type(arg[0])}]"
+            if isinstance(arg, Tuple):
+                return f"Tuple[{', '.join(map(render_arg_type, arg))}]"
```
pranavm-nvidia, line 114: Can this logic be generalized to all sequences? Something like:
```py
if isinstance(arg, Sequence):
    return f"{type(arg).title()}[{', '.join(map, render_arg_type, arg)}]"
```
Then we don't need special logic for lists either.

tripy/tripy/common/types.py
```diff
@@ -0,0 +1,32 @@
...
+
+@export.public_api(document_under="types.rst")
+class TensorLiteral:
+    """
+    The `sig` member of this class can be used as a type annotation for tensor literals.
+    A tensor literal can be a Python number or a sequence of tensor literals
+    (i.e., a sequence of numbers of any depth).
+    """
+
+    sig = Union[numbers.Number, Sequence["tripy.TensorLiteral.sig"]]
```
pranavm-nvidia, line 32: Do we need the class here? Could we just make it something like:
```py
NestedSequence = Union[numbers.Number, Sequence["tripy.NestedSequence"]]
```

slyubomirsky: I would prefer not to have a class, but is there a way to export it with documentation? The decorator only works on classes or functions.

tripy/tests/backend/mlir/test_compiler.py
slyubomirsky: The test works either way. I'll keep it as it was.

tripy/tests/test_function_registry.py
slyubomirsky: I think it was passing spuriously before (see the above comment about the pipe characters, I think they would have matched _anything_).

tripy/tripy/function_registry.py
slyubomirsky: I can do it that way. I had them separately there because I think it might make sense not to have the brackets for an empty list (i.e., the type of the contents is unknown), but it may not be an important distinction.

slyubomirsky: There is no `title()` method on many type instances, I noticed.

pranavm-nvidia: Oops, meant to do `type(arg).__name__.title()`

tripy/tests/test_function_registry.py
slyubomirsky: There's an issue with using `ANSI_ESCAPE` in `helper.raises`, namely it is relying on `pytest.raises` to do the matching. I am not sure it will be possible to replace the ANSI sequences at that stage, but I'll try it.

tripy/tripy/common/types.py
pranavm-nvidia: Yes, you can call the decorator just like a function. Here's an example: https://github.com/NVIDIA/TensorRT-Incubator/blob/main/tripy/tripy/config.py#L42. 
Maybe we can export type variables under a `tripy.types` submodule for which you'll also need to export the module itself like so: https://github.com/NVIDIA/TensorRT-Incubator/blob/main/tripy/tripy/config.py#L28

tripy/tests/integration/test_functional.py
slyubomirsky: Yes, you're right, there's no implicit conversion going on. Removing.

tripy/tripy/common/types.py
slyubomirsky: Ah, lovely. I'll do that, then.

tripy/tests/test_function_registry.py
pranavm-nvidia: Sure, let's leave it as it is for now then. If color codes become too painful later we can implement our own matching logic.

tripy/tripy/frontend/utils.py
slyubomirsky: Added validation in the decorator and in the registry, though I tried to keep the reporting simple in the latter case.

tripy/tripy/common/types.py
slyubomirsky: Implemented in this manner

tripy/tripy/types.py
```diff
@@ -0,0 +1,40 @@
...
+export.public_api()(sys.modules[__name__])
+
+tensor_literal = export.public_api(
+    document_under="types.rst",
+    autodoc_options=[":no-index:"],
+    module=sys.modules[__name__],
+    symbol="tensor_literal",
+)(Union[numbers.Number, Sequence["tripy.types.tensor_literal"]])
+
+tensor_literal.__doc__ = """Denotes the recursive type annotation for tensor literals.
```
slyubomirsky, line 37: This was weird, but it was the only way I could add in a doc that the UX test recognized. Is there another way to add a docstring to something that's not a class or function?

pranavm-nvidia: You can just add it directly below as a docstring:
```
tensor_literal = ...
"""
Denotes the ...
"""
```

tripy/tripy/frontend/shape.py
```diff
@@ -155,25 +155,30 @@ def __str__(self) -> str:
...
 
+    def _validate_add_argument(self, other):
+        if isinstance(other, Shape):
+            return
+        if not isinstance(other, Sequence) or (len(other) != 0 and not isinstance(other[0], int)):
+            raise_error(
+                "Invalid types for addition with a Tripy Shape: "
+                "Implicit conversions are done only for sequences of Python ints. "
+                "Consider calling tp.Shape for an explicit conversion.",
+                details=[other],
```
pranavm-nvidia, line 166: We can probably move some of this to details. Also we should indicate what `other` is prior to printing it:
```suggestion
                "Invalid type for addition with a Tripy Shape: ",
                details=[
                    "Implicit conversions are done only for sequences of Python ints. ", 
                    "Consider calling tp.Shape for an explicit conversion.",
                    "Note: Argument was: ",
                    other
                ],
```

tripy/tripy/frontend/trace/ops/binary_elementwise.py
```diff
@@ -183,7 +183,10 @@ def to_flat_ir(self, inputs, outputs):
     dtype_constraints={"other": "T1", constraints.RETURN_VALUE: "T1"},
     function_name="__radd__",
 )
-def __add__(self: Union["tripy.Tensor", Any], other: Union["tripy.Tensor", Any]) -> "tripy.Tensor":
+def __add__(
+    self: Union["tripy.Tensor", "tripy.types.tensor_literal"],
```
pranavm-nvidia, line 187: Is it worth having a type variable for this? I expect it'll be fairly common

tripy/tripy/types.py
```diff
@@ -0,0 +1,40 @@
...
+
+import numbers
+import sys
+from typing import Union, Sequence
+
+from tripy import export
+
+export.public_api()(sys.modules[__name__])
+
+tensor_literal = export.public_api(
```
pranavm-nvidia, line 30: Let's use PascalCase for types IMO:
```suggestion
TensorLiteral = export.public_api(
```
Also, "tensor literal" seems a bit misleading, since the type is actually just a Python collection, and *not* a tensor

tripy/tripy/types.py
pranavm-nvidia: That's weird. It seems to work fine in `config.py`

tripy/tripy/types.py
slyubomirsky: Yeah that's a good point on the name. I'm not sure what the best alternative would be. `NumberArray`, perhaps? Numpy calls such values "array-like"

pranavm-nvidia: It's really just a nested sequence of numbers. Maybe `NumberSequence` or `NestedNumberSequence`?

tripy/tripy/frontend/shape.py
slyubomirsky: Done as suggested, though I had to update the tests too.

tripy/tripy/frontend/trace/ops/binary_elementwise.py
slyubomirsky: Good idea. There were nearly 30 instances, so I added a variable `ConvertibleToTensor`. Not 100% sure on the name.

pranavm-nvidia: `ConvertibleToTensor` seems fine. Another option is `TensorLike` (similar to `ArrayLike` in numpy).

slyubomirsky: I like `TensorLike`, I'll go with that.

tripy/tripy/types.py
slyubomirsky: I'll go with `NestedNumberSequence`. A little wordy but users would probably want to use `TensorLike` regardless.

pranavm-nvidia (approved):

tripy/tripy/frontend/utils.py
slyubomirsky: Resolving since there don't seem to be objections to how it's handled here. We can revisit if it becomes a performance issue (I doubt that will happen) or we want to change the reporting.

PR #189, opened 2024-09-06
Fix arange, add floordiv, fix Fill op

parthchadha:
This PR adds the following:
1. `arange`: allows start, stop or size  (or all) to be tensors.
2. Adds `floordiv` operation.
3. Allows fill operation to accept value as input.

### DIFF
--- a/tripy/tripy/frontend/ops/tensor_initializers.py
+++ b/tripy/tripy/frontend/ops/tensor_initializers.py
@@ -15,7 +15,6 @@
 # limitations under the License.
 #
 
-import math
 import numbers
 from typing import Optional, Sequence, Union
 
@@ -370,7 +369,7 @@ def arange(
     },
     dtype_constraints={"dtype": "T1", constraints.RETURN_VALUE: "T1"},
 )
-def arange(stop: numbers.Number, dtype: "tripy.dtype" = datatype.float32) -> "tripy.Tensor":
+def arange(stop: Union[numbers.Number, "tripy.Tensor"], dtype: "tripy.dtype" = datatype.float32) -> "tripy.Tensor":
     r"""
     Returns a 1D tensor containing a sequence of numbers in the half-open interval
     :math:`[0, \text{stop})` incrementing by 1.
@@ -392,7 +391,6 @@ def arange(stop: numbers.Number, dtype: "tripy.dtype" = datatype.float32) -> "tr
         assert (cp.from_dlpack(output).get() == np.arange(5, dtype=np.float32)).all()
     """
     from tripy.common.datatype import int64
-    from tripy.frontend.tensor import Tensor
 
     if dtype == int64:
         raise_error("Known issue with i64. Arange currently does not work with int64 inputs. Issue #116")
--- a/tripy/tripy/frontend/ops/tensor_initializers.py
+++ b/tripy/tripy/frontend/ops/tensor_initializers.py
@@ -299,7 +299,10 @@ def triu(tensor: "tripy.Tensor", diagonal: int = 0) -> "tripy.Tensor":
     dtype_constraints={"dtype": "T1", constraints.RETURN_VALUE: "T1"},
 )
 def arange(
-    start: numbers.Number, stop: numbers.Number, step: numbers.Number = 1, dtype: "tripy.dtype" = datatype.float32
+    start: Union[numbers.Number, "tripy.Tensor"],
+    stop: Union[numbers.Number, "tripy.Tensor"],
+    step: Union[numbers.Number, "tripy.Tensor"] = 1,
+    dtype: "tripy.dtype" = datatype.float32,
 ) -> "tripy.Tensor":
     r"""
     Returns a 1D tensor containing a sequence of numbers in the half-open interval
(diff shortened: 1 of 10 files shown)


tripy/tripy/frontend/ops/tensor_initializers.py
```diff
@@ -330,21 +333,32 @@ def arange(
...
             "Arange tensor is empty.",
             details=[
                 f"start={start}, stop={stop}, step={step}",
             ],
         )
+
+    if isinstance(size, Tensor):
+        from tripy.frontend.trace.ops.cast import cast
+
+        size = cast(size, int32)
```
pranavm-nvidia, line 358: Can't we express this via dtype constraints?

yizhuoz004 (approved):

tripy/tripy/function_registry.py
```diff
@@ -102,14 +104,23 @@ def matches_type(name: str, annotation: type, arg: Any):
             # In cases where a type is not available at the time of function definition, the type
             # annotation may be provided as a string. Since we need the actual type, we just
             # eval it here.
-            if isinstance(annotation, str):
+            import importlib
+
+            if isinstance(annotation, (str, ForwardRef)):
```
pranavm-nvidia, line 109: @slyubomirsky also implemented this in his PR: https://github.com/NVIDIA/TensorRT-Incubator/pull/185
Which one should we keep?

parthchadha: I can use @slyubomirsky's implementation since it implements broader support.

slyubomirsky: Just curious, where did `ForwardRef` come into play here? I did not know about `__forward_arg__`, so I think I could use this to handle tensor literals without a workaround.

tripy/tripy/frontend/trace/ops/binary_elementwise.py
```diff
@@ -432,6 +432,74 @@ def __rtruediv__(self: numbers.Number, other: Union["tripy.Tensor", Any]) -> "tr
...
+@constraints.dtype_info(
+    dtype_variables={"T1": ["float32", "float16", "bfloat16", "int8", "int32", "int64"]},
+    dtype_constraints={"self": "T1", "other": "T1", constraints.RETURN_VALUE: "T1"},
+)
+def __floordiv__(self: Union["tripy.Tensor", Any], other: Union["tripy.Tensor", Any]) -> "tripy.Tensor":
+    """
+    Performs an elementwise floor division.
+
+    Args:
+        self: Tensor to be divided by other.
```
yizhuoz004, line 446: Need to mention "floor division"

tripy/tripy/frontend/trace/ops/binary_elementwise.py
```diff
@@ -432,6 +432,74 @@ def __rtruediv__(self: numbers.Number, other: Union["tripy.Tensor", Any]) -> "tr
...
+        a = tp.Tensor([4.0, 6.0])
+        b = tp.Tensor([3.0, 4.0])
+        output = a // b
+
+        assert np.array_equal(cp.from_dlpack(output).get(), np.array([1.0, 1.0]))
+    """
+    from tripy.frontend.trace.ops.cast import cast
+    from tripy.common.datatype import int32
+
+    return cast(cast(BinaryElementwise.build([self, other], BinaryElementwise.Kind.DIV), int32), self.dtype)
```
pranavm-nvidia, line 466: Is there no floordiv stableHLO op? Casting won't work for values outside the int32 range, which I think could happen quite easily

yizhuoz004: Yeah, there's no floordiv in stablehlo, I only found "div + floor". Besides we better add a rewriter in MLIR-TRT that matches `div + floor` -> TRT elementwise layer `kFLOOR_DIV`. 
Still easiest approach is TRT dialect.

tripy/tripy/frontend/ops/tensor_initializers.py
parthchadha: size is computed from start, stop and step which are inputs to this function.  (`size = 0 - ((start - stop) // step)`)

pranavm-nvidia: But if we require start/stop/step to be `int32` then we won't need to cast here right?

parthchadha: These can be of float type, ex: `tp.arange(2.3, 0.8, -0.2)`

tripy/tripy/frontend/trace/ops/binary_elementwise.py
parthchadha: I have switched to div + floor but this caused a segfault in mlir-tensorrt: https://github.com/NVIDIA/TensorRT-Incubator/issues/208. Will use the cast approach until segfault is fixed in mlir-tensorrt.

tripy/tripy/frontend/ops/tensor_initializers.py
```diff
@@ -291,7 +290,10 @@ def triu(tensor: "tripy.Tensor", diagonal: int = 0) -> "tripy.Tensor":
     dtype_constraints={"dtype": "T1", constraints.RETURN_VALUE: "T1"},
 )
 def arange(
-    start: numbers.Number, stop: numbers.Number, step: numbers.Number = 1, dtype: "tripy.dtype" = datatype.float32
+    start: Union[numbers.Number, "tripy.Tensor"],
```
pranavm-nvidia, line 293: Should we mention that these need to be scalars?

PR #191, opened 2024-09-09
Adds layer metadata APIs

pranavm-nvidia:


mlir-tensorrt/compiler/include/mlir-tensorrt-c/Compiler/Compiler.h
```diff
@@ -28,6 +28,8 @@
 #include "mlir-c/Support.h"
 #include "mlir-executor-c/Common/Common.h"
 #include "mlir-executor-c/Support/Status.h"
+#include "mlir-tensorrt-dialect/Utils/Types.h"
+#include "mlir/CAPI/IR.h"
```
christopherbate, line 32: Both of these inclusions are illegal.  This is the public (i.e. exported) C interface of a shared library.

mlir-tensorrt/tensorrt/include/mlir-tensorrt-dialect/Utils/Types.h
```diff
@@ -0,0 +1,12 @@
+#ifndef MLIR_TENSORRT_UTILS_TYPES_H
+#define MLIR_TENSORRT_UTILS_TYPES_H
+
+#include "mlir-c/IR.h"
+#include <functional>
+#include <string>
+
+namespace mlirtrt {
```
christopherbate, line 8: We go through a lot of trouble to have a C-pure interface for a reason. You can't just mix in C++.

christopherbate (changes requested):
Here's how to fix this up:

1. Remove the `Types.h` header you created. That is not necessary.

2. Refer to this basic outline:

In NetworkEncoder.h, you should just use `std::function` directly. Notice that we use `Operation*` here, not `MlirOperation`.

```
  std::function<std::string(Operation*)> layerMetadataCallback;
```

`NetworkEncoder.cpp` and `StableHloToExecutable.h` would be changed accordingly to reflect the  use of `Operation*`.

Now the CAPI can be reworked as follows. First, you can't use `std::function` in the C API since it is plain C. You must use function pointers. The idea here is that the 
user must specify a callback that communicates a string across the C API boundary. The caller (say in the C++ PyBind binding code) will construct a C++ function and pass it as a raw function pointer. There are two considerations:

1. To avoid exposing a whole string interface to the C API, we can just pass an "append" function for the caller to invoke in order to construct the string (again using function pointers).
2. When we construct a function that is passed-by-pointer across the C API boundary, we can use C++ lambda's to specify this function, but we are not allowed to have captured variables. Therefore, if we require capturing some object, we must 'lift' that state out into a separate "user data" field which can be treated opaquely by the API. There are two instances of this at work in the below example. We use one raw `void*` to represent the `std::string` being constructed by the CAPI implementation code. Then we also use one `void*` to represent the "context" of the callback passed from the CAPI client/caller code. This allows for wrapping things like the `py::object` combing from Python into an effectively *static* C function handle. 

To do this, first in `mlir-tensorrt-c/Compiler/Compiler.h`, change `MTRT_MetadataCallback` to be just a typedef of a function pointer type:

```
typedef void (*MTRT_MetadataCallback)(MlirOperation op, MlirStringCallback append, void *appendCtx, void *userData);
```

Now in `CAPI/Compiler/Compiler.cpp` we can do this:

```

MTRT_Status
mtrtStableHloToExecutableOptionsSetTensorRTTranslationMetadataCallback(
    MTRT_StableHLOToExecutableOptions options, MTRT_MetadataCallback callback, void * userData) {
  StableHLOToExecutableOptions *cppOpts = unwrap(options);

  // Note: this is a pure C function pointer. Cannot capture, hence the need
  // to lift the accumulation state into `appendCtx`.
  auto appendCallback = [](MlirStringRef str, void *appendCtx) {
     std::string &accum = *reinterpret_cast<std::string*>(appendCtx);
     accum += std::string(str.data, str.length);
   };

  // Note: capture here is OK since we are now constructing a C++ std::function.
  // It's important that we just pass the user data as given.
  cppOpts->layerMetadataCallback = [&](Operation *op) -> std::string {
    std::string accum;
    void *appendCtx = reinterpret_cast<void*>(&accum);
    callback(wrap(op), &appendFunc, appendCtx, userData);
    return accum;
  };
  return mtrtStatusGetOk();
}
```

Now the tricky part here is how to call it from the client side (PyBind).

Below I assume `py::object` is a callable object passed from Python. Exact syntax may vary, though.

```
.def("set....", 
[](PyStableHLOToExecutableOptions &self, py::object callback) {
   
   // We are constructing a C function, so we cannot capture. 
   // We must lift the function state into a separate struct.
   struct MetadataCallbackUserData { 
      py::object pyCallback;
   };


   MTRT_MetadataCallback callback = [](MlirOperation op, MlirStringCallback append,
        void* appendCtx, void *userDataVoid) {

       auto *userData = static_cast<MetadataCallbackUserData*>(userDataVoid);
       
       // Call the actual Python callback.
       // Wrap this in a try...catch.
       std::string result = userData->pyCallback(op);

       // Send the string data across the CAPI boundary.
       append(MlirStringRef{result.data(),result.size()}, appendCtx) 
   };
   
   // This is stack data, but it only ever gets accessed inside the call below. 
   MetadataCallbackUserData userData {
     std::move(callback);
   };
      
   // omitted: handle errors. 
   mtrtStableHloToExecutableOptionsSetTensorRTTranslationMetadataCallback(
     self, callback, reinterpret_cast<void*>(&userData));
});
```

mlir-tensorrt/python/bindings/Compiler/CompilerPyBind.cpp
```diff
@@ -270,7 +275,40 @@ PYBIND11_MODULE(_api, m) {
...
+              try {
+                result = pyCallback(op);
+              } catch (const std::exception &e) {
+                std::cerr << e.what() << std::endl;
+              }
+
+              append(MlirStringRef{result.data(), result.size()}, appendCtx);
+            };
+
+            self.callback = pyCallback;
```
pranavm-nvidia, line 304: Seems like this part is causing segfaults in some cases. Not entirely sure what the cause is, but it could be that the callback is being overwritten by another one and for some reason, something is still pointing to the old address.

pranavm-nvidia: Interestingly, I cannot reproduce the segault if I run the problematic test(s) individually. It's only when they're run together (even though they are run serially) that the segfault happens.

pranavm-nvidia: Valgrind output:
```
tests/backend/test_compiler_api.py::TestExecutable::test_incorrect_arguments[args0-kwargs0-Missing argument: b] ==9363== Invalid read of size 8
==9363==    at 0x1BC5B5E1C: std::_Function_base::_M_empty() const (std_function.h:247)
==9363==    by 0x1BC5EE5D4: std::function<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > (MlirOperation)>::operator bool() const (std_function.h:573)
==9363==    by 0x1BC5EE541: std::function<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > (MlirOperation)>::function(std::function<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > (MlirOperation)> const&) (std_function.h:389)
==9363==    by 0x1BC5800F2: pybind11_init__api(pybind11::module_&)::$_3::operator()((anonymous namespace)::PyStableHLOToExecutableOptions&, std::function<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > (MlirOperation)>) const::{lambda(MlirOperation, void (*)(MlirStringRef, void*), void*, void*)#1}::operator()(MlirOperation, void (*)(MlirStringRef, void*), void*, void*) const (python/bindings/Compiler/CompilerPyBind.cpp:291)
==9363==    by 0x1BC5800B0: pybind11_init__api(pybind11::module_&)::$_3::operator()((anonymous namespace)::PyStableHLOToExecutableOptions&, std::function<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > (MlirOperation)>) const::{lambda(MlirOperation, void (*)(MlirStringRef, void*), void*, void*)#1}::__invoke(MlirOperation, void (*)(MlirStringRef, void*), void*, void*) (python/bindings/Compiler/CompilerPyBind.cpp:288)
```

pranavm-nvidia: Looks like the problem is that the options instance isn't living long enough:
```
==10578==  Address 0x52588278 is 24 bytes inside a block of size 40 free'd
==10578==    at 0x484C6BF: operator delete(void*) (vg_replace_malloc.c:1131)
==10578==    by 0x1BC57D613: std::default_delete<(anonymous namespace)::PyStableHLOToExecutableOptions>::operator()((anonymous namespace)::PyStableHLOToExecutableOptions*) const (unique_ptr.h:85)
```

pranavm-nvidia: This is quite strange - at the point where we call `compiler_stablehlo_to_executable`:
```
==8341==    by 0x1C0047526: mtrtCompilerStableHLOToExecutable (compiler/lib/CAPI/Compiler/Compiler.cpp:248)
==8341==    by 0x1BC580B11: pybind11_init__api(pybind11::module_&)::$_4::operator()((anonymous namespace)::PyCompilerClient&, MlirOperation, (anonymous namespace)::PyStableHLOToExecutableOptions&) const (python/bindings/Compiler/CompilerPyBind.cpp:322)
```
the options have already been freed:
```
==8341==  Address 0x518050e8 is 24 bytes inside a block of size 40 free'd
==8341==    at 0x484C6BF: operator delete(void*) (vg_replace_malloc.c:1131)
==8341==    by 0x1BC57D613: std::default_delete<(anonymous namespace)::PyStableHLOToExecutableOptions>::operator()((anonymous namespace)::PyStableHLOToExecutableOptions*) const (unique_ptr.h:85)
==8341==    by 0x1BC57D582: std::unique_ptr<(anonymous namespace)::PyStableHLOToExecutableOptions, std::default_delete<(anonymous namespace)::PyStableHLOToExecutableOptions> >::~unique_ptr() (unique_ptr.h:361)
==8341==    by 0x1BC57D0BB: pybind11::class_<(anonymous namespace)::PyStableHLOToExecutableOptions>::dealloc(pybind11::detail::value_and_holder&) (pybind11.h:1872)
```

This doesn't make sense to me. Shouldn't the arguments to a function live at least for the scope of that function?

pranavm-nvidia: The problem was that the closure was capturing a pointer by reference instead of value.

mlir-tensorrt/python/bindings/Compiler/CompilerPyBind.cpp
```diff
@@ -19,7 +19,9 @@
 #include "mlir/Bindings/Python/PybindAdaptors.h"
 #include "pybind11/pybind11.h"
 #include "llvm/Support/DynamicLibrary.h"
+#include <iostream>
```
christopherbate, line 22: This should be removed

pranavm-nvidia: I will also need to remove the logging when we catch an exception in the Python callback in that case. Unless there is another logging mechanism in place?

pranavm-nvidia: @christopherbate updated to use `llvm::errs` instead.

mlir-tensorrt/compiler/include/mlir-tensorrt-c/Compiler/Compiler.h
```diff
@@ -60,6 +60,10 @@ typedef struct MTRT_StableHLOToExecutableOptions {
   void *ptr;
 } MTRT_StableHLOToExecutableOptions;
 
+typedef void (*MTRT_MetadataCallback)(MlirOperation op,
```
christopherbate, line 63: Please add documentation comment

mlir-tensorrt/compiler/include/mlir-tensorrt-c/Compiler/Compiler.h
```diff
@@ -77,6 +81,11 @@ MLIR_CAPI_EXPORTED MTRT_Status mtrtStableHloToExecutableOptionsSetDebugOptions(
     const char **debugTypes, size_t debugTypeSizes,
     const char *dumpIrTreeDir = nullptr, const char *dumpTensorRTDir = nullptr);
 
+MLIR_CAPI_EXPORTED MTRT_Status
```
christopherbate, line 84: Please add documentation comment that explains what this method does

mlir-tensorrt/tensorrt/include/mlir-tensorrt-dialect/Utils/Options.h
```diff
@@ -124,7 +129,9 @@ class OptionsContext : public llvm::cl::SubCommand {
   void print(llvm::raw_ostream &os) const;
 
   /// Get a hash derived from the string representation of the options.
-  llvm::hash_code getHash() const;
+  virtual llvm::hash_code getHash() const;
```
christopherbate, line 132: I think it would be cleaner to just return `std::optional<llvm::hash_code>` and eliminate the `shouldInvalidateCache` method

christopherbate (approved):

PR #202, opened 2024-09-11
Add scalar shape class to Tripy

parthchadha:
This PR adds `ScalarShape` to Tripy which is encodes a value that is sliced out of a `Shape` tensor.

### DIFF
--- a/tripy/tripy/frontend/shape.py
+++ b/tripy/tripy/frontend/shape.py
@@ -25,6 +25,62 @@ import tripy.frontend.utils as frontend_utils
 
 
 @export.public_api()
+class ShapeScalar(Tensor):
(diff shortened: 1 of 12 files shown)


tripy/tripy/frontend/shape.py
```diff
@@ -24,6 +24,62 @@
 import tripy.frontend.utils as frontend_utils
 
 
+@export.public_api()
+class ScalarShape(Tensor):
```
pranavm-nvidia, line 28: Maybe `ShapeScalar` is more accurate? It is a scalar that came from a shape, not a shape that is scalar.

pranavm-nvidia (approved):

PR #203, opened 2024-09-11
Add tp.pad

yizhuoz004:


yizhuoz004:
Blocked until translation from `stablehlo.dynamic_pad` -> `tensorrt.slice` is fixed.

tripy/tripy/frontend/trace/ops/pad.py
```diff
@@ -0,0 +1,143 @@
...
+    dtype_constraints={"input": "T1", constraints.RETURN_VALUE: "T1"},
+)
+def pad(input: "tripy.Tensor", padding_sizes: Sequence[Tuple], padding_value: Union[int, float] = 0) -> "tripy.Tensor":
+    r"""
+    Pads `input` with `padding_value` of given `padding_sizes`.
+
+    Args:
+        input: The input tensor.
+
+        padding_sizes: A sequence of padding sizes of each dimension. Its length must equal to the rank
```
parthchadha, line 108:
```suggestion
        padding_sizes: A sequence of padding sizes of each dimension. Its length must be equal to the rank
```

tripy/tripy/frontend/trace/ops/pad.py
```diff
@@ -0,0 +1,143 @@
...
+            of `input`. Each element of `padding_size` is a tuple of integers or scalars `(low, high)`,
```
parthchadha, line 109: `padding_size` can also be a `Tensor`.

tripy/tripy/frontend/trace/ops/pad.py
```diff
@@ -0,0 +1,143 @@
...
+    from tripy.frontend.trace.ops.concatenate import concatenate
+    from tripy.frontend.trace.ops.unsqueeze import unsqueeze
+
+    if not any(isinstance(s, Tensor) for s in padding_sizes):
+        return Tensor(padding_sizes, dtype=int32)
+
+    sizes_1d = []
+    for size in padding_sizes:
+        if isinstance(size, Tensor):
+            assert size.rank == 0
```
parthchadha, line 87:
```suggestion
            assert size.rank == 0, f"Size expected to be of rank 0, got {size.rank}."
```

tripy/tripy/frontend/trace/ops/pad.py
yizhuoz004: I don't think passing `padding_sizes` as a `Nx2` Tensor is a common use case, it is always more convenient for users to construct a sequence than a Tensor.

However, using `ScalarShape` in `padding_sizes` is supported, for example:
```py
pad(inp, [(0, a.shape[0]), (0, a.shape[1])])
```

tripy/tripy/frontend/trace/ops/pad.py
```diff
@@ -0,0 +1,145 @@
...
+        inputs.insert(1, pad_val_tensor)
+        inputs.append(interior_pad_tensor)
+        # set padding size tensors' shape
+        # because stablehlo requires static shapes
+        inputs[2].shape = pad_size_shape
+        inputs[3].shape = pad_size_shape
+        DynamicPadOp.build(inputs, outputs)
+
+
+def _convert_pad_sizes(padding_sizes):
```
pranavm-nvidia, line 75: Why don't we just use the convert to shapes decorator?

tripy/tripy/frontend/trace/ops/pad.py
```diff
@@ -0,0 +1,145 @@
...
+@constraints.dtype_info(
+    dtype_variables={
+        "T1": ["float32", "float16", "bool", "int32"],
+    },
+    dtype_constraints={"input": "T1", constraints.RETURN_VALUE: "T1"},
+)
+def pad(
+    input: "tripy.Tensor",
+    padding_sizes: Sequence[Tuple[Union[int, "tripy.ShapeScalar"]]],
+    padding_value: Union[int, float] = 0,
```
pranavm-nvidia, line 104: Can we use `pad` and `value` for these? Also, should we add a `mode` argument so we can eventually support things like reflection padding?

tripy/tripy/frontend/trace/ops/pad.py
yizhuoz004: There's an extra pre-process step: `padding_low, padding_high = list(zip(*padding_sizes))`
It would be more convenient if convert to shapes decorator is just a standalone function.

tripy/tripy/frontend/trace/ops/pad.py
yizhuoz004: Do you mean changing the argument name to `pad` and `value`? The function name is `pad`, might be a bit confusing.

Yes we should have a `mode` argument, but stablehlo dynamic_pad op does not support it. We will need tensorrt dialect for that, could discuss with Chris next time.

pranavm-nvidia: Yeah, `pad` and `value` is what torch uses. We can add a `mode` argument that only supports `constant` for now maybe? Then we won't need API changes later.

yizhuoz004: Updated the API.

parthchadha (approved):

tripy/tripy/frontend/trace/ops/pad.py
```diff
@@ -0,0 +1,154 @@
...
+    from tripy.frontend.trace.ops.concatenate import concatenate
+    from tripy.frontend.trace.ops.unsqueeze import unsqueeze
+
+    if not any(isinstance(s, Tensor) for s in padding_sizes):
+        return Tensor(padding_sizes, dtype=int32)
+
+    sizes_1d = []
+    for size in padding_sizes:
+        if isinstance(size, Tensor):
+            assert size.rank == 0, f"Size expected to be of rank 0, got {size.rank}."
```
pranavm-nvidia, line 87: This needs to be a user-visible error right? We should use `raise_error` in that case

pranavm-nvidia (approved):

tripy/tripy/frontend/trace/ops/pad.py
yizhuoz004: This is just an internal assert to help debug, `pad` has type constraint to be `Sequence[Tuple[Union[int, "tripy.ShapeScalar"]]]`.

PR #204, opened 2024-09-12
Add tp.MaxPool

yizhuoz004:


tripy/tripy/frontend/module/pooling.py
```diff
@@ -0,0 +1,138 @@
...
+from tripy import export, utils
+from tripy.frontend.module.module import Module
+
+from tripy.common.exception import raise_error
+
+
+@export.public_api(document_under="operations/modules")
+@dataclass
+@utils.constant_fields(["dtype", "normalized_shape"])
+class MaxPool(Module):
```
pranavm-nvidia, line 30: We distinguish modules from functions based on whether the op has weights. So in this case, I think a function would be better.

yizhuoz004: Makes sense, switched to function.

yizhuoz004:
http://10.110.37.18:8001/operations/functions/maxpool.html

tripy/tests/integration/test_pooling.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+    @pytest.mark.parametrize("dtype", [tp.float32, tp.float16])
+    def test_maxpool_2d(self, kernel_dims, stride, padding, dtype):
+        inp_tp = tp.reshape(tp.arange(64, dtype=dtype), (1, 1, 8, 8))
+        out = tp.maxpool(inp_tp, kernel_dims=kernel_dims, stride=stride, padding=padding)
+
+        torch_padding = (padding[0][0], padding[1][0])
+        pool_torch = torch.nn.MaxPool2d(kernel_size=kernel_dims, stride=stride, padding=torch_padding)
+        expected = pool_torch(torch.from_dlpack(inp_tp).to("cpu"))
+        assert torch.allclose(expected, torch.from_dlpack(out).to("cpu"))
```
pranavm-nvidia, line 41: For completeness, we should also check the shape when using `allclose`.

tripy/tripy/frontend/trace/ops/pooling.py
```diff
@@ -0,0 +1,157 @@
...
+            window_dims=window_dims,
+            window_strides=window_strides,
+            padding=padding,
+        )
+
+
+@export.public_api(document_under="operations/functions")
+@constraints.dtype_info(
+    dtype_variables={
+        "T1": ["float32", "float16"],
```
pranavm-nvidia, line 86: Are these really the only types supported?

tripy/tripy/frontend/trace/ops/pooling.py
```diff
@@ -0,0 +1,157 @@
...
+                "Padding must have the same length as kernel_dims.",
+                [f"Got padding={padding}, ", f"kernel_dims={kernel_dims}"],
+            )
+        if any(len(pad_size) != 2 for pad_size in padding):
+            raise_error(
+                f"Padding must be provided as a sequence of pairs of integers.",
+                details=[f"Supplied padding attribute: {padding} contains sequences that are not of length 2."],
+            )
+    else:
+        padding = [(0, 0)] * spatial_dims
```
pranavm-nvidia, line 155: A lot of this is very similar to what we have for convolution. I'm wondering if we can reuse any of those checks? Unless we think the checks would be different enough that it doesn't help.

yizhuoz004: Refactored this part into a `check_conv_pooling_args` util func.

tripy/tripy/frontend/trace/ops/pooling.py
```diff
@@ -0,0 +1,148 @@
...
+    from tripy.common.datatype import int8
+
+    spatial_dims = len(kernel_dims)
+    if spatial_dims != 2 and spatial_dims != 3:
+        raise_error("Unsupported kernel_dims, must be 2D or 3D.", [f"Got kernel_dims={kernel_dims}"])
+    if input.dtype == int8 and spatial_dims != 2:
+        raise_error(
+            "Unsupported kernel_dims for int8 dtype.",
+            [f"Only 2D kernel_dims is supported for int8, got {kernel_dims}."],
+        )
```
pranavm-nvidia, line 142: Where does this restriction come from? I'd be more inclined to let it fail in the backend so that we're not artificially imposing a constraint.

yizhuoz004: TRT does not support 3D max pooling for int8, it gives an error that I think is not that clean:
```
[optimizer.cpp::getFormatRequirements::3301] Error Code 2: Internal Error (Assertion !n->candidateRequirements.empty() failed. No supported formats for [tensorrt.pooling] (t15))
```
I can remove the frontend check if this is fine.

tripy/tests/frontend/trace/ops/test_pooling.py
```diff
@@ -0,0 +1,39 @@
...
+        a = tp.ones((1, 1, 4, 4))
+        with helper.raises(tp.TripyException, "Unsupported kernel_dims, must be 2D or 3D."):
+            tp.maxpool(a, (2,))
+
+    def test_invalid_stride(self):
+        a = tp.ones((1, 1, 4, 4))
+        with helper.raises(tp.TripyException, "Stride must have the same length as kernel_dims."):
+            tp.maxpool(a, (2, 2), stride=(1,))
+
+    def test_invalid_padding_1(self):
```
pranavm-nvidia, line 31:
```suggestion
    def test_invalid_padding_length(self):
```

tripy/tests/frontend/trace/ops/test_pooling.py
```diff
@@ -0,0 +1,39 @@
...
+        a = tp.ones((1, 1, 4, 4))
+        with helper.raises(tp.TripyException, "Padding must have the same length as kernel_dims."):
+            tp.maxpool(a, (2, 2), padding=((1, 1),))
+
+    def test_invalid_padding_2(self):
```
pranavm-nvidia, line 36:
```suggestion
    def test_invalid_padding_contents(self):
```

pranavm-nvidia (approved):

PR #209, opened 2024-09-16
Add modulo operator

parthchadha:
Fixes https://github.com/NVIDIA/TensorRT-Incubator/issues/201

### DIFF
--- a/tripy/tripy/frontend/trace/ops/binary_elementwise.py
+++ b/tripy/tripy/frontend/trace/ops/binary_elementwise.py
@@ -35,6 +35,7 @@ class BinaryElementwise(BaseTraceOp):
         MUL = " * "
         DIV = " / "
         FLOOR_DIV = " // "
+        MOD = " % "
         MAXIMUM = "maximum"
         MINIMUM = "minimum"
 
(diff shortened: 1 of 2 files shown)


tripy/tripy/frontend/trace/ops/binary_elementwise.py
```diff
@@ -144,6 +145,39 @@ def to_flat_ir(self, inputs, outputs):
             DivideOp.build(inputs, [divide_out])
             # Then apply FloorOp to the result of the division
             FloorOp.build([divide_out], outputs)
+        elif self.kind == BinaryElementwise.Kind.MOD:
```
yizhuoz004, line 148: We can use `stablehlo.remainder`: https://github.com/openxla/stablehlo/blob/main/docs/spec.md#remainder

parthchadha: @yizhuoz004 remainder and modulo are not the same for negative numbers and I think adding conditionals for checking negative number and adjusting for that will be compute intensive compared to the current implementation.

yizhuoz004: I see, it confuses me that stablehlo has `remainder` but not `modulo` which is more common...

tripy/tripy/frontend/trace/ops/binary_elementwise.py
```diff
@@ -528,6 +562,73 @@ def __rfloordiv__(self: Union["tripy.Tensor", Any], other: Union["tripy.Tensor",
...
+        :caption: Example
+
+        a = tp.Tensor([4.0, 6.0])
+        b = tp.Tensor([3.0, 4.0])
+        output = a % b
+
+        assert np.array_equal(cp.from_dlpack(output).get(), np.array([1.0, 2.0]))
+    """
+    from tripy.frontend.trace.ops.cast import cast
+    from tripy.common.datatype import int32
```
yizhuoz004, line 594: imports are not needed.

tripy/tripy/frontend/trace/ops/binary_elementwise.py
```diff
@@ -528,6 +562,73 @@ def __rfloordiv__(self: Union["tripy.Tensor", Any], other: Union["tripy.Tensor",
...
+        :linenos:
+        :caption: Example
+
+        a = tp.Tensor([4.0, 6.0])
+        output = 2 % a
+
+        assert np.array_equal(cp.from_dlpack(output).get(), np.array([2.0, 2.0]))
+    """
+    from tripy.frontend.trace.ops.cast import cast
+    from tripy.common.datatype import int32
```
yizhuoz004, line 627: Same here

yizhuoz004 (approved):

PR #210, opened 2024-09-16
Fix reshape annotation to allow tensor input

parthchadha:
Fixes below bug: 
```
--> /tripy/examples/nanogpt/model.py:92 in <lambda>()
      |
   92 |         multi_heads = lambda x: tp.transpose(tp.reshape(x, multi_head_output_shape), 1, 2)
      | 

Could not find an implementation for function: 'reshape'.
    Candidate overloads were:

    --> /tripy/tripy/constraints.py:75 in reshape()
          |
       75 | def reshape(input: "tripy.Tensor", shape: Union["tripy.Shape", Sequence[Union[int, "tripy.Tensor"]]]) -> "tripy.Tensor":
       76 |     ...
          | 

    Not a valid overload because: For parameter: 'shape', expected an instance of type: 'typing.Union[ForwardRef('tripy.Shape'), typing.Sequence[typing.Union[int, ForwardRef('tripy.Tensor')]]]' but got argument of type: 'Tensor'.
```

### DIFF
--- a/tripy/tripy/frontend/trace/ops/reshape.py
+++ b/tripy/tripy/frontend/trace/ops/reshape.py
@@ -80,7 +80,9 @@ def reshape_impl(
     },
     dtype_constraints={"input": "T1", "shape": "T2", constraints.RETURN_VALUE: "T1"},
 )
-def reshape(input: "tripy.Tensor", shape: Union["tripy.Shape", Sequence[Union[int, "tripy.Tensor"]]]) -> "tripy.Tensor":
+def reshape(
+    input: "tripy.Tensor", shape: Union["tripy.Shape", Sequence[Union[int, "tripy.ShapeScalar"]]]
+) -> "tripy.Tensor":
     """
     Returns a new tensor with the contents of the input tensor in the specified shape.
 
--- a/tripy/tests/integration/test_arange.py
+++ b/tripy/tests/integration/test_arange.py
@@ -48,9 +48,7 @@ class TestArange:
         ):
             tp.arange(0, 5, 0)
 
-    def test_tensor_inputs(self):
-        start = tp.Tensor(1.0)
-        stop = tp.Tensor(5.0)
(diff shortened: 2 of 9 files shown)


tripy/tripy/frontend/trace/ops/reshape.py
```diff
@@ -80,7 +80,9 @@ def reshape_impl(
     },
     dtype_constraints={"input": "T1", "shape": "T2", constraints.RETURN_VALUE: "T1"},
 )
-def reshape(input: "tripy.Tensor", shape: Union["tripy.Shape", Sequence[Union[int, "tripy.Tensor"]]]) -> "tripy.Tensor":
+def reshape(
+    input: "tripy.Tensor", shape: Union["tripy.Shape", "tripy.Tensor", Sequence[Union[int, "tripy.Tensor"]]]
```
yizhuoz004, line 84: Should it be `Union["tripy.Shape", "tripy.ScalarShape", Sequence[Union[int, "tripy.Tensor"]]]`?

pranavm-nvidia: Also, it should be a sequence of shapescalars and not tensors right?

yizhuoz004: And there are few more places to update, search for `Sequence[Union[int, "tripy.Tensor"]]]` in the codebase:
- fill
- expand
- ones, zeros
- iota
- reshape

parthchadha: It can be Tensor also and not necessarily `ShapeScalar`. Yes, the Sequence should be updated to `ShapeScalar`.

pranavm-nvidia: It can only be a tensor once we have DDS right? Until then it must be a shape of some kind

pranavm-nvidia (approved):

PR #211, opened 2024-09-17
Fix nanoGPT sample reshape

yizhuoz004:


tripy/examples/nanogpt/model.py
```diff
@@ -76,19 +75,15 @@ def __init__(self, config):
         self.zeros = tp.zeros((1, 1, self.seq_len, self.seq_len), dtype=config.dtype)
 
     def __call__(self, x: tp.Tensor):
-        B, T, C = tp.reshape(x.shape[0], (1,)), tp.reshape(x.shape[1], (1,)), tp.reshape(x.shape[2], (1,))
+        B, T = x.shape[0], x.shape[1]
```
pranavm-nvidia, line 78:
```suggestion
        B, T = x.shape[0:2]
```

pranavm-nvidia: Why do we call it `T` here? Isn't it sequence length? Usually that's `S`.

tripy/examples/nanogpt/model.py
```diff
@@ -108,9 +103,7 @@ def __call__(self, x: tp.Tensor):
         # (batch_size, num_heads, seq_len, seq_len) @ (batch_size, num_heads, seq_len, head_size) -> (batch_size, num_heads, seq_len, head_size)
         out = att @ v
         out = tp.cast(out, x.dtype)
-        out_shape = tp.concatenate([B, T, tp.Tensor([self.embedding_size])], dim=0)
-        # WAR to prevent computing output rank in infer_rank for reshape, will be addressed by #228
-        out_shape.trace_tensor.shape = (3,)
+        out_shape = [B, T, self.embedding_size]
         out = tp.reshape(tp.transpose(out, 1, 2), out_shape)
```
pranavm-nvidia, line 107: We can probably just inline the shapes here and above

pranavm-nvidia (approved):

tripy/examples/nanogpt/model.py
yizhuoz004: Yes it is sequence length, I think it's pretty common to use `T`, nanoGPT uses T: https://github.com/karpathy/nanoGPT/blob/master/model.py#L53

tripy/examples/nanogpt/model.py
yizhuoz004: Updated.

PR #212, opened 2024-09-18
[Tripy][Bugfix] Use correct types in `__str__` method for `ShapeScalar`

slyubomirsky:
Noticed a small error: The `__str__` method for `ShapeScalar` assumed that evaluating the scalar would give a list output, resulting in a type error, since the actual result is a scalar. This PR fixes that and adds a check to the unit tests.

### DIFF
--- a/tripy/tests/frontend/test_shape.py
+++ b/tripy/tests/frontend/test_shape.py
@@ -56,10 +56,9 @@ class TestShapeScalar:
         assert isinstance(s, tp.ShapeScalar)
         assert s.trace_tensor.producer.inputs == []
 
-        if isinstance(value, int):
-            assert s.__str__() == f"shape_scalar({value})"
-        else:
-            assert s.__str__() == f"shape_scalar({int(value.tolist())})"
+    def test_scalar_shape_str_method(self):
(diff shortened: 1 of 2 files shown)


parthchadha (approved):

tripy/tests/frontend/test_shape.py
```diff
@@ -40,13 +40,27 @@ def other_values(request):
...
-        s = tp.ShapeScalar(values)
+        s = tp.ShapeScalar(value)
 
         assert isinstance(s, tp.ShapeScalar)
         assert s.trace_tensor.producer.inputs == []
 
+        if isinstance(value, int):
+            assert s.__str__() == f"shape_scalar({value})"
+        else:
+            assert s.__str__() == f"shape_scalar({int(value.tolist())})"
```
pranavm-nvidia, line 62: Can we add a separate test for this? I don't think it should matter how the `ShapeScalar` was constructed, so we don't need to parameterize it over that.

slyubomirsky: Done

pranavm-nvidia (approved):

PR #215, opened 2024-09-19
Fix stablehlo pad converter

yizhuoz004:


pranavm-nvidia (approved):

christopherbate:
Thanks. Just needs test in `test/Conversion/StablehloToTensorRT/stablehlo-to-tensorrt.mlir`. There are several `stablehlo.pad` conversions tests but looks like they are missing a case for non-zero "low-padding".

yizhuoz004:
> Thanks. Just needs test in `test/Conversion/StablehloToTensorRT/stablehlo-to-tensorrt.mlir`. There are several `stablehlo.pad` conversions tests but looks like they are missing a case for non-zero "low-padding".

Added the non-zero "low-padding" test. Thanks!

PR #216, opened 2024-09-19
Add functionality for cartesian from polar coords

akhilg-nv:


tripy/tripy/frontend/ops/cartesian_via_polar.py
```diff
@@ -0,0 +1,67 @@
...
+
+@export.public_api(document_under="operations/functions")
+@constraints.dtype_info(
+    dtype_variables={"T1": ["float32", "float16", "bfloat16", "float8"]},
+    dtype_constraints={"abs": "T1", "angles": "T1", constraints.RETURN_VALUE: "T1"},
+)
+def cartesian_via_polar(abs: "tripy.Tensor", angles: "tripy.Tensor") -> "tripy.Tensor":
+    r"""
+    Constructs the real-valued cartesian coordinates from magnitude and angle representing polar coordinates. For input
+    ``abs`` and ``angle`` of shape :math:`(m_1, m_2, \ldots, m_i),` this function returns a new real tensor of shape
```
pranavm-nvidia, line 29:
```suggestion
    ``abs`` and ``angles`` of shape :math:`(m_1, m_2, \ldots, m_i),` this function returns a new real tensor of shape
```

tripy/tripy/frontend/ops/cartesian_via_polar.py
```diff
@@ -0,0 +1,67 @@
...
+
+from tripy import export, constraints
+
+
+@export.public_api(document_under="operations/functions")
+@constraints.dtype_info(
+    dtype_variables={"T1": ["float32", "float16", "bfloat16", "float8"]},
+    dtype_constraints={"abs": "T1", "angles": "T1", constraints.RETURN_VALUE: "T1"},
+)
+def cartesian_via_polar(abs: "tripy.Tensor", angles: "tripy.Tensor") -> "tripy.Tensor":
```
parthchadha, line 26: Why do we need this as an API in Tripy and why can't it be a user function? Can you point to similar functionality provided by torch, jax?

akhilg-nv: This recreates the functionality of [torch.polar](https://pytorch.org/docs/stable/generated/torch.polar.html) and [torch.view_as_real](https://pytorch.org/docs/stable/generated/torch.view_as_real.html). Since we don't support complex numbers, I think it's useful to have an existing function that directly accomplishes the same purpose.

pranavm-nvidia: Maybe it's better as a helper in the example?

pranavm-nvidia:
@akhilg-nv should we close this PR?

Issue #214, opened 2024-09-19
Refactor `convert_inputs_to_tensors` decorator to move automatic shape conversion logic to a separate helper


pranavm-nvidia:
Right now, the `convert_inputs_to_tensors` decorator is doing far too much: not only does it convert inputs to tensors and cast them, it can also convert them to shape tensors. The first part of its functionality should be restricted to operator overloads only - we don't want other APIs to do any implicit casting or conversion. The second part can be applied more broadly to make it easy to write ergonomic code. 

To achieve this we can split the decorator so that the functionality that converts arguments to shape tensors is provided as a separate decorator. 

We will probably also need to update `_make_stack_info_message` in `tripy/tripy/common/exception.py` to omit this decorator from the stack information when generating error messages.

pranavm-nvidia:
Implemented by #217

PR #217, opened 2024-09-20
[Tripy] Separate converting shape args from converting tensor args

slyubomirsky:
Addresses issue #214 and separates shape argument conversion from the `convert_inputs_to_tensors` decorator. Since most uses of the shape argument conversion would only convert one argument (usually named shape), this decorator takes a list of arguments to convert rather than converting all arguments by default. It also does not accept arguments that aren't sequences of Python numbers or Tripy Tensors, since it would make less sense to allow external data formats to silently pass through in cases that aren't operator overloads.

### DIFF
--- a/tripy/docs/post0_developer_guides/how-to-add-new-ops.md
+++ b/tripy/docs/post0_developer_guides/how-to-add-new-ops.md
@@ -200,10 +200,10 @@ import tripy.frontend.utils as frontend_utils
 # If we needed to provide any special autodoc options, we could use the `autodoc_options` parameter.
 @export.public_api(document_under="tensor_operations")
 
-# The `convert_inputs_to_tensors` decorator converts function arguments to Tensors.
-# This is what makes it possible for the user to use Python numbers in Tripy functions (e.g. `tensor + 1`)
-# In this case, we want `shape` to turn into a `tripy.Shape` instead of a regular `Tensor`.
-@frontend_utils.convert_inputs_to_tensors(shape_argument=["shape"], exclude=["dim", "dtype"])
+# The `convert_shape_inputs` decorator converts the specified function arguments into `tripy.Shape`s rather than Tensors,
+# which would allow for using Python numbers and sequences. The `convert_inputs_to_tensors` decorator more generally converts
+# function arguments into Tripy tensors and is also commonly used in the codebase.
+@frontend_utils.convert_shape_inputs(["shape"])
 def theta(shape: Tuple[int], dim: int = 0, dtype: datatype.dtype = datatype.float32) -> "tripy.Tensor":
     # For any public facing interfaces, we have documentation requirements which you can read
     # about in the 'Docs README' (linked below). The docstring we've implemented here
--- a/tripy/docs/post0_developer_guides/how-to-add-new-ops.md
+++ b/tripy/docs/post0_developer_guides/how-to-add-new-ops.md
@@ -200,7 +200,7 @@ import tripy.frontend.utils as frontend_utils
 # If we needed to provide any special autodoc options, we could use the `autodoc_options` parameter.
 @export.public_api(document_under="tensor_operations")
 
-# The `convert_shape_inputs` decorator converts the specified function arguments into `tripy.Shape`s rather than Tensors,
+# The `convert_shape_inputs` decorator converts the specified function arguments into `tripy.Shape`s,
 # which would allow for using Python numbers and sequences. The `convert_inputs_to_tensors` decorator more generally converts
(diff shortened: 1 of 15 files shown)


tripy/docs/post0_developer_guides/how-to-add-new-ops.md
```diff
@@ -200,10 +200,10 @@ import tripy.frontend.utils as frontend_utils
 # If we needed to provide any special autodoc options, we could use the `autodoc_options` parameter.
 @export.public_api(document_under="tensor_operations")
 
-# The `convert_inputs_to_tensors` decorator converts function arguments to Tensors.
-# This is what makes it possible for the user to use Python numbers in Tripy functions (e.g. `tensor + 1`)
-# In this case, we want `shape` to turn into a `tripy.Shape` instead of a regular `Tensor`.
-@frontend_utils.convert_inputs_to_tensors(shape_argument=["shape"], exclude=["dim", "dtype"])
+# The `convert_shape_inputs` decorator converts the specified function arguments into `tripy.Shape`s rather than Tensors,
```
pranavm-nvidia, line 203:
```suggestion
# The `convert_shape_inputs` decorator converts the specified function arguments into `tripy.Shape`s,
```

tripy/tripy/frontend/utils.py
```diff
@@ -68,86 +185,13 @@ def convert_inputs_to_tensors(
     sync_arg_types = utils.default(sync_arg_types, [])
     exclude = utils.default(exclude, [])
     unpack_argument = utils.default(unpack_argument, [])
-    shape_argument = utils.default(shape_argument, [])
 
     def impl(func):
```
pranavm-nvidia, line 189: Could we assert that `func` is a magic method? We should disallow this decorator for most APIs.

pranavm-nvidia (approved):

yizhuoz004 (approved):

tripy/tripy/frontend/utils.py
slyubomirsky: It would be easy to check the name, but there are exceptions in the codebase:
* `quantize` and `dequantize`
* The conversion for slicing is done in `slice_helper`
* `minimum` and `maximum` are binary elementwise in TRT, but they're not Python operators

We could encode the exceptions explicitly (which would thus discourage the use of the decorator in cases other than operator overloads) or manually do conversions in these cases. Doing manual conversions might make sense other than for `slice_helper`, though it has the disadvantage of not filling in as much stack information.

pranavm-nvidia: `quantize`, `dequantize`, `minimum`, and `maximum` shouldn't be doing automatic conversions IMO. `slice_helper` is really just a hack since our decorator doesn't handle slices correctly. We could add an exception for it now and eventually fix it in the decorator so we don't need the helper.

slyubomirsky: Put in a simple check, though I had to change the tests for the decorator (we can add them to the exception list if we would prefer normal-looking names there). For quantize and dequantize, I kept in a (manual) conversion for the scale argument because many call sites passed in single numbers and it makes sense to accept them.

pranavm-nvidia: @yizhuoz004 do you know why Q/DQ ops aren't implemented using frontend ops? The division part at least could be done in the frontend which would also simplify the trace op and not require us to explicitly convert `scale` to a tensor.

yizhuoz004: There are 2 reasons:
- We want to see Q/DQ ops in Trace IR, if implemented using frontend ops, we lose them.
- More stable and concise for pattern matching, with frontend ops, the number of stablehlo ops will explode because of the binary elementwise shape comparisons.

slyubomirsky: Are there any further concerns remaining with the decorator?

tripy/tests/frontend/test_utils.py
```diff
@@ -171,29 +179,35 @@ def test_convert_shape_unsqueeze_tensors(self):
         assert isinstance(s, tp.Shape)
         assert cp.from_dlpack(s).get().tolist() == [1, 2]
 
+    def test_convert_only_named_argument_to_shape(self):
```
pranavm-nvidia, line 182: nit: maybe "specified" instead of "named" so as not to confuse it with named in the sense of named vs. positional. 
```suggestion
    def test_convert_only_specified_argument_to_shape(self):
```

tripy/tripy/frontend/utils.py
pranavm-nvidia: Looks good now

PR #224, opened 2024-09-26
Various bug fixes to fix L1 tests

pranavm-nvidia:
- Adds tests to ensure we only use ASCII characters in our code and markdown files. Some non-ASCII characters that look like ASCII characters were breaking tests that make assumptions about the encoding.

- Updates dtype constraint tests to properly construct inputs for various APIs.

- Moves dtype constraint tests to L0 since they only add about 20 seconds and were breaking very often otherwise.

- Updates binary elementwise ops to only return `ShapeScalar`s if *all* inputs are also `ShapeScalar`s.

- Corrects dtype constraints for several APIs.

- Simplifies `mean` implementation.

- Corrects type annotations for some APIs and introduces a `ShapeLike` type.

- Adds support for `Optional` during type checking.

- Updates `convert_inputs_to_tensors` to now potentially convert tensors to `Tensor` subclasses. See the note in the code for details on the logic.

### DIFF
--- a/tripy/tripy/frontend/trace/ops/binary_elementwise.py
+++ b/tripy/tripy/frontend/trace/ops/binary_elementwise.py
@@ -15,12 +15,13 @@
 # limitations under the License.
 #
 
+import numbers
 from dataclasses import dataclass
 from typing import Any, Union
-import numbers
+
 import tripy.frontend.trace.ops.utils as op_utils
 import tripy.frontend.utils as frontend_utils
-from tripy import export, constraints
+from tripy import constraints, export
 from tripy.common import datatype
 from tripy.frontend.ops.registry import TENSOR_METHOD_REGISTRY
 from tripy.frontend.trace.ops.base import BaseTraceOp
@@ -50,7 +51,6 @@ class BinaryElementwise(BaseTraceOp):
 
     def infer_shape_output_idxs(self, inputs):
         # permit one input to be a shape but require the output to be a shape
-        from tripy.frontend.tensor import Tensor
         from tripy.frontend.shape import Shape, ShapeScalar
         from tripy.utils import Result
 
@@ -70,9 +70,7 @@ class BinaryElementwise(BaseTraceOp):
                     ]
                 )
             return Result.ok({"shape": [0]})
-        elif any(
-            map(lambda t: isinstance(t, ShapeScalar) or (isinstance(t, int) and not isinstance(t, Tensor)), inputs)
-        ):
+        elif all(map(lambda t: isinstance(t, ShapeScalar), inputs)):
             # Binary operation on ShapeScalar should yield another ShapeScalar.
             return Result.ok({"scalar": [0]})
         else:
@@ -90,9 +88,7 @@ class BinaryElementwise(BaseTraceOp):
     def infer_dtypes(self):
         self.outputs[0].dtype = self.inputs[0].dtype
 
-    def broadcast_inputs(self, inputs, outputs):
-        from tripy.common.datatype import int32
-        from tripy.flat_ir.ops import MaxOp
+    def broadcast_inputs(self, inputs):
         from tripy.flat_ir.tensor import FlatIRTensor
 
         rank = max(inputs[0].rank, inputs[1].rank)
@@ -134,7 +130,7 @@ class BinaryElementwise(BaseTraceOp):
         from tripy.flat_ir.ops import AddOp, DivideOp, FloorOp, MaxOp, MinOp, MulOp, PowOp, SubtractOp
         from tripy.flat_ir.tensor import FlatIRTensor
 
-        inputs = self.broadcast_inputs(inputs, outputs)
+        inputs = self.broadcast_inputs(inputs)
 
         if self.kind == BinaryElementwise.Kind.FLOOR_DIV:
             # First apply DivideOp
@@ -222,7 +218,7 @@ class Comparison(BinaryElementwise):
     def to_flat_ir(self, inputs, outputs):
         from tripy.flat_ir.ops import CompareOp
 
-        inputs = self.broadcast_inputs(inputs, outputs)
+        inputs = self.broadcast_inputs(inputs)
         CompareOp.build(inputs, outputs, compare_direction=self.kind.compare_direction)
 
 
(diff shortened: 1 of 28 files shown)


tripy/tripy/frontend/utils.py
```diff
@@ -212,9 +240,30 @@ def impl(func):
         @functools.wraps(func)
         def wrapper(*args, **kwargs):
             from tripy.frontend.tensor import Tensor
+            from tripy.frontend.shape import Shape, ShapeScalar
 
             all_args = utils.merge_function_arguments(func, *args, **kwargs)
 
+            # TODO: Should this be done a different way? Maybe we should manually specify which
+            # arguments to check across, similar to sync_arg_types? This could break in cases where
+            # we have operations that specifically require different tensor types.
```
pranavm-nvidia, line 249: Several open questions here:
1. Should there be a parameter that controls which inputs we apply this logic across?
2. Does the hierarchy even make sense?
3. How do we ensure the behavior of ops is also consistent with this hierarchy? Ops are slightly different in that the types of the inputs should determine the type of the output, while in this decorator we only care about the inputs.

slyubomirsky: Very curious issue to ponder. Per [my comment](https://github.com/NVIDIA/TensorRT-Incubator/issues/226#issuecomment-2377909735) in #226, I think putting Shape and Tensor in a hierarchy in this manner might not be advisable. It might cause fewer surprises to require manual conversions between these types where it's ambiguous, though `ShapeScalar` does not present this sort of problem.

Perhaps we could have an API by which operator implementations could check what types they do or do not accept for certain arguments?

slyubomirsky: We could also take the rank into account, since shapes must be of rank 1 and shape scalars must be of rank 0.

pranavm-nvidia: @slyubomirsky I'm not quite sure what the best mechanism would be to disallow mixing Shape/Tensor in these APIs. I've filed an issue for now: https://github.com/NVIDIA/TensorRT-Incubator/issues/233

In the meantime, I think this hierarchy works for all the ops that we currently have. Maybe we can revisit this if/when it breaks in the future? I would like to check this in so we will finally have a (mostly) passing L1.

slyubomirsky: If we want to have a hierarchy as a stopgap measure, that's fine. However, we should make it a lattice in that case.
```
             Tensor
            /         \
        Shape   ShapeScalar
```
All `Shape`s are `Tensor`s and all `ShapeScalar`s are `Tensor`s, but `ShapeScalar`s are not `Shape`s.

pranavm-nvidia: But when you have `Shape <op> ShapeScalar`, you would generally want a `Shape` back, since the scalar will be broadcasted. The priority ordering here does not imply anything about the inheritance hierarchy which is indeed the lattice you describe.

slyubomirsky: Ah that's true, I hadn't taken into account broadcasting. That works, in that case.

tripy/tripy/frontend/trace/ops/binary_elementwise.py
```diff
@@ -70,9 +70,7 @@ def infer_shape_output_idxs(self, inputs):
                     ]
                 )
             return Result.ok({"shape": [0]})
-        elif any(
-            map(lambda t: isinstance(t, ShapeScalar) or (isinstance(t, int) and not isinstance(t, Tensor)), inputs)
-        ):
+        elif all(map(lambda t: isinstance(t, ShapeScalar), inputs)):
```
slyubomirsky, line 73: Does not need to be done in this PR, but we might want to rename `infer_shape_output_idxs` since it's now being used to make judgments about `ShapeScalar`s too. We could fold that into discussion of #233

tripy/tripy/frontend/utils.py
```diff
@@ -28,12 +28,33 @@
...
 
-    assert not isinstance(arg, Tensor)
-    arg = Tensor(arg, dtype=dtype)
+    if isinstance(arg, numbers.Number) and TensorType == Shape:
+        # Shapes require 1D values.
+        arg = [arg]
+
+    arg = TensorType(arg)
+    if dtype is not None:
+        arg = cast(arg, dtype=dtype)
```
slyubomirsky, line 57: Perhaps we should disallow this branch or have an assert related to dtype for Shape/ShapeScalar since those have to be int32. Not strictly necessary.

tripy/tripy/frontend/utils.py
```diff
@@ -212,9 +240,28 @@ def impl(func):
         @functools.wraps(func)
         def wrapper(*args, **kwargs):
             from tripy.frontend.tensor import Tensor
+            from tripy.frontend.shape import Shape, ShapeScalar
 
             all_args = utils.merge_function_arguments(func, *args, **kwargs)
 
+            # TODO (#233): Disallow mixing Tensor/Shape.
```
slyubomirsky, line 247: Per previous discussion, I think this policy is reasonable as a stopgap measure, especially since we could simply not use the decorator if we run into a case where we think we should be able to mix types. I think the comment should mention the bit about ShapeScalars likely being broadcast into shapes, as the max type policy wouldn't work otherwise.

slyubomirsky (approved):
Some minor comments, though we could leave broader discussion to #233. Thank you for the various corrections and improvements.

tripy/tripy/frontend/trace/ops/binary_elementwise.py
pranavm-nvidia: Mentioned in the issue.

tripy/tripy/frontend/utils.py
pranavm-nvidia: I think we can actually allow other types since `cast` will then return a `Tensor` instead, which may still work with the operation.

tripy/tripy/frontend/utils.py
pranavm-nvidia: Updated the comment.

tripy/tripy/frontend/utils.py
slyubomirsky: True. I think that's fine and wouldn't lead to surprising behavior, since the only casting is with `sync_dtype`.

PR #230, opened 2024-09-27
Allow repeat to accept shapescalar along with int for repeats argument

parthchadha:


tripy/tripy/backend/mlir/utils.py
```diff
@@ -93,8 +93,14 @@ def get_mlir_scalar_attr(mlir_dtype, value):
 
 
 def list_to_dense_attr(data: List, mlir_dtype):
+    from tripy.frontend.shape import ShapeScalar
+
     if isinstance(data, numbers.Number):
         return [get_mlir_scalar_attr(mlir_dtype, data)]
+
+    if isinstance(data, ShapeScalar):
+        return [get_mlir_scalar_attr(mlir_dtype, data.tolist())]
```
pranavm-nvidia, line 102: This will trigger an eval. Is that what we want?

pranavm-nvidia (approved):

tripy/tripy/backend/mlir/utils.py
parthchadha: Check latest commit. The case which leads to evaluation of `ShapeScalar` is now handled in construction of `Shape`

pranavm-nvidia (approved):

PR #231, opened 2024-09-27
Add output_shape argument to resize ops

yizhuoz004:
Add `output_shape` argument to all resize ops in TRT dialect.

### DIFF
--- a/mlir-tensorrt/tensorrt/lib/TensorRT/IR/Verification.cpp
+++ b/mlir-tensorrt/tensorrt/lib/TensorRT/IR/Verification.cpp
@@ -936,12 +936,12 @@ LogicalResult tensorrt::ResizeNearestOp::verify() {
   auto outputRank = outputType.getRank();
   const int64_t resizeDim = std::min(static_cast<int64_t>(3), outputRank);
   for (int64_t i = outputRank - 1; i >= 0; --i) {
-    if (inputType.getDimSize(i) != outputType.getDimSize(i)) {
-      if (outputRank - i > resizeDim) {
+    if (inputType.isDynamicDim(i) || outputType.isDynamicDim(i))
+      continue;
+    if (inputType.getDimSize(i) != outputType.getDimSize(i))
+      if (outputRank - i > resizeDim)
         return emitOpError("only supports resizing on the innermost min(3, "
                            "rank(input)) dimensions");
-      }
-    }
   }
 
   if (getScales().has_value()) {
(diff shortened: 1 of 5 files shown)


mlir-tensorrt/tensorrt/lib/TensorRT/IR/Verification.cpp
```diff
@@ -955,12 +955,11 @@ LogicalResult tensorrt::ResizeNearestOp::verify() {
...
 
   for (int64_t i = 0; i < resizeDim; ++i) {
-    if (inputType.isDynamicDim(outputRank - 1 - i) ||
-        outputType.isDynamicDim(outputRank - 1 - i)) {
-      if (!getScales().has_value())
+    if (inputType.isDynamicDim(outputRank - 1 - i)) {
+      if (!getOutputShape())
         return emitOpError(
-            "output innermost min(3, rank(input)) dimension that resize on "
-            "cannot be dynamic when resize scales parameter is NOT "
```
yizhuoz004, line 963: @christopherbate I wonder if this is too strict, for example even with input spatial dims are dynamic, if `scales` is given, the op can still be valid, because I saw this function infers the output shape: https://github.com/NVIDIA/TensorRT-Incubator/blob/main/mlir-tensorrt/tensorrt/lib/TensorRT/IR/TypeInferenceInterfaceImpls.cpp#L1219

shelkesagar29:
@yizhuoz004 can you please make sure commit has both `tile` and `body`.
Otherwise such commits can't be merged into the core repo.

Fr example, for this commit you can have
[tensorrt] Add output_shape argument to resize ops
... Body goes here that explain work in short or detail depending on you ....


Idea here is just 
`git log` on master in core repo should give enough information to know changes added by each commit.

mlir-tensorrt/tensorrt/lib/TensorRT/IR/Verification.cpp
christopherbate: If `output_shape` is given, you can skip the entire outer `for` loop since.

If `output_shape` is not given, then you must have either 

1. The output result shape is static
2. All input dimensions are static and  the resize scales are given.

Basically you just need to return an error if the output shape cannot somehow be inferred (statically or dynamically).

christopherbate (approved):
LGTM, but please also add a new translation test under `test/Target/TensorRT/resize.mlir`.

mlir-tensorrt/tensorrt/lib/TensorRT/IR/Verification.cpp
yizhuoz004: Updated, thanks

PR #232, opened 2024-09-27
Add shape inference for resize ops in TRT dialect

yizhuoz004:
This MR needs https://github.com/NVIDIA/TensorRT-Incubator/pull/231.

### DIFF
--- a/mlir-tensorrt/tensorrt/test/TensorRT/dynamic-shape-inference.mlir
+++ b/mlir-tensorrt/tensorrt/test/TensorRT/dynamic-shape-inference.mlir
@@ -1,0 +1,70 @@
+// RUN: tensorrt-opt %s -split-input-file -test-tensorrt-shape-inference | FileCheck %s
+
+func.func @test_resize_linear(%arg0: tensor<10x10xf32>) -> (index, index) {
+  %result = tensorrt.resize_linear {
+    coordinateTransformation = #tensorrt.resize_coordinate_transformation<kALIGN_CORNERS>,
+    selectorForSinglePixel = #tensorrt.resize_selector<kUPPER>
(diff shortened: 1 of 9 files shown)


mlir-tensorrt/tensorrt/test/TensorRT/dynamic-shape-inference.mlir
```diff
@@ -0,0 +1,45 @@
...
+
+// CHECK-LABEL: test_resize_linear
+// CHECK-NEXT: %[[c20:.+]] = arith.constant 20 : index
+// CHECK-NEXT: return %[[c20]], %[[[c20]]] : index, index
+
+// -----
+
+func.func @test_resize_dynamic_batch(%arg0: tensor<?x1x10x10xf32>) -> (index, index, index, index) {
+  %result = tensorrt.resize_linear {
+    coordinateTransformation = #tensorrt.resize_coordinate_transformation<kALIGN_CORNERS>,
```
christopherbate, line 24: You'll need to add a test where `output_shape` is used. These two tests only cover static result shapes (in spatial dimensions)

christopherbate (changes requested):

mlir-tensorrt/tensorrt/test/TensorRT/dynamic-shape-inference.mlir
yizhuoz004: Added.

christopherbate:
For clustering, you can add test here https://github.com/NVIDIA/TensorRT-Incubator/blob/main/mlir-tensorrt/test/Dialect/Plan/materialize-shape-calculations.mlir

Issue #229, opened 2024-09-27
dlpack non-linear stride


parthchadha:
Torch tensor with non-linear strides when fed to Tripy does not respect the strides.

PR #236, opened 2024-09-28
Add checks for non-canonical strides

jhalakpatel:
https://github.com/NVIDIA/TensorRT-Incubator/issues/229

MLIR-TensorRT requires strides for function arguments and results in canonical order.

https://github.com/NVIDIA/TensorRT-Incubator/pull/252 adds a check to validate memref stride against a canonical stride order. In Tripy, memref strides are derived from framework DL Pack tensors. Creating a memref with a non-canonical DL Pack tensor stride throws an exception.

Add a try-catch block to catch such an exception and augment with suggestions on creating a DL Pack tensor with canonical stride for Tripy-supported frameworks.

Add unit tests to create a non-canonical stride tensor to validate exceptions and suggestions.

### DIFF
--- a/tripy/tests/frontend/test_stride.py
+++ b/tripy/tests/frontend/test_stride.py
@@ -1,0 +1,56 @@
+#
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import pytest
+import re
+
+import cupy as cp
+import numpy as np
+import torch
+
+import tripy as tp
+from tests.helper import raises
+
+
+class TestStride:
+    def test_non_canonical_stride(self):
+        test_cases = [
+            (
+                torch.arange(12, dtype=torch.float32).reshape(3, 4).transpose(0, 1),
+                lambda x: x.contiguous(),
+                lambda x: x.clone(memory_format=torch.contiguous_format),
+            ),
+            (
+                cp.arange(12, dtype=cp.float32).reshape(3, 4).transpose(1, 0),
+                cp.ascontiguousarray,
+                lambda x: x.copy(order="C"),
+            ),
+            (
+                np.arange(12, dtype=np.float32).reshape(3, 4).transpose(1, 0),
+                np.ascontiguousarray,
+                lambda x: x.copy(order="C"),
+            ),
+        ]
+
+        for array, contiguous_func, copy_func in test_cases:
+            # Test for exception with non-canonical strides
+            with pytest.raises(tp.TripyException, match="Non-canonical strides are not supported for Tripy tensors"):
+                tp.Tensor(array)
+
+            # Test successful creation with contiguous array
+            assert tp.Tensor(contiguous_func(array)) is not None
+            assert tp.Tensor(copy_func(array)) is not None
--- a/tripy/tripy/backend/mlir/memref.py
+++ b/tripy/tripy/backend/mlir/memref.py
@@ -15,9 +15,12 @@
 # limitations under the License.
 #
 
+import re
+
 from functools import lru_cache
 from typing import Sequence
 
+from tripy.utils import raise_error
 from tripy.backend.mlir import utils as mlir_utils
 from tripy.common import device as tp_device
 from tripy.common import utils as common_utils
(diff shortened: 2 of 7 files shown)


tripy/tripy/frontend/ops/equal.py
```diff
@@ -0,0 +1,59 @@
...
+from tripy import export, constraints
+from tripy.common.exception import raise_error
+
+
+@export.public_api(document_under="operations/functions")
+@constraints.dtype_info(
+    dtype_variables={"T1": ["float32", "float16", "bfloat16", "float8", "int4", "int8", "int32", "int64", "bool"]},
+    dtype_constraints={"a": "T1", "b": "T1"},
+)
+def equal(a: "tripy.Tensor", b: "tripy.Tensor") -> bool:
```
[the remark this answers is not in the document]
pranavm-nvidia: If we're following numpy/torch semantics, `allclose` doesn't account for shape while `equal` does. If we want to add `equal`, we should also check for shape below. Otherwise `allclose(..., rtol=0, atol=0)` should be equivalent to this.

tripy/tests/integration/test_equal.py
```diff
@@ -0,0 +1,89 @@
...
+        dtype_map = {
+            tp.float32: torch.float32,
+            tp.float16: torch.float16,
+            tp.bfloat16: torch.bfloat16,
+            tp.int32: torch.int32,
+            tp.int64: torch.int64,
+            tp.int8: torch.int8,
+            tp.bool: torch.bool,
+        }
+        return dtype_map[tp_dtype]
```
pranavm-nvidia, line 65:
```suggestion
    @pytest.mark.parametrize(
        "tensor_a, tensor_b, dtype, expected",
        [
            ([1, 2, 3], [1, 2, 3], tp.int64, True),
            ([1, 2, 3], [1, 2, 4], tp.int32, False),
            ([1.0, 2.0, 3.0], [1.0, 2.0, 3.0], tp.float16, True),
            ([1.0, 2.0, 3.0], [1.0, 2.0, 1.0], tp.bfloat16, False),
            ([1.0, 2.0, 3.0], [1.0, 2.0, 3.00001], tp.float32, False),
            ([True, False, True], [True, False, True], tp.bool, True),
        ],
    )
    def test_equal(self, tensor_a, tensor_b, dtype, expected):
        a = tp.Tensor(tensor_a, dtype=dtype)
        b = tp.Tensor(tensor_b, dtype=dtype)

        out = tp.equal(a, b)

        assert out == expected
```

tripy/tripy/frontend/ops/equal.py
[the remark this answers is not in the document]
jhalakpatel: I see. Yes, I see a few places where we used `bool(tp.all(a == b))` so I thought it would be helpful to add such an op in frontend.
@pranavm-nvidia I will fix equal to check for shape, type, and stride.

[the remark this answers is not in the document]
jhalakpatel: @pranavm-nvidia I think all.close also compare the shapes.
```
    def test_torch_all_close(self):
        t = torch.arange(12)
        r = t.reshape(3, 4)
        print(t.shape)
        print(r.shape)
>       assert torch.allclose(t, r)
E       RuntimeError: The size of tensor a (12) must match the size of tensor b (4) at non-singleton dimension 1

tests/integration/test_equal.py:98: RuntimeError
---------------------------------------------------------------------------------------------------------------------- Captured stdout call -----------------------------------------------------------------------------------------------------------------------
torch.Size([12])
torch.Size([3, 4])
```

I will remove `tp.equal` operation for now and use `tp.allclose` where required.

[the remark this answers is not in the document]
jhalakpatel: Let's add `tp.equal` operation in a separate PR.

tripy/tripy/frontend/trace/ops/storage.py
```diff
@@ -54,21 +55,40 @@ def __init__(
...
+            canonical_stride = utils.get_canonical_stride(self.shape)
+            error_message = (
+                f"Non-canonical strides detected:\n"
+                f"  Shape: {self.shape}\n"
+                f"  Current stride: {self.stride}\n"
+                f"  Expected canonical stride: {canonical_stride}\n"
+                f"Non-canonical strides are not supported for Tripy tensors. "
+                f"This usually occurs when the tensor is not contiguous in memory. "
+                f"To resolve this issue:\n"
+                f"1. Call .contiguous() or .ascontiguousarray() on your tensor before converting to a Tripy tensor.\n"
```
parthchadha, line 86: We should check if the incoming tensor is from Pytorch/Jax/Numpy and suggest appropriate message

tripy/tripy/utils/utils.py
```diff
@@ -214,6 +214,102 @@ def get_shape(data):
...
+
+    Args:
+        shape: The shape of the tensor.
+        stride: The stride to check.
+
+    Returns:
+        True if the stride is canonical, False otherwise.
+
+    Note:
+        Allows arbitrary strides for dimensions with size 0 or 1 to account for framework discrepancies.
```
jhalakpatel, line 304: @parthchadha @pranavm-nvidia It took me a while to figure this out that the stride support in PyTorch and the corresponding DL pack tensor is [broken](https://discuss.pytorch.org/t/strides-are-corrupted-after-to-dlpack-from-dlpack/178255). 
TVM seems to relax the check for dimensions with size = 1 [here](https://github.com/apache/tvm/pull/14797).

jhalakpatel: I think @akhilg-nv also ran into this issue a while back.

parthchadha: Yeah, I think @akhilg-nv's PR in mlir-tensorrt to relax the condition was merged.

tripy/tripy/backend/mlir/memref.py
```diff
@@ -65,7 +66,37 @@ def create_memref_view(data):
...
+        if "torch.Tensor" in obj_type:
+            return "PyTorch Tensor", "tensor.contiguous() or tensor.clone()"
+        elif "jaxlib" in obj_type or "jax.numpy" in obj_type:
+            return "JAX Array", "jax.numpy.asarray(array) or jax.numpy.copy(array)"
+        elif "numpy.ndarray" in obj_type:
+            return "NumPy Array", "np.ascontiguousarray(array) or array.copy(order='C')"
+        elif "cupy.ndarray" in obj_type:
+            return "CuPy Array", "cp.ascontiguousarray(array) or array.copy(order='C')"
+        else:
+            return "Unknown Type", "Cannot suggest method for unknown type"
```
parthchadha, line 82:
```suggestion
            return f"Unknown Type {obj_type}", "Cannot suggest method for unknown type"
```

tripy/tests/utils/test_utils.py
```diff
@@ -109,3 +110,45 @@ def test_gen_uid(self, inputs, outputs, expected_prefix):
...
+            (None, None),
+        ],
+    )
+    def test_get_stride(
+        self,
+        shape: Optional[Sequence[int]],
+        expected_stride: Optional[Sequence[int]],
+    ):
+        if shape is not None:
+            assert utils.get_canonical_stride(shape) == expected_stride
```
pranavm-nvidia, line 138: Looks like this test does nothing if `shape` is `None`. Can we remove those cases and remove this check?

tripy/tripy/backend/mlir/memref.py
pranavm-nvidia, line 82: Can we just return `None` here and drop that entire part of the message if `None` is returned?
```suggestion
            return None, None
```

tripy/tripy/backend/mlir/memref.py
```diff
@@ -65,7 +66,37 @@ def create_memref_view(data):
...
+        canonical_stride = tripy_utils.get_canonical_stride(memref.shape)
+        error_message = (
+            f"Non-canonical strides detected:\n"
+            f"  Shape: {memref.shape}\n"
+            f"  Current stride: {tripy_utils.make_tuple(memref.strides)}\n"
+            f"  Expected canonical stride: {canonical_stride}\n"
+            f"Non-canonical strides are not supported for Tripy tensors. "
+            f"This usually occurs when the tensor is not contiguous in memory. "
+            f"To resolve this issue:\n"
+            f"For {tensor_type}, use {contiguous_suggestion} to ensure contiguity before converting to a Tripy tensor."
```
pranavm-nvidia, line 96:
```suggestion
            f"This usually occurs when the tensor is not contiguous in memory. " + (
            f"To resolve this issue:\n"
            f"For {tensor_type}, use {contiguous_suggestion} to ensure contiguity before converting to a Tripy tensor." 
            if tensor_type is not None else "")
```

tripy/tripy/frontend/tensor.py
```diff
@@ -171,6 +172,9 @@ def dtype(self):
     def rank(self):
         return self.trace_tensor.rank
 
+    def stride(self):
```
parthchadha, line 175: I am not sure why we need to expose stride method for frontend tensors. This information is mostly incorrect since after compilation the strides chosen by TRT for internal tensor can be changed.

parthchadha (approved):

tripy/tripy/frontend/tensor.py
jhalakpatel: Agreed. I have removed the stride field.

tripy/tests/frontend/test_stride.py
```diff
@@ -0,0 +1,68 @@
...
+    def tripy_byte_order_strides(self, data):
+        return tuple(s * data.dtype.itemsize for s in tp.Tensor(data).stride())
+
+    def test_non_canonical_stride(self):
+        # PyTorch test
+        t_torch = torch.arange(12, dtype=torch.float32).reshape(3, 4)
+        a_torch = t_torch.transpose(0, 1)
+        with pytest.raises(tp.TripyException) as excinfo:
+            tp.Tensor(a_torch)
+        self.assert_error_message(excinfo, "PyTorch Tensor", "tensor.contiguous() or tensor.clone()")
```
pranavm-nvidia, line 44: Could we do:
```suggestion
        with raises(tp.TripyException, "Non-canonical strides are not supported for Tripy tensors"):
            tp.Tensor(a_torch)
```
and then remove the `assert_error_message` helper?

pranavm-nvidia: Also, maybe instead of checking with torch/numpy/cupy tensors here, we can just unit test `check_tensor_type_and_suggest_contiguous` separately?

jhalakpatel: Agreed.

pranavm-nvidia (approved):

jhalakpatel:
Closing this PR in favor of https://github.com/NVIDIA/TensorRT-Incubator/pull/273 since this one is running into CI failure due to fork based developement.

PR #243, opened 2024-10-01
Tripy layer metadata

pranavm-nvidia:


pranavm-nvidia:
Once we need this, we should make it on by default for compiled mode and off for eager mode. We probably also want to disable it in most of the unit tests.

pranavm-nvidia:
Closing until we decide to implement this.

PR #248, opened 2024-10-02
[tensorrt] Fix tensorrt ResizeCubic verification

yizhuoz004:
The dimension check is not applied if input/output dimension is dynamic.


shelkesagar29 (approved):
LGTM

PR #252, opened 2024-10-04
[API/MemRef] Implement canonical stride validation for MemRefValue creation

jhalakpatel:
Add optional stride validation in `MemRefValue::create` to compute canonical stride and compare against given strides while creaing a memref view from DLPack tensors. We need to handle special cases for zero-sized and unit-sized dimensions since frameworks deal with them arbitrarily while converting to the corresponding DLPack tensor. Add Python tests to verify both canonical and non-canonical stride validation.

### DIFF
--- a/mlir-tensorrt/executor/lib/Runtime/API/API.cpp
+++ b/mlir-tensorrt/executor/lib/Runtime/API/API.cpp
@@ -671,12 +671,50 @@ static StatusOr<int64_t> getFootprintInBytes(llvm::ArrayRef<int64_t> shape,
   return sizeBytes;
 }
 
+static llvm::SmallVector<int64_t> getCanonicalStride(const llvm::ArrayRef<int64_t>& shape) {
+    if (shape.empty())
+        return {};
+
+    llvm::SmallVector<int64_t> canonicalStride(shape.size(), 1);
+    int64_t cumulativeProduct = 1;
+
+    for (int64_t dimIndex = shape.size() - 1; dimIndex >= 0; --dimIndex) {
+        bool isFirstZeroDim = (shape[dimIndex] == 0 && dimIndex != static_cast<int64_t>(shape.size()) - 1);
+        // For dimensions with size 0 or 1, the stride can be arbitrary.
+        // We set it to 1 here, but other values would also be valid.
+        if (isFirstZeroDim || shape[dimIndex] == 1)
+            canonicalStride[dimIndex] = 1;
+        else
+            canonicalStride[dimIndex] = cumulativeProduct;
+        // For zero-sized dimensions (except the last one), we don't update the cumulative product
+        // This allows for consistent handling of zero-sized dimensions across different frameworks
+        cumulativeProduct *= isFirstZeroDim ? 1 : shape[dimIndex];
+    }
+
+    return canonicalStride;
+}
+
+static bool areStridesEquivalent(llvm::ArrayRef<int64_t> shape,
+                          llvm::ArrayRef<int64_t> stride,
+                          llvm::ArrayRef<int64_t> expectedStride) {
+  if (shape.size() != stride.size() || shape.size() != expectedStride.size())
+    return false;
+
+  for (size_t i = 0; i < shape.size(); ++i)
+    // Allow arbitrary strides for dimensions with size 0 or 1
+    // This accounts for discrepancies in how different frameworks handle these cases
+    if (stride[i] != expectedStride[i] && shape[i] != 0 && shape[i] != 1)
+      return false;
+
+  return true;
+}
+
 StatusOr<std::unique_ptr<MemRefValue>> MemRefValue::create(
     RuntimeClient *client, mlirtrt::runtime::PointerType addressSpace,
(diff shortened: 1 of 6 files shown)


mlir-tensorrt/executor/lib/Runtime/API/API.cpp
```diff
@@ -691,6 +729,17 @@ StatusOr<std::unique_ptr<MemRefValue>> MemRefValue::create(
     return getInvalidArgStatus("a specific device must be provided for MemRefs "
                                "that are device-visible");
 
+  // Check if given strides match canonical stride
+  if (!strides.empty() && !shape.empty()) {
+      std::vector<int64_t> canonicalStride = getCanonicalStride(shape);
+      if (!areStridesEquivalent(shape, strides, canonicalStride)) {
```
christopherbate, line 735: do not merge, I suspect there's some miscommunication here, let's discuss offline

christopherbate (changes requested):

mlir-tensorrt/python/bindings/Runtime/RuntimePyBind.cpp
```diff
@@ -371,11 +371,11 @@ createMemRefViewFromDLPack(PyRuntimeClient &client, py::capsule capsule) {
     s = mtrtMemRefCreateExternal(client, addressSpace, bytesPerElement * 8,
                                  reinterpret_cast<uintptr_t>(data), offset,
                                  rank, shape, strides, device, elementType,
-                                 &result);
+                                 &result, true /*assertCanonicalStrides*/);
```
christopherbate, line 374: Why are we forcing this to true here? It should be a a parameter to the function.

christopherbate (changes requested):
The assertion has to be a parameter to the Python function as well, not just the C++ function.

mlir-tensorrt/python/bindings/Runtime/RuntimePyBind.cpp
jhalakpatel: I have updated the PR to add Python API assertion check option.

mlir-tensorrt/test/python/mlir_tensorrt_runtime/test_create_memref.py
```diff
@@ -514,3 +514,82 @@ def create_dangling_memref():
...
+
+    print("Testing canonical stride: assert_canonical_strides = False")
+    canonical_result = check_canonical_stride(client, assert_canonical_strides=False)
+
+
+print("Test memref strides")
+test_memref_strides()
+
+# CHECK-LABEL: Test memref strides
+# CHECK-NEXT: Testing non-canonical stride: assert_canonical_strides = True
```
christopherbate, line 587: It looks like the error message is not being correctly matched. Maybe just print out the error message, e.g.

```# CHECK-NEXT: Testing non-canonical stride: assert_canonical_strides = True
# CHECK-NEXT: Received error message: .... specific message string ....```

No reason to do string matching in the Python exception, just do it in the `CHECK` statements?

File: mlir-tensorrt/executor/lib/Runtime/API/API.cpp [hunk 4]
```diff
@@ -691,6 +729,19 @@ StatusOr<std::unique_ptr<MemRefValue>> MemRefValue::create(
     return getInvalidArgStatus("a specific device must be provided for MemRefs "
                                "that are device-visible");
 
+  // Check if given strides match canonical stride
+  if (assertCanonicalStrides && *assertCanonicalStrides) {
+    std::vector<int64_t> canonicalStride = getCanonicalStride(shape);
```
christopherbate, line 734: It seems worthwhile to just combine the `getCanonicalStride` function and the `areStridesEquivalent` to avoid creating the `std::vector`, especially since speed was your concern for moving this check to C++.

mlir-tensorrt/executor/lib/Runtime/API/API.cpp
```diff
@@ -671,12 +671,50 @@ static StatusOr<int64_t> getFootprintInBytes(llvm::ArrayRef<int64_t> shape,
   return sizeBytes;
 }
 
+static std::vector<int64_t> getCanonicalStride(const llvm::ArrayRef<int64_t>& shape) {
```
christopherbate, line 674: Can use `llvm::SmallVector` here

christopherbate (approved):
Some minor comments, otherwise LGTM.

mlir-tensorrt/test/python/mlir_tensorrt_runtime/test_create_memref.py
jhalakpatel: Agreed. I have updated to only use CHECK statements.

mlir-tensorrt/executor/lib/Runtime/API/API.cpp
jhalakpatel: We want to print the expected canonical stride in the error message. I don't think we can avoid creating an extra vector copy.

PR #253, opened 2024-10-04
[test] Corrects checks in layer metadata callback test

pranavm-nvidia:
Corrects checks in layer metadata callback test


shelkesagar29 (approved):
LGTM. I also tested this locally.

PR #254, opened 2024-10-04
Cast fp16 to fp32 in tolist, remove FP16MemoryView

yizhuoz004:


tripy/tests/common/test_utils.py
```diff
@@ -60,33 +59,3 @@ def test_get_element_type():
 def convert_list_to_array(values, dtype, expected):
```
pranavm-nvidia, line 59: Not part of your change, but this actually wouldn't have been running as a test because it's missing the `test_` prefix
```suggestion
def test_convert_list_to_array(values, dtype, expected):
```

pranavm-nvidia (approved):

yizhuoz004:
@pranavm-nvidia I found we still need to print `memref` in a few places, this requires `Float16MemoryView`. Actually `fp16` is the only precision that `memoryview` would raise an exception, other dtypes (bfloat16, int8) returns a nested list with data represented as uint8.

tripy/tripy/flat_ir/flat_ir.py
```diff
@@ -352,22 +352,11 @@ def register_tensor(self, tensor: "FlatIRTensor") -> "FlatIRTensor":
...
-            if volume(op.data.shape) < VOLUME_THRESHOLD_FOR_MEMREF:
-                l = tolist(op.data)
-            else:
-                l = [op.data.ptr]
-            data = list_to_tuple(l if isinstance(l, List) else [l])
-        elif isinstance(op.data, int) or isinstance(op.data, float) or isinstance(op.data, bool):
-            data = list_to_tuple(
-                op.data,
-            )
+            data = (op.data.ptr,)
```
pranavm-nvidia, line 359: Won't this prevent us from deduplicating small constants?

yizhuoz004: It will, but practically, if constant data is given by `MemrefValue`, it comes from users (such as weights & inputs). The small constants we want to deduplicate are generated internally where data is a list.

pranavm-nvidia: Ah makes sense. Could you add a comment here to that effect?

pranavm-nvidia (approved):

tripy/tripy/flat_ir/flat_ir.py
yizhuoz004: Added, thanks!

PR #255, opened 2024-10-04
[mlir-tensorrt] Add commit message check to CI

shelkesagar29:
This MR adds commit message check in `mlir-tensorrt` CI. If commit message check fails, it prints out
commit message and canonical format expected.

### DIFF
--- a/.github/workflows/mlir-tensorrt-ci.yml
+++ b/.github/workflows/mlir-tensorrt-ci.yml
@@ -43,7 +43,7 @@ jobs:
           fetch-depth: 5
 
       # Run initial format check
-      - name: Run python format and clang check
+      - name: Run commit message, python format, and c++ clang format check
         uses: addnab/docker-run-action@v3
         with:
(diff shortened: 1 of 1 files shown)


shelkesagar29:
Tried regex match using `bash` (grep and sed) , to avoid new file creation but multiline match with regex we have doesn't work.

pranavm-nvidia:
@shelkesagar29 you could add this as a git hook (`prepare-commit-msg` or `comit-msg`) instead, which would allow us to catch it preemptively.

shelkesagar29:
> @shelkesagar29 you could add this as a git hook (`prepare-commit-msg` or `comit-msg`) instead, which would allow us to catch it preemptively.

But we want this check just for `miler-tensorrt` commits and not for every commit.
Also, this will need checking regex in bash. Regex we have is a multiline match which is not working in bash. Thus, I am doing it now in python

yizhuoz004 (approved):

PR #258, opened 2024-10-08
[mlir-tensorrt] Add support for non-DPS calling convention

jhalakpatel:
This commit introduces support for a non-Destination-Passing Style (non-DPS) calling convention in mlir-tensorrt, while maintaining the existing DPS-style interface.

The changes aim to allow users to compile and execute a mlir-tensorrt executable without allocating output memrefs in advance. Removing the output memref allocation restriction alleviates users from computing output shapes in advance and allocating output memrefs. This is critical since it is part of the performance-critical execution loop.

Deferred output allocation has an added advantage as we no longer need to allocate shape upper bound output buffers and copy the exact output buffer from TensorRT results into mlir-tensorrt output buffers.

This approach also allows us to support data-dependent shapes since the outputs are not required to be allocated before execution.

The non-DPS style calling convention implementation leverages the `nvinfer1::IOutputAllocator` interface for deferred output allocation. Interface functions such as `reallocateOutputAsync` and `notifyShapes` store the allocated output buffer address and its shape.

User call sites with existing DPS-style calling convention:

```
        out_args = ...  # Requires output shape calculation and output buffer allocation before execution.
        session.execute_function(
            "main", in_args=in_args, out_args=out_args, stream=self.stream._active_cuda_stream
        )
```

Now, we can do the following with non-DPS style convention:
```
        outputs = session.execute_function(
            "main", in_args=in_args, stream=self.stream._active_cuda_stream
        )
```

Key changes include:

1. Update the `Plan` dialect to implement non-DPS calling convention:
   - Updated `PlanAllocTensorsPass` and `CreateClosedRegionsPass` to handle non-DPS convention
   -  Update Plan transformation to convert DPS and non-DPS group ops to `CallAllocOp` and `CallOp` respectively.
2. Enhanced TensorRT runtime support:
   - Modified `ConvertTensorRTToTensorRTRuntime` to support both calling conventions
   - Added `EnqueueAllocOp` for non-DPS style execution
   - Added conversion from `EnqueueAllocOp` to executor `CallOp`.
   - Implemented `OutputAllocator` and `CustomTensorRTOuputAllocator` classes with proper lifetime management using `OutputAllocatorTracker`.
3. Updated API and runtime interfaces:
   - Modified `executeFunctionWithLuaBackend` to support both DPS and non-DPS styles
   - Updated Python bindings to allow non-DPS styles.


mlir-tensorrt/executor/include/mlir-executor-c/Runtime/Runtime.h
```diff
@@ -375,15 +386,22 @@ static inline bool mtrtRuntimeSessionIsNull(MTRT_RuntimeSession session) {
...
 
 /// Using `session`, execute the pubic function with the specified name.
-/// The `inArgs` and `outArgs` are arrays for input arguments and destination
-/// arguments, respectively. Input arguments may be MemRefs or scalars, but
-/// destination arguments must be MemRefs.
-/// A stream may optionally be specified, otherwise pass the result of
-/// `mtrtStreamGetNull()`.
+/// The `inArgs`, `outArgs`, and `results` are arrays for input arguments and
+/// optional destination arguments and results, respectively. Input arguments
+/// may be MemRefs or scalars, but destination arguments and results must be
```
christopherbate, line 391: It should not be required that `results` are MemRefs. This is not required by the compiler. For example, the functions we often use for testing with `mlir-tensorrt-runner` usually return an integer code a'la normal C/C++ `main` functions.

mlir-tensorrt/executor/include/mlir-executor-c/Runtime/Runtime.h
```diff
@@ -375,15 +386,22 @@ static inline bool mtrtRuntimeSessionIsNull(MTRT_RuntimeSession session) {
...
+/// MemRefs. If `outArgs` are present, it is expected that `results` are empty
```
christopherbate, line 392: Similarly, you can have both "output arguments" and "results", just like how in C++ you can have a function like 

```
int my_func(int in, int *out);
```

mlir-tensorrt/executor/include/mlir-executor-c/Runtime/Runtime.h
```diff
@@ -375,15 +386,22 @@ static inline bool mtrtRuntimeSessionIsNull(MTRT_RuntimeSession session) {
...
+/// and vice-versa. A stream may optionally be specified, otherwise pass the
+/// result of `mtrtStreamGetNull()`.
 MLIR_CAPI_EXPORTED MTRT_Status mtrtRuntimeSessionExecuteFunction(
     MTRT_RuntimeSession session, MTRT_StringView name,
     const MTRT_RuntimeValue *inArgs, size_t numInArgs,
-    const MTRT_RuntimeValue *outArgs, size_t numOutArgs, MTRT_Stream stream);
+    const MTRT_RuntimeValue *outArgs, size_t numOutArgs,
+    MTRT_RuntimeValue *results, size_t numResults, MTRT_Stream stream,
```
christopherbate, line 399: It shouldn't be necessary to pass `numResults` here. You can just say that `results` must point to an array of minimum size given by `mtrtRuntimeSessionGetNumResults`.

mlir-tensorrt/executor/include/mlir-executor-c/Runtime/Runtime.h
```diff
@@ -375,15 +386,22 @@ static inline bool mtrtRuntimeSessionIsNull(MTRT_RuntimeSession session) {
...
+    MTRT_RuntimeClient client);
+
```
christopherbate, line 401: It might be better to expose an API entirely for the function signature.

mlir-tensorrt/executor/include/mlir-executor/Runtime/API/API.h
```diff
@@ -120,7 +120,7 @@ std::string_view stringifyPointerType(PointerType ptrType);
 //===----------------------------------------------------------------------===//
 // TypeView
 // This section includes classes that form the TypeUnion:
-// MemrefTypeView, ScalarTypeView, ExternalOpaqueTypeView
+// MemRefTypeView, ScalarTypeView, ExternalOpaqueTypeView
```
christopherbate, line 123: Please make NFC rename changes a separate PR or at least a separate commit in the PR. Isolating NFC changes is always better since it lowers the review burden.

mlir-tensorrt/executor/include/mlir-executor/Support/Allocators.h
```diff
@@ -128,6 +128,94 @@ class PinnedMemoryAllocator {
   std::unique_ptr<BlockEventQueue> pendingBlockEvents;
 };
 
+//===----------------------------------------------------------------------===//
+// OutputAllocator and CustomTensorRTOuputAllocator
+//===----------------------------------------------------------------------===//
+
+//!
+//! Class to allocate memory for outputs with data-dependent shapes. The sizes
+//! of those are unknown so pre-allocation is not possible.
```
christopherbate, line 137: We use triple-forward-slash `///` for documentation comments

mlir-tensorrt/executor/include/mlir-executor/Support/Allocators.h
```diff
@@ -128,6 +128,94 @@ class PinnedMemoryAllocator {
...
+  virtual void setOutputSize(const int64_t outputSize) = 0;
+  virtual void *reallocateOutputAsync(char const *tensorName,
+                                      void *currentMemory, uint64_t size,
+                                      uint64_t alignment,
+                                      CudaStream /*stream*/) = 0;
+  virtual void notifyShape(char const *tensorName, const int64_t *dims,
+                           int64_t nbDims) = 0;
+};
+
+
```
christopherbate, line 153: It's not clear why we need these two classes instead of one class. Why do we need all these different methods? Similarly, what is the tracker class for?

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -360,6 +421,25 @@ static Status enqueueV3Wrapper(AllocTracker &tracker,
     return getStatusWithMsg(StatusCode::InternalError,
                             "failed to set input-consumed event");
 
+  if (hasResults) {
+    int64_t* desc = reinterpret_cast<int64_t*>(*outputDescriptors);
```
christopherbate, line 425: Somewhere in this file you'll have to specifically document the ABI for the output descriptors. Something with a bit more structure to wrap the `uintptr_t` would be great, e.g. `struct OutputDescriptors` could wrap the uinptr_t and provide methods for retrieving the number of results and information about each output, setting the shape/strides/ptr, etc.

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -341,10 +399,13 @@ prepareBuffers(const AllocTracker &allocTracker,
 
 static Status enqueueV3Wrapper(AllocTracker &tracker,
```
christopherbate, line 400: This function got a lot more complex with the addition of `outputDescriptors`. I think we should just have two separate functions -- keep the original (which wasn't very big anyway) and just have a separate function for the situation where TensorRT is going to allocate the outputs.

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -369,6 +449,27 @@ static Status enqueueV3Wrapper(AllocTracker &tracker,
...
+  if (hasResults) {
+    int64_t* desc = reinterpret_cast<int64_t*>(*outputDescriptors);
+
+    // Number of results are known in advance.
+    int64_t descIndex = 0;
+    int64_t nbResults = desc[descIndex++];
+    for (int64_t i = 0; i < nbResults; ++i) {
+      auto allocator = static_cast<CustomTensorRTOuputAllocator *>(
+          outputAllocatorTracker.get(i));
+      desc[descIndex++] = allocator->mOutputDims.size(); // rank
```
christopherbate, line 461: Rank is always statically known. The compiler should generate executor IR to pre-populate the rank field, and this line should simply become a validation point -- read the pre-specified rank and compare with `mOutputDims.size()`. If they are different, then we should return an error, especially since the compiler generated an allocation that is only large enough to hold results with a specific rank.

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -369,6 +449,27 @@ static Status enqueueV3Wrapper(AllocTracker &tracker,
...
+      desc[descIndex++] =
+          reinterpret_cast<uintptr_t>(allocator->mOutputPtr); // output pointer
+      std::copy_n(allocator->mOutputDims.data(), allocator->mOutputDims.size(),
+                  desc + descIndex); // Copy input shape
+      descIndex += allocator->mOutputDims.size();
+      std::vector<int64_t> strides(allocator->mOutputDims.size(), 1); // For now, use 1 strides.
```
christopherbate, line 467: This definitely isn't right. That will cause runtime errors for anything that tries to load from this buffer. What is the stride conventions for data allocated by TensorRT? Is it canonical strides or are there are specific set of layouts that may be used?

mlir-tensorrt/executor/include/mlir-executor/Support/Allocators.h
jhalakpatel: I copied this code from one of my Draft MR's. Class `OutputAllocator` here defines the interface that can be overridden by the user (aka. Tripy) to provide the implementation. 
At the same time, we would like to provide a default implementation in the compiler i.e. `CustomTensorRTOutputAllocator`.

jhalakpatel: `OutputAllocatorTracker` binds the lifetime of `OutputAllocator` and allocates the memory it owns to the session.
May be it should remain alive as long as `RuntimeClient` is alive. I am open to suggestions here.

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
jhalakpatel: That is a good idea. I will address this in next commit.

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
jhalakpatel: Looks like MLIR-TRT does not use `nvinfer1::setAllowedFormat` API which allows formats other than `nvinfer1::TensorFormat::kLINEAR`. So for all practical purposes, stride here is canonical.

jhalakpatel: I think we can use `IExecutionContext::getTensorStrides` API to retrieve stride information.

mlir-tensorrt/compiler/include/mlir-tensorrt/Dialect/TensorRTRuntime/IR/TensorRTRuntimeOps.td
```diff
@@ -116,4 +116,57 @@ def TensorRTRuntime_EnqueueOp : TensorRTRuntime_Op<"enqueue", [
   }];
 }
 
+//===----------------------------------------------------------------------===//
+// EnqueueAllocOp
+//===----------------------------------------------------------------------===//
+
+def Output_Desc : TypeDef<CUDA_Dialect, "Output_Desc", []> {
```
christopherbate, line 123: This definitely should not be required

mlir-tensorrt/compiler/lib/Dialect/TensorRTRuntime/IR/TensorRTRuntime.cpp
```diff
@@ -113,6 +113,62 @@ void EnqueueOp::getEffects(
...
+                           "but input arg ")
+               << idx << " has type " << operand.getType();
+    }
+  }
+
+  // Verify that all results are either all tensors or all memrefs
+  if (getNumResults() > 0) {
+    bool allTensors = getResult(0).getType().isa<TensorType>();
+    for (auto result : getResults()) {
+      if (result.getType().isa<TensorType>() != allTensors) {
```
christopherbate, line 143: - `Type::isa` is deprecated, use `llvm::isa<RankedTensorType>(result.getType())`. 

- Use `RankedTensorType`. `TensorType` allows for unranked Tensors, which we do not allow here. 

- You can iterate over types with `getResultTypes()`.

- Drop braces on single-statement `if` body: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements

mlir-tensorrt/compiler/lib/Dialect/TensorRTRuntime/IR/TensorRTRuntime.cpp
```diff
@@ -113,6 +113,62 @@ void EnqueueOp::getEffects(
...
+      Type elType = mlir::getElementTypeOrSelf(operand.getType());
+      if (!elType.isInteger(32))
+        return emitOpError("host tensor arguments must have element type i32, "
+                           "but input arg ")
+               << idx << " has type " << operand.getType();
+    }
+  }
+
+  // Verify that all results are either all tensors or all memrefs
+  if (getNumResults() > 0) {
```
christopherbate, line 140: Zero results should be a verification error. At least one result is required.

mlir-tensorrt/compiler/lib/Dialect/TensorRTRuntime/IR/TensorRTRuntime.cpp
```diff
@@ -113,6 +113,62 @@ void EnqueueOp::getEffects(
...
+    bool allTensors = getResult(0).getType().isa<TensorType>();
+    for (auto result : getResults()) {
+      if (result.getType().isa<TensorType>() != allTensors) {
+        return emitOpError("all results must be of the same type (all tensors "
+                           "or all memrefs)");
+      }
```
christopherbate, line 146: We should also be verifying the layout (stride + offset) information on memref results, e.g. 

```
trtrt.alloc_enqueue .... -> memref<?x?x?xf32>
```

implies the identity layout (canonical row major strides).

```
trtrt.alloc_enqueue .... -> memref<?x?x?xf32, strided<[?, ?, ?, ?], offset: ?>>
```

indicates that the strides are unknown.

Not being able to know anything about the strides is very worst-case since it disables many possible optimizations.

 We need this information from TensorRT -- what layouts of results are possible/allowed? Can we enforce that canonical strides will always be returned from TensorRT using the output allocator? If we can, then the verifier should enforce that canonical layouts are used. 

Currently for `trtrt.enqueue` we are effectively enforcing canonical strides for input and output buffers.

mlir-tensorrt/compiler/lib/Dialect/TensorRTRuntime/IR/TensorRTRuntime.cpp
```diff
@@ -113,6 +113,62 @@ void EnqueueOp::getEffects(
...
+    }
+  }
+  return success();
+}
+
+void EnqueueAllocOp::getEffects(
+    SmallVectorImpl<SideEffects::EffectInstance<MemoryEffects::Effect>>
+        &effects) {
+  // This op allocates memory for its results
```
christopherbate, line 155: Memory effects are restricted to the case where the op has all memref types. If operands/results have tensor types, then just return from this function before modifying `effects` (don't add allocation effect either).

mlir-tensorrt/compiler/lib/Dialect/TensorRTRuntime/Transforms/BufferizableOpInterfaceImpl.cpp
```diff
@@ -153,18 +160,144 @@ struct EnqueueOpInterface
...
+    auto enqueueAllocOp = cast<EnqueueAllocOp>(op);
+    MLIRContext *ctx = op->getContext();
+    Location loc = op->getLoc();
+    rewriter.setInsertionPoint(enqueueAllocOp);
+
+    // Handle inputs
+    SmallVector<Value> newInputBuffers;
+    newInputBuffers.reserve(enqueueAllocOp.getInputs().size());
+    for (auto [idx, input] : llvm::enumerate(enqueueAllocOp.getInputs())) {
+      if (isScalar(&input)) {
```
christopherbate, line 220: This operation cannot have scalar inputs by construction. Values in `getInputs()` can only be tensors or memrefs. This can just be

```
assert(isa<RankedTensorType>(input.getType()) && "expected tensor inputs");
```

mlir-tensorrt/compiler/lib/Dialect/TensorRTRuntime/Transforms/BufferizableOpInterfaceImpl.cpp
```diff
@@ -153,18 +160,144 @@ struct EnqueueOpInterface
...
+    outputBufferTypes.reserve(enqueueAllocOp.getNumResults());
+    for (unsigned i = 0; i < enqueueAllocOp.getNumResults(); ++i) {
+      Type resultType = enqueueAllocOp->getResultTypes()[i];
+      MemRefType memRefType;
+      if (auto tensorType = resultType.dyn_cast<TensorType>()) {
+        memRefType = MemRefType::get(tensorType.getShape(), tensorType.getElementType(),
+                                     MemRefLayoutAttrInterface(),
+                                     plan::MemorySpaceAttr::get(ctx, plan::MemorySpace::device));
+      } else {
+        memRefType = resultType.cast<MemRefType>();
```
christopherbate, line 280: I think this case should be an error/assertion. We shouldn't get to this point in the function.

mlir-tensorrt/compiler/lib/Dialect/TensorRTRuntime/Transforms/BufferizableOpInterfaceImpl.cpp
```diff
@@ -153,18 +160,144 @@ struct EnqueueOpInterface
...
+      newInputBuffers.push_back(*buffer);
+    }
+
+    // Handle results
+    SmallVector<Type> outputBufferTypes;
+    outputBufferTypes.reserve(enqueueAllocOp.getNumResults());
+    for (unsigned i = 0; i < enqueueAllocOp.getNumResults(); ++i) {
+      Type resultType = enqueueAllocOp->getResultTypes()[i];
+      MemRefType memRefType;
+      if (auto tensorType = resultType.dyn_cast<TensorType>()) {
```
christopherbate, line 275: Don't use `Type::dyn_cast`, use free function `llvm::dyn_cast`.

mlir-tensorrt/compiler/lib/Dialect/TensorRTRuntime/Transforms/BufferizableOpInterfaceImpl.cpp
```diff
@@ -153,18 +160,144 @@ struct EnqueueOpInterface
...
+                                     MemRefLayoutAttrInterface(),
+                                     plan::MemorySpaceAttr::get(ctx, plan::MemorySpace::device));
+      } else {
+        memRefType = resultType.cast<MemRefType>();
+      }
+      outputBufferTypes.push_back(memRefType);
+    }
+
+    // Create the new operation
+    rewriter.create<EnqueueAllocOp>(
```
christopherbate, line 286: You are missing some sort of API call to tell the bufferization driver that you are remapping the results of `op` to the results of this new op you are creating here.

Since you didn't do that, it i creating a `memref.alloc` to fill in the gap when it traverses later consumers.

I believe it should look like 

```
auto bufferizedOp = rewriter.create<EnqeueAllocOp>(...);
replaceOpWithBufferizedValues(rewriter, op, bufferizedOp.getResults());
```

mlir-tensorrt/compiler/lib/Dialect/TensorRTRuntime/Transforms/BufferizableOpInterfaceImpl.cpp
```diff
@@ -49,16 +56,16 @@ struct EnqueueOpInterface
   /// outputs in our use-case.
   bool bufferizesToMemoryRead(Operation *op, OpOperand &opOperand,
                               const bufferization::AnalysisState &state) const {
-    EnqueueOp callOp = cast<EnqueueOp>(op);
```
christopherbate, line 52: NFC refactor changes should go in a separate PR.

mlir-tensorrt/compiler/lib/Dialect/TensorRTRuntime/Transforms/BufferizableOpInterfaceImpl.cpp
```diff
@@ -41,6 +41,13 @@ static bool isInMemorySpace(Type memrefType, plan::MemorySpace memType) {
   return space == plan::MemorySpaceAttr::get(type.getContext(), memType);
 }
 
+static bool isScalar(Value *input) {
```
christopherbate, line 44: Don't pass Value by pointer/reference unless it's an output parameter. 

This unction isn't needed as discussed below. 

Also, `isa` can take multiple template parameters to express a disjunction over possible types e.g. `isa<TensorType, MemRefType>(input.getType())`

christopherbate (changes requested):

mlir-tensorrt/compiler/include/mlir-tensorrt/Dialect/TensorRTRuntime/IR/TensorRTRuntimeOps.td
jhalakpatel: Oops. I forgot to delete this.

mlir-tensorrt/compiler/lib/Dialect/TensorRTRuntime/Transforms/BufferizableOpInterfaceImpl.cpp
jhalakpatel: Sure. I will club all the NFC changes together in a separate PR. I am making changes in the same PR to track changes as of now.

mlir-tensorrt/executor/include/mlir-executor/Runtime/API/API.h
jhalakpatel: Sure. I will do so.

mlir-tensorrt/compiler/lib/Dialect/TensorRTRuntime/IR/TensorRTRuntime.cpp
jhalakpatel: As I mentioned in the other thread, MLIR-TRT does not use the `nvinfer1::setAllowedFormat` API, which allows formats other than nvinfer1::TensorFormat::kLINEAR. So, for all practical purposes, stride here is canonical.
I will add a check in the verifier for canonical strides. Let me know if you meant this as I might have misunderstood you.

Also, How can I generate this assembly format: `memref<?x?x?xf32, strided<[?, ?, ?, ?], offset: ?>>`?

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/Options.h
```diff
@@ -34,6 +34,9 @@ namespace mlirtrt::compiler {
 /// DebugOptions are options that are common to different compiler API
 /// interfaces.
 struct DebugOptions {
+  /// Dump textual pipeline passes
+  bool dumpTextualPipeline = false;
```
jhalakpatel, line 38: @christopherbate I will move such changes to NFC PR's once we finalize the implementation.

mlir-tensorrt/executor/lib/Conversion/MemRefToExecutor.cpp
```diff
@@ -548,6 +550,43 @@ void executor::populateMemRefToExecutorPatterns(
 }
 
 namespace {
+
+class RemoveNoOpClonePattern : public OpRewritePattern<bufferization::CloneOp> {
```
christopherbate, line 554: This is illegal. Clone indicates creating a copy, which you are not doing here. This indicates something else is missing in the pipeline before this point.

mlir-tensorrt/executor/lib/Conversion/MemRefToExecutor.cpp
```diff
@@ -548,6 +550,43 @@ void executor::populateMemRefToExecutorPatterns(
...
+                                PatternRewriter &rewriter) const override {
+    if (op.getInput().getType() == op.getOutput().getType()) {
+      rewriter.replaceOp(op, op.getInput());
+      return success();
+    }
+    return failure();
+  }
+};
+
+class RemoveRedundantCastsPattern
```
christopherbate, line 568: This should also be removed. Requiring this means something is wrong before running this pass.

christopherbate (changes requested):

mlir-tensorrt/compiler/include/mlir-tensorrt/Dialect/Plan/IR/PlanOps.td
```diff
@@ -131,10 +131,57 @@ def Plan_InlineGroupOp : Plan_GroupOpBase<"inline_group", [
...
+
+    /// Populate the `input_attrs` from an array of BoundsAttrs.
+    void setInputAttrsAttr(ArrayRef<BoundsAttr> boundsAttrs) {
+      setInputAttrsAttr(::mlir::ArrayAttr::get(
+        getOperation()->getContext(),
+        ArrayRef<Attribute>(boundsAttrs.begin(), boundsAttrs.end())
+      ));
+    }
+
+    void getSuccessorRegionsBase(RegionBranchPoint point,
```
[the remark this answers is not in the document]
jhalakpatel: Sure. Given those methods need to access functionality from the base class, I am just moving them to separate class methods instead of a common implementation. Their implementation is only a few lines so it is not worth deduplicating.

mlir-tensorrt/compiler/include/mlir-tensorrt/Dialect/TensorRTRuntime/IR/TensorRTRuntimeOps.td
```diff
@@ -116,4 +116,49 @@ def TensorRTRuntime_EnqueueOp : TensorRTRuntime_Op<"enqueue", [
...
+    `(` $inputs `)`
+    attr-dict `:` functional-type($inputs, $results)
+  }];
+
+  let hasVerifier = 1;
+
+  let extraClassDeclaration = [{
+    /// Return true if the operand at the specified index is a host tensor
+    /// argument.
+    bool isOperandOnHost(int64_t operandIdx) {
```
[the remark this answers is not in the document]
jhalakpatel: I agree. Since I am anyway breaking this PR into smaller ones, I will update the other op as well.

mlir-tensorrt/compiler/lib/Conversion/TensorRTToTensorRTRuntime/TensorRTToTensorRTRuntime.cpp
```diff
@@ -57,6 +57,79 @@ struct RewriteConstants : public OpRewritePattern<tensorrt::ConstantOp> {
...
+    outputs = callOp.getOutputs();
+  } else if (auto callAllocOp = dyn_cast<tensorrt::CallAllocOp>(op)) {
+    trtFunc = callAllocOp.getFuncCallee(symbolTable);
+    inputs = callAllocOp.getInputs();
+    // CallAllocOp doesn't have outputs as operands
+  } else {
+    return failure();
+  }
+
+  solver.load<dataflow::DeadCodeAnalysis>();
```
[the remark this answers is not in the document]
jhalakpatel: I will just move them out of the function and loop in this PR.

mlir-tensorrt/executor/lib/Conversion/MemRefToExecutor.cpp
jhalakpatel: Sure. This is not required.

mlir-tensorrt/executor/lib/Conversion/MemRefToExecutor.cpp
jhalakpatel: @christopherbate Let me know if you any suggestion on how to fix this.

jhalakpatel:
This PR will eventually be closed as it has been broken down to smaller PRs:

Non-DPS style functional changes are open for review. I have broken them down to more manageable PR's for review:
1. Output allocator in TensorRT Module: https://github.com/NVIDIA/TensorRT-Incubator/pull/282
2. Some improvements to variable names: https://github.com/NVIDIA/TensorRT-Incubator/pull/283
3. Adding an option to dump textual pipeline: https://github.com/NVIDIA/TensorRT-Incubator/pull/284
4. Update Plan dialect to support non-DPS calling convention: https://github.com/NVIDIA/TensorRT-Incubator/pull/285
5. Update TensorRT runtime dialect to use non-dps calling convention: https://github.com/NVIDIA/TensorRT-Incubator/pull/286
6. Update TensorRT dialect: https://github.com/NVIDIA/TensorRT-Incubator/pull/287
7. Update Lua Runtime: https://github.com/NVIDIA/TensorRT-Incubator/pull/288

pranavm-nvidia:
Closing stale PRs

PR #261, opened 2024-10-08
Update pyproject.toml to use 0.1.34 MLIR-TRT

yizhuoz004:


yizhuoz004:
```
FAILED tests/spec_verification/test_dtype_constraints.py::test_dtype_constraints[__mod__-invalid:T1=float8]
FAILED tests/spec_verification/test_dtype_constraints.py::test_dtype_constraints[__rfloordiv__-invalid:T1=float8]
FAILED tests/spec_verification/test_dtype_constraints.py::test_dtype_constraints[__rmod__-invalid:T1=float8]
FAILED tests/spec_verification/test_dtype_constraints.py::test_dtype_constraints[__floordiv__-invalid:T1=float8]
```

@pranavm-nvidia Do you know how the type constraints can fail here? I tested locally that float8 will raise the unsupported dtype exception, but the pytest just fails.

pranavm-nvidia:
> @pranavm-nvidia Do you know how the type constraints can fail here? I tested locally that float8 will raise the unsupported dtype exception, but the pytest just fails.

@yizhuoz004 the exceptions you're seeing locally are probably the ones raised by the constraints decorator itself. When we do dtype constraint testing, we disable those - `with helper.config("enable_dtype_checking", False)` (otherwise the test would be pointless). Looks like FP8 is now working for these operations, so I guess we should update the constraints accordingly?

pranavm-nvidia (approved):

Issue #233, opened 2024-09-27
Disallow mixing Tensor/Shape types in operations


pranavm-nvidia:
The behavior of some operations, e.g. binary elementwise, is not obvious when mixing `Tensor`/`Shape` types. We may want to disallow mixing types in such cases. See https://github.com/NVIDIA/TensorRT-Incubator/pull/224#discussion_r1777700616 for more context.

pranavm-nvidia:
Other TODOs:
- 
- Rename/refactor `infer_shape_output_idxs` to better reflect it's purpose of also reporting on `ShapeScalar`s.

slyubomirsky:
We might want to decide on the terminology to use here, for what to call the variants of `Tensor`. "Tensor types" might work, but it seems too easily confused with datatypes. Maybe "tensor variants"?

Currently, the vast majority of uses of `convert_inputs_to_tensors` are for binary elementwise ops (the only other is for `slice_helper`), where combining `Shape`s with `Tensor`s could cause confusion. (By contrast, I doubt that combining `ShapeScalar`s with `Tensor`s or `Shape`s is likely to cause confusion). Since there are relatively few uses, perhaps we could just have a flag for disallowing combining different tensor variants? The error could give a hint to have an explicit conversion if that's what the user really wants (which is what we do for several of the `Shape` methods already). The downside is that it might be a little clumsy to require explicit conversions, but I'm not sure there is any reason to believe that combining `Tensor`s and `Shape`s is something that would be commonly needed. (We could see if it happens in any of our examples.)

Incidentally, for `slice_helper`, I would argue that mixing tensor variants is unlikely to be indicative of an error, since the decorator would only be checking the _index_ arguments.

slyubomirsky:
Closing because we eliminated `Shape`s altogether.

PR #265, opened 2024-10-11
[Tripy] Disallow passing both Shapes and Tensors with the `convert_inputs_to_tensors` decorator

slyubomirsky:
Addresses issue #233. Per discussions, there are not many remaining uses of the `convert_inputs_to_tensors` decorator in the codebase and it also seems that in just about all of those cases, it would not make sense to accept both shapes and tensors.

### DIFF
--- a/tripy/tripy/frontend/ops/repeat.py
+++ b/tripy/tripy/frontend/ops/repeat.py
@@ -95,6 +95,6 @@ def repeat(input: "tripy.Tensor", repeats: Union[int, "tripy.ShapeScalar"], dim:
     repeat_mask = concatenate(
         [reshape(repeats, (1,)) if idx == dim else Tensor([1]) for idx in range(input.rank)], dim=0
     )
-    new_shape = input.shape.multiply(repeat_mask)
+    new_shape = input.shape.multiply(Shape(repeat_mask))
     out = reshape(out, new_shape)
     return out
--- a/tripy/tripy/frontend/trace/ops/binary_elementwise.py
+++ b/tripy/tripy/frontend/trace/ops/binary_elementwise.py
@@ -50,23 +50,24 @@ class BinaryElementwise(BaseTraceOp):
         return f"{self.outputs[0].name} = {op_str}"
 
     def infer_shape_output_idxs(self, inputs):
-        # permit one input to be a shape but require the output to be a shape
+        # To avoid issues, we do not permit both Shape and Tensor args and recommend explicit conversion if they do need to combined.
+        # ShapeScalar args can be broadcast up into Shapes and so do not pose an issue.
         from tripy.frontend.shape import Shape, ShapeScalar
         from tripy.utils import Result
 
         if any(map(lambda t: isinstance(t, Shape), inputs)):
-            # if there is a non-shape input, it must be rank 1 or 0 to avoid broadcasting
-            if not all(map(lambda t: t.rank <= 1, inputs)):
-                invalid_indices_message = ", ".join(
-                    map(
-                        lambda i: f"Index {i} (rank {inputs[i].rank})",
-                        filter(lambda i: inputs[i].rank > 1, range(len(inputs))),
-                    )
-                )
+            # Note: Most binary elementwise ops use the convert_inputs_to_tensors decorator, which also checks for this.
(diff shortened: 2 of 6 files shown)


slyubomirsky:
I will handle renaming `infer_shape_output_idxs` separately, since it has turned out not to be related to the decorator.

tripy/tripy/frontend/ops/repeat.py
```diff
@@ -95,6 +95,6 @@ def repeat(input: "tripy.Tensor", repeats: Union[int, "tripy.ShapeScalar"], dim:
     repeat_mask = concatenate(
         [reshape(repeats, (1,)) if idx == dim else Tensor([1]) for idx in range(input.rank)], dim=0
```
pranavm-nvidia, line 96: Does changing this part to:
```suggestion
        [reshape(repeats, (1,)) if idx == dim else Shape([1]) for idx in range(input.rank)], dim=0
```
remove the need to cast the `repeat_mask` to a `Shape` below?

pranavm-nvidia: Actually, now I'm wondering if it can be simplified to:
```py
repeat_mask = [1] * input.rank
repeat_mask[dim] = repeats
```
That seems more natural to write (and probably what a user would expect to be able to do).

tripy/tripy/frontend/trace/ops/binary_elementwise.py
```diff
@@ -50,23 +50,24 @@ def __str__(self):
...
-            # if there is a non-shape input, it must be rank 1 or 0 to avoid broadcasting
-            if not all(map(lambda t: t.rank <= 1, inputs)):
-                invalid_indices_message = ", ".join(
-                    map(
-                        lambda i: f"Index {i} (rank {inputs[i].rank})",
-                        filter(lambda i: inputs[i].rank > 1, range(len(inputs))),
-                    )
-                )
+            # Note: Most binary elementwise ops use the convert_inputs_to_tensors decorator, which also checks for this.
+            if any(map(lambda t: not isinstance(t, (Shape, ShapeScalar)), inputs)):
```
pranavm-nvidia, line 60: Why does this need to be checked here? Wouldn't the convert decorators already have checked it?

tripy/tripy/frontend/utils.py
```diff
@@ -245,26 +245,44 @@ def wrapper(*args, **kwargs):
...
-                key=lambda typ: {Tensor: 2, Shape: 1, ShapeScalar: 0}.get(typ, -1),
-            )
-            if issubclass(MaxType, Tensor):
-                TensorType = MaxType
+            if shape_scalar_encountered and not len(tensor_args) and not len(shape_args):
+                TensorType = ShapeScalar
+            if len(tensor_args):
+                TensorType = Tensor
+            if len(shape_args):
+                TensorType = Shape
```
pranavm-nvidia, line 285: I think this can be simplified to:
```py
types = {type(arg) for _, arg in all_args if isinstance(arg, Tensor)}
shape_scalar_encountered = ShapeScalar in types
types -= {ShapeScalar}

if len(types) > 1:
    raise_error(
        f"{func.__name__} expects tensor arguments to have matching class types, "
        f"but got mixed `tp.Tensor` and `tp.Shape` arguments",
        [
            "Consider explicitly converting using tp.Shape(tensor) or shape.as_tensor()\n"
            "Note: argument types were: " + ", ".join(f"{name}: {type(arg)}" for name, arg in all_args)
        ],
    )

TensorType = None
if types:
    TensorType = types.pop()
elif shape_scalar_encountered:
    TensorType = ShapeScalar
```

tripy/tripy/frontend/trace/ops/binary_elementwise.py
slyubomirsky: `min` and `max` don't use the decorator. It would otherwise be entirely redundant, yes.

tripy/tripy/frontend/utils.py
slyubomirsky: Very Pythonic! I will do that.

slyubomirsky:
Will reopen on the main repo to avoid the CI issue and will carry over suggestions.

PR #266, opened 2024-10-11
[Tripy] Rename `infer_shape_output_idxs` to `infer_tensor_variants`

slyubomirsky:
Partly addresses #233. `infer_shape_output_idxs` is no longer a very descriptive name since it can distinguish `Shape` outputs and `ShapeScalar` outputs. This PR renames it to `infer_tensor_variants`. Additionally, it uses an enum for the dict key to avoid having magic string constants throughout the codebase.

Happy to consider other names as well.

### DIFF
--- a/tripy/tripy/frontend/trace/ops/utils.py
+++ b/tripy/tripy/frontend/trace/ops/utils.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 #
 
+from enum import Enum
 from typing import Any, List, Optional, Sequence
 
 from tripy import utils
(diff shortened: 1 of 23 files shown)


tripy/tripy/frontend/trace/ops/utils.py
```diff
@@ -49,24 +50,34 @@ def get_broadcast_dim(dim1, dim2):
...
         """
-        Common override for `infer_shape_output_idxs`: Treat the outputs as shapes if the *first* input is a shape.
+        Treat the outputs as shapes if the *first* input is a shape.
         """
         from tripy.frontend.shape import Shape
 
         if isinstance(inputs[0], Shape):
-            return Result.ok({"shape": list(range(len(self.outputs)))})
+            return Result.ok({TensorVariants.SHAPE: list(range(len(self.outputs)))})
         return Result.ok({})
```
pranavm-nvidia, line 76: Why don't we return a `List[TensorVariants]` that has the same length as the outputs?
To me it feels a bit more natural to write:
```py
for out, tensor_variant in zip(outputs, tensor_variants):
    if tensor_variant == TensorVariant.SHAPE:
        ...
    elif tensor_variant == TensorVariant.SCALAR:
        ...
```
compared to the dictionary we have now.

tripy/tripy/frontend/trace/ops/utils.py
```diff
@@ -49,24 +50,34 @@ def get_broadcast_dim(dim1, dim2):
...
 
+"""
+Enum representing the tensor variants encoded in the result of `infer_tensor_variants`.
+"""
 
-class ShapeOutputIdxPolicies:
+
+class TensorVariants(Enum):
+    SHAPE = 1
+    SCALAR = 2
```
pranavm-nvidia, line 63: We can actually just use the `Shape` and `ShapeScalar` types directly instead of this enum. With that and the other change I suggested, the call-site can be simplified to:
```py
# If there are any shape outputs, we infer the length. 
if any(typ == Shape for typ in res.value):
    inferred_lengths = op.infer_len()

for idx, typ in enumerate(res.value):
    if typ is not None:
        outputs[idx] = typ(outputs[idx])
        if inferred_lengths[idx] is not None:
            out_trace_tensors[idx].shape = [inferred_lengths[idx]]
```

tripy/tripy/frontend/trace/ops/utils.py
slyubomirsky: Good idea.

tripy/tripy/frontend/trace/ops/utils.py
slyubomirsky: Not introducing the enum is a good idea too.

slyubomirsky:
Will reopen on the main repo to avoid the CI issue and will carry over suggestions.

PR #270, opened 2024-10-15
Overhaul backend function execution for improved performance and flexibility

jhalakpatel:
This PR replaces the DPS-style calling convention with a non-DPS approach, eliminating the requirement for call sites to preallocate output buffers. This change enables us to bypass the computation of output shapes and advance allocation of output buffers, laying the groundwork for supporting data-dependent shapes where network outputs can have dynamic dimensions.

The underlying compiler stack has been enhanced to avoid allocating oversized buffers and eliminate an extra device-to-device copy operation from TensorRT-allocated memory to MLIR-TRT managed memory.

Additionally, we've improved the copy operation to support copying to host memory. This enhancement removes the need to track output device allocations for device-to-host copies. Previously, copy outputs were restricted to device allocations; now they can be allocated on both device and host.

Tests have been updated to align with the new calling convention, ensuring compatibility and correctness.

### DIFF
--- a/tripy/tests/integration/test_quantize.py
+++ b/tripy/tests/integration/test_quantize.py
@@ -117,5 +117,6 @@ class TestQuantize:
         input = tp.ones((4, 4))
         scale = tp.ones((4,))
         quantized = tp.quantize(input, scale, tp.int8, dim=0)
+        quantized_int32 = tp.cast(quantized, tp.int32)
 
-        assert bool(tp.all(quantized == tp.ones((4, 4), dtype=tp.int8)))
+        assert bool(tp.all(quantized_int32 == tp.ones((4, 4), dtype=tp.int32)))
--- a/tripy/tripy/backend/api/executable.py
+++ b/tripy/tripy/backend/api/executable.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 import base64
 import inspect
-from typing import Sequence, Union
+from typing import Sequence, Union, Tuple, Callable
 
 import mlir_tensorrt.runtime.api as runtime
 
@@ -37,13 +37,11 @@ class Executable:
     """
 
     # The constructor is intentionally undocumented because it is not meant to be called by users.
-    # TODO(#155): output_devices is not needed after they can be queried from executable
-    def __init__(self, executable, arg_names, output_devices):
+    def __init__(self, executable, arg_names):
         self._executable = executable
         self._executor = Executor(self._executable)
         self._arg_names = arg_names
         self._num_expected_args = len(arg_names)
-        self._output_devices = output_devices
         self._executable_signature = self._executable.get_signature("main")
 
         # Build a signature so the executable works with `inspect.signature`
@@ -51,7 +49,7 @@ class Executable:
         for name in self._arg_names:
             params.append(inspect.Parameter(name, inspect.Parameter.POSITIONAL_OR_KEYWORD, annotation=Tensor))
 
-        return_annotation = Tensor if self._executable_signature.get_num_output_args() == 1 else Sequence[Tensor]
+        return_annotation = Tensor if self._executable_signature.get_num_results() == 1 else Sequence[Tensor]
 
         self.__signature__ = inspect.Signature(params, return_annotation=return_annotation)
 
@@ -128,7 +126,7 @@ class Executable:
             tensor.eval()
 
         try:
-            executor_outputs = self._executor.execute(self._output_devices, input_tensors)
+            executor_outputs = self._executor.execute(input_tensors)
         except runtime.MTRTException as err:
             # TODO: Evaluate whether this should be moved into the executor
             if "function expects a memref type with element type" in str(err):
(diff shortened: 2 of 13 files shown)


tripy/tripy/backend/mlir/executor.py
```diff
@@ -129,49 +56,10 @@ def execute(self, output_devices: List[device], inputs: List["Tensor"] = []) ->
...
-        # TODO(#155): MLIR-TensorRT should allow output tensor placements on host.
-        for idx, out_info in enumerate(out_tensor_info):
-            if out_info.device.kind != "gpu":
-                self.runtime_client.copy_to_host(
-                    device_memref=out_args[idx],
-                    existing_host_memref=outputs[idx],
-                    stream=self.stream._active_cuda_stream,
-                )
-
+        # For now return results on GPU.
```
jhalakpatel, line 64: Let's see if we can get rid of this copy operation.

tripy/tripy/frontend/tensor.py
```diff
@@ -178,11 +178,11 @@ def eval(self) -> runtime.MemRefValue:
 
         compiler = Compiler(trt_builder_opt_level=0)
         executable = compiler.compile(mlir, flat_ir=flat_ir)
-        executor = Executor(executable)
+        self.executor = Executor(executable)
```
jhalakpatel, line 181: @pranavm-nvidia @parthchadha Allocated output pointer is owned by the `RuntimeClient` (`Executor` owns `RuntimeClient`), we need to ensure bind the lifetime of the runtime client to end-to-end Tripy execution.

This approach does not work if later try to access the memref pointer:
```
        input = tp.arange(1, 4, dtype=tp.float32)
        print(cp.from_dlpack(input))
```

tripy/tests/integration/test_quantize.py
```diff
@@ -117,5 +117,6 @@ def test_non_constant_scale(self):
         input = tp.ones((4, 4))
         scale = tp.ones((4,))
         quantized = tp.quantize(input, scale, tp.int8, dim=0)
+        quantized_int32 = tp.cast(quantized, tp.int32)
 
-        assert bool(tp.all(quantized == tp.ones((4, 4), dtype=tp.int8)))
+        assert bool(tp.all(quantized_int32 == tp.ones((4, 4), dtype=tp.int32)))
```
jhalakpatel, line 122: This is required because MLIR-TRT now does not support comparing i8 values. https://github.com/NVIDIA/TensorRT-Incubator/blob/main/mlir-tensorrt/tensorrt/lib/TensorRT/IR/TensorRTVersionCompatibility.cpp#L152

tripy/tripy/backend/api/executable.py
```diff
@@ -221,11 +226,16 @@ def add(a, b):
...
+        num_input_args = self._executable_signature.get_num_input_args()
+        num_output_args = self._executable_signature.get_num_output_args()
+        num_results = self._executable_signature.get_num_results()
+
+        assert not (num_output_args and num_results), "Cannot have both output arguments and results"
+
+        if num_output_args:
+            return [self._get_arg_info(idx + num_input_args) for idx in range(num_output_args)]
+        else:
+            return [self._get_result_info(idx) for idx in range(num_results)]
```
pranavm-nvidia, line 238: Why don't we just return an empty list when there are no outputs?

tripy/tripy/frontend/tensor.py
```diff
@@ -188,7 +188,7 @@ def eval(self) -> runtime.MemRefValue:
         executor = Executor(executable)
         # Upon computing the value of this tensor, we switch it to have a `Storage`
         # parameter so that it does not need to be computed again.
-        data = executor.execute([out.device for out in flat_ir.outputs])
+        data = executor.execute()
         executor.stream.synchronize()
         assert len(data) == 1, "Expects only one output from mlir_tensorrt.compiler executor"
         data = data[0]
```
pranavm-nvidia, line 194: Could we also remove the hack a few lines below?
```py
self.trace_tensor.device = flat_ir.outputs[0].device
```

tripy/tripy/backend/api/executable.py
pranavm-nvidia: I guess what is the distinction between output vs. result?

jhalakpatel: For Tripy, we are only returning results, so we could get rid of output arguments. MLIR-TRT supports function which both accept output arguments and return results.
```
func add(int a, int b, int* r) {
    *r = a + b;
    return *r;
}
```

jhalakpatel: `num_output_args` check is preserving old path. I will get rid of it.

tripy/tripy/frontend/tensor.py
jhalakpatel: There are still a few issues:
1. MLIR-TensorRT still allocates inputs only on the device.
2. This PR fixes the copy operation only.


There are still issues since the input get always allocated on device. There will be a device mismatch between inputs (which are always on device) vs output which could be now on (host as well device).

```
    def infer_devices(self):
        """
        Infers devices for the operation and updates output tensor devices accordingly.
        """
        assert (
>           self.inputs and len(self.outputs) == 1 and all(inp.device == self.inputs[0].device for inp in self.inputs)
        ), "Default implementation cannot handle cases where there are no inputs, multiple outputs, or multiple inputs with different devices. Please override."
E       AssertionError: Default implementation cannot handle cases where there are no inputs, multiple outputs, or multiple inputs with different devices. Please override.
```

tripy/tests/integration/test_quantize.py
pranavm-nvidia: Is that a bug? This was working before

tripy/tripy/frontend/trace/ops/reduce.py
```diff
@@ -413,7 +413,7 @@ def _arg_min_max_impl(tensor: "tripy.Tensor", kind: ArgMinMax.Kind, dim: Optiona
 @export.public_api(document_under="operations/functions")
 @constraints.dtypes(
     constraints={"input": "T1", constraints.RETURN_VALUE: "T2"},
-    variables={"T1": ["float32", "float16", "bfloat16", "int32", "bool", "int8"], "T2": ["int32"]},
```
pranavm-nvidia, line 416: Why was support dropped for these types?

jhalakpatel: @shelkesagar29 Do you know why that is the case?

tripy/tripy/frontend/tensor.py
```diff
@@ -186,10 +186,11 @@ def eval(self) -> runtime.MemRefValue:
 
         compiler = Compiler(trt_builder_opt_level=0)
         executable = compiler.compile(mlir, flat_ir=flat_ir)
+        # Ensure that session and client are available as long as tensor lives.
```
jhalakpatel, line 189: Remove this comment.

jhalakpatel:
Closing as opened a new PR: https://github.com/NVIDIA/TensorRT-Incubator/pull/489. I lost permissions to directly commit to the project so I am moving to fork based development for now.

PR #277, opened 2024-10-16
[Tripy] Disallow passing both Shapes and Tensors with the `convert_inputs_to_tensors decorator`

slyubomirsky:
Addresses issue https://github.com/NVIDIA/TensorRT-Incubator/issues/233, reopened from #265 and incorporating suggestions (thanks, @pranavm-nvidia). Per discussions, there are not many remaining uses of the `convert_inputs_to_tensors` decorator in the codebase and it also seems that in just about all of those cases, it would not make sense to accept both shapes and tensors.

### DIFF
--- a/tripy/tripy/frontend/utils.py
+++ b/tripy/tripy/frontend/utils.py
@@ -248,41 +248,36 @@ def convert_inputs_to_tensors(
             # Disallow mixing Tensor and Shape by default. If it makes sense in a given function
             # to have both Tensor and Shape arguments, that might suggest that custom handling
             # rather than relying on this decorator would make sense.
-            tensor_args = []
-            shape_args = []
-            # Don't need special handling for ShapeScalars because they get broadcast up into larger sizes
-            # as needed -- can treat them as a tensor or a shape whenever we need
-            shape_scalar_encountered = False
-            for arg_name, arg in all_args:
-                if arg_name in exclude:
-                    continue
-                if isinstance(arg, Shape):
-                    shape_args.append((arg_name, arg))
-                elif isinstance(arg, ShapeScalar):
-                    shape_scalar_encountered = True
-                elif isinstance(arg, Tensor):
-                    tensor_args.append((arg_name, arg))
+            types = set(
+                {
+                    # There are other subclasses of Tensor, like Parameter and DefaultParameter.
+                    # Unless otherwise specified, we treat them as ordinary Tensors.
+                    Tensor if type(arg) != Shape and type(arg) != ShapeScalar else type(arg)
+                    for arg_name, arg in all_args
+                    if isinstance(arg, Tensor) and arg_name not in exclude
+                }
+            )
+            # We usually can treat ShapeScalars as either tensors or shapes due to broadcasting, so we can remove them from the below check.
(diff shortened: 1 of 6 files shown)


tripy/tripy/frontend/utils.py
```diff
@@ -245,26 +245,39 @@ def wrapper(*args, **kwargs):
...
-                (type(arg) for arg_name, arg in all_args if arg_name not in exclude),
-                key=lambda typ: {Tensor: 2, Shape: 1, ShapeScalar: 0}.get(typ, -1),
+            # Disallow mixing Tensor and Shape by default. If it makes sense in a given function
+            # to have both Tensor and Shape arguments, that might suggest that custom handling
+            # rather than relying on this decorator would make sense.
+            types = set(
+                {
+                    # There are other subclasses of Tensor, like Parameter and DefaultParameter.
+                    # Unless otherwise specified, we treat them as ordinary Tensors.
+                    Tensor if type(arg) != Shape and type(arg) != ShapeScalar else type(arg)
```
pranavm-nvidia, line 255:
```suggestion
                    Tensor if type(arg) not in {Shape, ShapeScalar} else type(arg)
```

tripy/tripy/frontend/utils.py
```diff
@@ -245,26 +245,39 @@ def wrapper(*args, **kwargs):
...
-            # That is, if we have an operation with a mixture of types, all arguments are
-            # converted to the type with the highest precedence, e.g. Tensor + Shape -> Tensor.
-            TensorType = None
-            MaxType = max(
-                (type(arg) for arg_name, arg in all_args if arg_name not in exclude),
-                key=lambda typ: {Tensor: 2, Shape: 1, ShapeScalar: 0}.get(typ, -1),
+            # Disallow mixing Tensor and Shape by default. If it makes sense in a given function
+            # to have both Tensor and Shape arguments, that might suggest that custom handling
+            # rather than relying on this decorator would make sense.
+            types = set(
```
pranavm-nvidia, line 251: The `set()` here is not needed.

tripy/tripy/frontend/utils.py
```diff
@@ -245,26 +245,39 @@ def wrapper(*args, **kwargs):
...
+                    for arg_name, arg in all_args
+                    if isinstance(arg, Tensor) and arg_name not in exclude
+                }
             )
-            if issubclass(MaxType, Tensor):
-                TensorType = MaxType
+            # We usually can treat ShapeScalars as either tensors or shapes due to broadcasting, so we can remove them from the below check.
+            shape_scalar_encountered = ShapeScalar in types
+            if ShapeScalar in types:
+                types.remove(ShapeScalar)
```
pranavm-nvidia, line 263: This is equivalent to:
```suggestion
            types -= {ShapeScalar}
```

slyubomirsky: Good to know, I forgot to wrap the RHS in a set and was wondering why it didn't work.

tripy/tripy/frontend/utils.py
slyubomirsky: Removed, thanks.

pranavm-nvidia (approved):

PR #282, opened 2024-10-17
[Lua/TensorRTModule] Implement OutputAllocator for TensorRT execution

jhalakpatel:
This commit adds the OutputAllocator and OutputAllocatorImpl classes to manage memory allocation for TensorRT outputs. It provides methods for setting and retrieving tensor data, shape, and allocation information. Implemented `OutputDescriptor` class to aid in extracting rank, result, ptr etc. from memory buffer refernced by output descriptor pointer. Also, added a lua method to ensure allocators lifetime is coupled with the session.

### DIFF
--- a/mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
+++ b/mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
@@ -110,6 +110,170 @@ public:
   std::shared_ptr<nvinfer1::ICudaEngine> engine;
 };
 
+/// OutputAllocatorImpl - Implementation of TensorRT's IOutputAllocator
+/// interface. This class manages memory allocation for TensorRT outputs.
+class OutputAllocatorImpl : public nvinfer1::IOutputAllocator {
+public:
+  OutputAllocatorImpl(AllocTracker *tracker) : mTracker(tracker) {}
+  /// It is the responsibilty of runtime client to release the buffer.
+  ~OutputAllocatorImpl() {}
+
+  // Disable copy and move operations
+  OutputAllocatorImpl(const OutputAllocatorImpl &) = delete;
+  OutputAllocatorImpl &operator=(const OutputAllocatorImpl &) = delete;
+  OutputAllocatorImpl(OutputAllocatorImpl &&) = delete;
+  OutputAllocatorImpl &operator=(OutputAllocatorImpl &&) = delete;
+
+  void setTensorName(const std::string &name) { mTensorName = name; }
+  void setCurrentMemory(uintptr_t memory, int64_t size) {
+    mOutputPtr = memory;
+    mOutputSize = size;
+  }
+
+  void *reallocateOutputAsync(const char *name, void *memory, uint64_t size,
+                              uint64_t alignment,
+                              cudaStream_t stream) override {
+    assert((!mOutputPtr || reinterpret_cast<uintptr_t>(memory) == mOutputPtr) &&
+           "Output buffer mismatch");
+    assert(name == mTensorName && "Tensor name mismatch");
+    size = std::max(size, static_cast<uint64_t>(1));
+    if (size > mOutputSize) {
+      size = roundUp(size, alignment);
+      if (mOutputPtr)
+        mlirtrt::runtime::safeDeallocate(*mTracker, mOutputPtr,
+                                         CudaStreamPtr(stream));
+      mOutputPtr = 0;
+      mOutputSize = 0;
+      StatusOr<PointerInfo> memory =
+          mlirtrt::runtime::allocate(*mTracker, PointerType::device, size,
+                                     alignment, CudaStreamPtr(stream));
+      if (memory.isOk()) {
+        mOutputPtr = (*memory).ptr;
+        mOutputSize = memory->size;
+      }
+      return reinterpret_cast<void *>(mOutputPtr);
+    }
+    return reinterpret_cast<void *>(mOutputPtr);
+  }
+
+  void notifyShape(const char *name, const nvinfer1::Dims &dims) override {
+    assert(name == mTensorName && "Tensor name mismatch");
+    mOutputDims = dims;
+  }
+
+  // Accessor methods
+  const nvinfer1::Dims &getOutputDims() const { return mOutputDims; }
+  uintptr_t getOutputPtr() const { return mOutputPtr; }
+  uint64_t getOutputSize() const { return mOutputSize; }
+  const std::string &getTensorName() const { return mTensorName; }
+
+private:
+  static inline uint64_t roundUp(uint64_t m, uint64_t n) {
+    return llvm::divideCeil(m, n) * n;
+  }
+
+  uintptr_t mOutputPtr;
+  uint64_t mOutputSize{0};
+  nvinfer1::Dims mOutputDims;
+  std::string mTensorName;
+  AllocTracker *mTracker{nullptr};
+};
+
+/// OutputAllocator - Manages multiple OutputAllocatorImpl instances.
+class OutputAllocator {
+public:
+  explicit OutputAllocator(AllocTracker *tracker, int64_t nbResults) {
+    mAllocators.reserve(nbResults);
+    for (int64_t i = 0; i < nbResults; ++i)
+      mAllocators.push_back(std::make_unique<OutputAllocatorImpl>(tracker));
+  }
+
+  // Disable copy and move operations
+  OutputAllocator(const OutputAllocator &) = delete;
+  OutputAllocator &operator=(const OutputAllocator &) = delete;
+  OutputAllocator(OutputAllocator &&) = delete;
+  OutputAllocator &operator=(OutputAllocator &&) = delete;
+
+  void registerAllocator(size_t index, const char *name, uintptr_t ptr,
+                         int64_t size, nvinfer1::IExecutionContext *context) {
+    assert(index >= 0 && index < mAllocators.size() && "Index out of bounds");
+    mAllocators[index]->setTensorName(name);
+    mAllocators[index]->setCurrentMemory(ptr, size);
+    context->setOutputAllocator(name, mAllocators[index].get());
+  }
+
+  OutputAllocatorImpl *getAllocator(size_t index) {
+    assert(index >= 0 && index < mAllocators.size() && "Index out of bounds");
+    return mAllocators[index].get();
+  }
+
+private:
+  std::vector<std::unique_ptr<OutputAllocatorImpl>> mAllocators;
+};
+
+/// Manages output tensor descriptors for TensorRT execution.
+class OutputDescriptor {
+public:
+  OutputDescriptor(uintptr_t ptr)
+      : mData(reinterpret_cast<int64_t *>(ptr)),
+        mSize(calculateTotalSize(ptr)) {}
+
+  int64_t getNumberOfResults() const { return mData[0]; }
+
+  unsigned getRank(int resultIndex) const {
+    size_t index = getIndexForResult(resultIndex);
+    return static_cast<unsigned>(mData[index]);
+  }
+
+  void setTensorDataPtr(int resultIndex, uintptr_t ptr) {
+    size_t index = getIndexForResult(resultIndex);
+    mData[index + 1] = static_cast<int64_t>(ptr);
+  }
+
+  void setShape(int resultIndex, const std::vector<int64_t> &shape) {
+    size_t index = getIndexForResult(resultIndex);
+    unsigned rank = static_cast<unsigned>(mData[index]);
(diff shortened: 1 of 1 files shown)


christopherbate:
failing clang-format step and there is a merge conflict

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -198,6 +198,195 @@ class NvInferExecContextWrapper {
   /// the TRT network.
   std::vector<PinnedMemoryBlock> hostIOBuffers;
 };
+
+/// OutputAllocatorImpl - Implementation of TensorRT's IOutputAllocator
+/// interface. This class manages memory allocation for TensorRT outputs.
+class OutputAllocatorImpl : public nvinfer1::IOutputAllocator {
+public:
+  OutputAllocatorImpl() = default;
+  /// It is the responsibilty of runtime client to release the buffer.
```
christopherbate, line 207: It's the responsibility of the runtime client to release the buffer if the buffer is not cleaned up or returned from the entrypoint function. If the buffer is internal to the module (e.g. in the case of multiple TRT segments where one TRT 
segment's results never "escape" the function) then the compiler should insert explicit deallocations.

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -198,6 +198,195 @@ class NvInferExecContextWrapper {
...
+
+      if (mOutputPtr) {
+        cudaFree(mOutputPtr);
+      }
+
+      mOutputPtr = nullptr;
+      mOutputSize = 0;
+
+      void *newMemory;
+      if (cudaMalloc(&newMemory, size) == cudaSuccess) {
```
christopherbate, line 246: Any allocation needs to be  made via the `AllocTracker`, otherwise the runtime client won't have visibility

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -198,6 +198,195 @@ class NvInferExecContextWrapper {
...
+    assert(!mReallocateOutputCalled && "Duplicate call to reallocateOutput");
+    mReallocateOutputCalled = true;
+
+    size = std::max(size, static_cast<uint64_t>(1));
+
+    if (size > mOutputSize) {
+      size = roundUp(size, alignment);
+
+      if (mOutputPtr) {
+        cudaFree(mOutputPtr);
```
christopherbate, line 239: free needs to go through AllocTracker

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -198,6 +198,195 @@ class NvInferExecContextWrapper {
   /// the TRT network.
   std::vector<PinnedMemoryBlock> hostIOBuffers;
 };
+
+/// OutputAllocatorImpl - Implementation of TensorRT's IOutputAllocator
+/// interface. This class manages memory allocation for TensorRT outputs.
+class OutputAllocatorImpl : public nvinfer1::IOutputAllocator {
+public:
+  OutputAllocatorImpl() = default;
```
christopherbate, line 206: You'll need to pass `AllocTracker` to the object

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -198,6 +198,195 @@ class NvInferExecContextWrapper {
...
+public:
+  explicit OutputAllocator(int64_t nbResults);
+
+  // Disable copy and move operations
+  OutputAllocator(const OutputAllocator &) = delete;
+  OutputAllocator &operator=(const OutputAllocator &) = delete;
+  OutputAllocator(OutputAllocator &&) = delete;
+  OutputAllocator &operator=(OutputAllocator &&) = delete;
+
+  void registerAllocator(size_t index, const char *name, void *ptr,
```
christopherbate, line 299: What is this method for? It doesn't seem to be used?

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -198,6 +198,195 @@ class NvInferExecContextWrapper {
...
+  const nvinfer1::Dims &getOutputDims() const { return mOutputDims; }
+  bool isReallocateOutputCalled() const { return mReallocateOutputCalled; }
+  bool isNotifyShapeCalled() const { return mNotifyShapeCalled; }
+  void *getOutputPtr() const { return mOutputPtr; }
+  uint64_t getOutputSize() const { return mOutputSize; }
+  const std::string &getTensorName() const { return mTensorName; }
+
+private:
+  /// Round up the given value to the nearest multiple of n.
+  static inline uint64_t roundUp(uint64_t m, uint64_t n) {
```
christopherbate, line 275: Use `llvm::divideCeil`

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -198,6 +198,195 @@ class NvInferExecContextWrapper {
...
+           "TensorRT must invoke reallocateOutput first");
+    assert(!mNotifyShapeCalled && "Duplicate call to notifyShape");
+    assert(tensorName == mTensorName && "Tensor name mismatch");
+    mNotifyShapeCalled = true;
+    mOutputDims = dims;
+  }
+
+  // Accessor methods
+  const nvinfer1::Dims &getOutputDims() const { return mOutputDims; }
+  bool isReallocateOutputCalled() const { return mReallocateOutputCalled; }
```
christopherbate, line 267: Please remove various unused methods. The `mReallocateOutputCalled` seems like a debugging device; there shouldn't be any need to expose it via class' public API?

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -429,4 +656,25 @@ void mlirtrt::runtime::registerExecutorTensorRTModuleLuaRuntimeMethods(
...
+          sol::this_state state,
+          std::shared_ptr<NvInferExecContextWrapper> context,
+          CudaStreamPtr stream, uintptr_t outputDesc, sol::table va) {
+        ADD_TENSORRT_MODULE_RANGE("trtrt_enqueue_alloc");
+        sol::state_view luaState(state);
+        assert(context != nullptr);
+        assert(stream != nullptr && "expected valid stream");
+
+        OutputDescriptor desc(outputDesc);
+        auto &allocator = luaState["_trtrt_create_allocator"]
```
christopherbate, line 671: This whole indirection isn't necessary since this allocator doesn't need to outlive this function (I guess this is just leftover from when the output allocator thing was tied to the pointer lifetime)?

So we can just do `auto allocator = std::make_unique<OutputAllocator>(nbResults);` here?

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -198,6 +198,195 @@ class NvInferExecContextWrapper {
...
+    size_t index = getIndexForResult(resultIndex);
+    return static_cast<unsigned>(mData[index]);
+  }
+
+  void setTensorDataPtr(int resultIndex, uintptr_t ptr) {
+    size_t index = getIndexForResult(resultIndex);
+    mData[index + 1] = static_cast<int64_t>(ptr);
+  }
+
+  void setShape(int resultIndex, const std::vector<int64_t> &shape) {
```
christopherbate, line 346: This is not called?

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -198,6 +198,195 @@ class NvInferExecContextWrapper {
...
+    size_t index = getIndexForResult(resultIndex);
+    unsigned rank = static_cast<unsigned>(mData[index]);
+    assert(shape.size() == rank && "Shape size doesn't match the rank");
+    index += OUTPUT_DESC_FIXED_FIELDS;
+    std::copy(shape.begin(), shape.end(), mData + index);
+  }
+
+  void setStride(int resultIndex, const std::vector<int64_t> &stride) {
```
christopherbate, line 354: This is not called?

christopherbate (changes requested):

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
jhalakpatel: Yes, those methods were for debugging purposes. I will remove them from public API implementation.

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
jhalakpatel: This and all the below methods are used. There was a bad merge. I will update the PR with those changes.

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
jhalakpatel: Replied above.

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
jhalakpatel: Replied above.

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
jhalakpatel: IIUC, in case we have mutliple tensorrt segments, are you suggesting that only the last tensorrt segment would cross the final "main" function boundary and hence the rest of the output pointers belonging to rest of the tensorrt segments needs to be released by the compiler?

How do you suggest we implement this? We can simply just insert `safeDeallocate` in the `OutputAllocatorImpl` destructor.

christopherbate: It's not something you have to worry about. It's already taken care of by the system. You just need to be sure to use `AllocTracker` for all memory alloc/dealloc.

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -110,6 +110,137 @@ class NvInferEngineWrapper {
...
+                         int64_t size, nvinfer1::IExecutionContext *context);
+  OutputAllocatorImpl *getAllocator(size_t index);
+
+private:
+  std::vector<std::unique_ptr<OutputAllocatorImpl>> mAllocators;
+};
+
+OutputAllocator::OutputAllocator(AllocTracker *tracker, int64_t nbResults) {
+  mAllocators.reserve(nbResults);
+  for (int64_t i = 0; i < nbResults; ++i) {
```
christopherbate, line 225: Nit: remove braces 

https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements

christopherbate (approved):
Minor style comment otherwise lgtm

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
jhalakpatel: Make sense. Yes, runtime client should be able to call `safeDeallocate` on the pointer AFAICT.

PR #284, opened 2024-10-17
[Compiler] Split compiler pipeline into pass pipeline, compilation, and translation steps

jhalakpatel:


christopherbate:
The API here seems much too adhoc. It's adding a dedicated option just to control something printed to ``stderr`.  There's already a mechanism for that via the debug options `-debug`, `-debug-only`, etc.

I think a better path would be some of the following options:

1. add an API to get the textual pipeline as a string. This could would be something like 

```
client.get_textual_pipeline(opts)
``` 

2. An alternative option would leverage MLIR's existing bindings to PassPipeline, which is already capable of printing out the textual pipeline. This opens more opportunities for TriPy to directly manipulate the pass pipeline. The API would change so that instead of directly invoking `stablehlo_to_executable`, you instead first get the pass pipeline object, run compilation, and then invoke translation, something like:

```
pm = client.get_stablehlo_to_executable_pipeline(opts)
print(pm)
lowered_module = pm.run(module)
exe = client.translate_to_runtime_executable(module)
```

Here `exe` is the equivalent of what would be returned from `client.stablehlo_to_executable` now.

christopherbate (changes requested):
See comments above

jhalakpatel:
> The API here seems much too adhoc. It's adding a dedicated option just to control something printed to `stderr`. There's already a mechanism for that via the debug options `-debug`, `-debug-only`, etc.
> 
> I think a better path would be some of the following options:
> 
> 1. add an API to get the textual pipeline as a string. This could would be something like
> 
> ```
> ir = client.get_textual_pipeline(opts)
> ```
> 
> Then you can print that or write to file as desired. That would also enable us to control how the pipeline is printed in the backend. As you know, several passes in the textual pipeline dump from a PassManager are not actually registered. We would instead inspect the `opts` and then print out a custom, higher-level, pipeline sequence that can actually be run with the `mlir-tensorrt-opt` tool.
> 
> 2. An alternative option would leverage MLIR's existing bindings to PassPipeline, which is already capable of printing out the textual pipeline. This opens more opportunities for TriPy to directly manipulate the pass pipeline. The API would change so that instead of directly invoking `stablehlo_to_executable`, you instead first get the pass pipeline object, run compilation, and then invoke translation, something like:
> 
> ```
> pm = client.get_stablehlo_to_executable_pipeline(opts)
> print(pm)
> lowered_module = pm.run(module)
> exe = client.translate_to_runtime_executable(module)
> ```
> 
> Here `exe` is the equivalent of what would be returned from `client.stablehlo_to_executable` now.

I liked the approach (2) where we could control the pipeline passes from Tripy if required.
CC: @parthchadha @pranavm-nvidia

PR #285, opened 2024-10-18
[Dialect/Plan] Update Plan dialect to use non-dps calling convention

jhalakpatel:
The Plan dialect has been enhanced to support non-DPS calling conventions, with updates to PlanAllocTensorsPass and CreateClosedRegionsPass. Additionally, the Plan transformation now converts both non-DPS and DPS group operations to CallAllocOp and CallOp respectively, required for subsequent transformations.

### DIFF
--- a/mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/StableHloToExecutable.h
+++ b/mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/StableHloToExecutable.h
@@ -128,6 +128,9 @@ struct StableHLOToExecutableOptions : public mlir::OptionsContext {
   /// Whether to disallow host tensors in TensorRT clusters.
   bool disallowHostTensorsInTensorRTClusters = false;
 
+  /// Whether to use non-DPS style calling convention.
+  bool useNonDPSCallConv = false;
+
   /// Entrypoint function name.
   std::string entrypoint = "main";
 
--- a/mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/StableHloToExecutable.h
+++ b/mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/StableHloToExecutable.h
@@ -128,8 +128,9 @@ struct StableHLOToExecutableOptions : public mlir::OptionsContext {
   /// Whether to disallow host tensors in TensorRT clusters.
   bool disallowHostTensorsInTensorRTClusters = false;
 
-  /// Whether to use non-DPS style calling convention.
-  bool useNonDPSCallConv = false;
+  /// Use non-DPS style calling convention for entrypoint function
+  /// and backend types that support allocating results.
+  bool enableNonDPSReturns = false;
 
   /// Entrypoint function name.
   std::string entrypoint = "main";
--- a/mlir-tensorrt/compiler/include/mlir-tensorrt/Dialect/Plan/IR/PlanOps.td
+++ b/mlir-tensorrt/compiler/include/mlir-tensorrt/Dialect/Plan/IR/PlanOps.td
@@ -277,10 +277,10 @@ def Plan_InlineClosedGroupOp : Plan_InlineClosedGroupBase<"inline_closed_group",
 }
 
 //===----------------------------------------------------------------------===//
-// InlineClosedGroupNonDPSOp
+// InlineClosedAllocGroupOp
 //===----------------------------------------------------------------------===//
 
-def Plan_InlineClosedGroupNonDPSOp : Plan_InlineClosedGroupBase<"inline_closed_group_non_dps", [
+def Plan_InlineClosedAllocGroupOp : Plan_InlineClosedGroupBase<"inline_closed_alloc_group", [
   IsolatedFromAbove,
   SingleBlockImplicitTerminator<"plan::YieldOp">,
   DeclareOpInterfaceMethods<RegionBranchOpInterface,
@@ -289,10 +289,12 @@ def Plan_InlineClosedGroupNonDPSOp : Plan_InlineClosedGroupBase<"inline_closed_g
     ["getAsmBlockArgumentNames"]>
 ]> {
   let description = [{
-  The `plan.inline_closed_group_non_dps` operation is a variant of the
+  The `plan.inline_closed_alloc_group` operation is a variant of the
   `plan.inline_closed_group` operation that does not use destination-passing style
-  (DPS). It is isolated from above and explicitly captures input operands,
-  but unlike its DPS counterpart, it does not capture destination operands.
+  (DPS). It is isolated from above and explicitly captures input operands, but unlike
+  its DPS counterpart, it does not capture destination operands because its results must
+  be lowered to allocation(s). The allocations may or may not be of a size that can only 
+  be computed inside of the region.
   This operation takes input operands and their corresponding bounds attributes, 
   and produces results. The `input_attrs` hold bounds attribute information for 
   the input operands. The absence of bounds information is allowed (`none` bounds).
@@ -307,9 +309,9 @@ def Plan_InlineClosedGroupNonDPSOp : Plan_InlineClosedGroupBase<"inline_closed_g
   %0 = ... : tensor<?xf32> // A dynamically shaped operand
   %1 = ... : index         // A dynamic calculation of %0's extent
 
-  %2 = plan.inline_closed_group_non_dps target(#plan.cluster_target<tensorrt>)
+  %2 = plan.inline_closed_alloc_group target(#plan.cluster_target<tensorrt>)
     inputs(%0, %1 : tensor<?xf32>, index)
-    in_attrs [#plan.bounds<shape, , >, #plan.bounds<none>] -> tensor<?xf32> {
+    in_attrs [#plan.bounds<shape, [10], [20]>, #plan.bounds<none>]-> tensor<?xf32> {
     %3 = plan.with_shape %0 (%1) : (tensor<?xf32>, index) -> tensor<?xf32>
     %4 = stablehlo.exponential %3 : tensor<?xf32>
     yield %4 : tensor<?xf32>
@@ -353,7 +355,7 @@ def Plan_YieldOp : Plan_Op<"yield", [
       Terminator,
       ReturnLike,
       ParentOneOf<["plan::InlineGroupOp",
-                   "plan::InlineClosedGroupOp", "plan::InlineClosedGroupNonDPSOp"]>]> {
+                   "plan::InlineClosedGroupOp", "plan::InlineClosedAllocGroupOp"]>]> {
 
   let arguments = (ins Variadic<AnyType>:$results);
 
--- a/mlir-tensorrt/compiler/include/mlir-tensorrt/Dialect/Plan/IR/PlanOps.td
+++ b/mlir-tensorrt/compiler/include/mlir-tensorrt/Dialect/Plan/IR/PlanOps.td
@@ -131,11 +131,41 @@ def Plan_InlineGroupOp : Plan_GroupOpBase<"inline_group", [
 }
 
 //===----------------------------------------------------------------------===//
-// InlineClosedGroupOp
+// Plan_InlineClosedGroupBase
 //===----------------------------------------------------------------------===//
 
-def Plan_InlineClosedGroupOp : Plan_GroupOpBase<"inline_closed_group", [
-  IsolatedFromAbove,
+class Plan_InlineClosedGroupBase<string mnemonic, list<Trait> traits = []> :
+    Plan_GroupOpBase<mnemonic, traits # [IsolatedFromAbove]> {
+
+  code baseInlineClosedExtraClassDeclaration = baseExtraClassDeclaration # [{
+    // Common methods for both DPS and non-DPS versions
+    bool argHasTensorType(unsigned inputIdx) {
+      assert(inputIdx < getInputs().size() && "input index out-of-bounds");
+      return isa<RankedTensorType>(getInputs()[inputIdx].getType());
+    }
+
+    BoundsAttr getInputBoundsAttr(unsigned inputIdx) {
+      assert(inputIdx < getInputs().size() && "input index out-of-bounds");
+      return cast<BoundsAttr>(getInputAttrs()[inputIdx]);
+    }
+
+    /// Populate the `input_attrs` from an array of BoundsAttrs.
+    void setInputAttrsAttr(ArrayRef<BoundsAttr> boundsAttrs) {
+      setInputAttrsAttr(::mlir::ArrayAttr::get(
+        getOperation()->getContext(),
+        ArrayRef<Attribute>(boundsAttrs.begin(), boundsAttrs.end())
+      ));
+    }
+  }];
+
+  let extraClassDeclaration = baseInlineClosedExtraClassDeclaration;
+}
+
+//===----------------------------------------------------------------------===//
+// Plan_InlineClosedGroupOp
+//===----------------------------------------------------------------------===//
+
+def Plan_InlineClosedGroupOp : Plan_InlineClosedGroupBase<"inline_closed_group", [
   AttrSizedOperandSegments,
   DestinationStyleOpInterface,
   SingleBlockImplicitTerminator<"plan::YieldOp">,
@@ -226,22 +256,10 @@ def Plan_InlineClosedGroupOp : Plan_GroupOpBase<"inline_closed_group", [
                    CArg<"ArrayRef<BoundsAttr>", "{}">:$res_attrs)>
   ];
 
-  let extraClassDeclaration = baseExtraClassDeclaration # [{
+  let extraClassDeclaration = baseInlineClosedExtraClassDeclaration # [{
 
     MutableOperandRange getDpsInitsMutable() {
       return getOutsMutable();
-    }
-
-    /// Returns true if the `i-th` input argument has a tensor type.
-    bool argHasTensorType(unsigned inputIdx) {
-      assert(inputIdx < getInputs().size() && "input index out-of-bounds");
-      return isa<RankedTensorType>(getInputs()[inputIdx].getType());
-    }
-
-    /// Returns the i-th input argument's bounds attribute.
-    BoundsAttr getInputBoundsAttr(unsigned inputIdx) {
-      assert(inputIdx < getInputs().size() && "input index out-of-bounds");
-      return cast<BoundsAttr>(getInputAttrs()[inputIdx]);
     }
 
     ArrayRef<BlockArgument> getRegionOutArgs() {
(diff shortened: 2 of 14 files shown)


mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/StableHloToExecutable.h
```diff
@@ -128,6 +128,9 @@ struct StableHLOToExecutableOptions : public mlir::OptionsContext {
   /// Whether to disallow host tensors in TensorRT clusters.
   bool disallowHostTensorsInTensorRTClusters = false;
 
+  /// Whether to use non-DPS style calling convention.
```
christopherbate, line 131: This should say something along the lines of "use non-DPS calling convention for the entrypoint function and backend types that support allocating results"

mlir-tensorrt/compiler/include/mlir-tensorrt/Dialect/Plan/IR/PlanOps.td
```diff
@@ -255,16 +273,75 @@ def Plan_InlineClosedGroupOp : Plan_GroupOpBase<"inline_closed_group", [
...
-      setInputAttrsAttr(::mlir::ArrayAttr::get(
-        getOperation()->getContext(),
-        ArrayRef<Attribute>(boundsAttrs.begin(), boundsAttrs.end())
-      ));
-    }
+//===----------------------------------------------------------------------===//
+// InlineClosedGroupNonDPSOp
+//===----------------------------------------------------------------------===//
+
+def Plan_InlineClosedGroupNonDPSOp : Plan_InlineClosedGroupBase<"inline_closed_group_non_dps", [
```
christopherbate, line 283: Let's use a name similar to the pattern used in the TensorRT ops which you are adding, e.g. 

```suggestion
def Plan_InlineClosedAllocGroupOp : Plan_InlineClosedGroupBase<"inline_closed_alloc_group", [
```

mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/AllocTensors.cpp
```diff
@@ -856,11 +856,13 @@ class AllocTensorsPass
       }
     }
 
-    // First rewrite public functions to conform to DPS style.
     IRRewriter rewriter(ctx);
-    if (failed(rewriteNotPrivateFuncsToDPS(rewriter, op))) {
-      op->emitError("Failed to convert non-private functions to DPS");
-      return signalPassFailure();
+    if (!useNonDPSCallConv) {
```
christopherbate, line 860: Missing test for `plan-alloc-tensors`,

You'll need to add a `RUN` line in `test/Dialect/Plan/plan-alloc-tensors.mlir` where this option is invoked with `true` and add `CHECK` statements to show what is the desired result.

mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/OutlineClusters.cpp
```diff
@@ -292,8 +302,13 @@ createFunctionsFromRegions(RewriterBase &rewriter, Region &region,
       return WalkResult::advance();
     }
 
-    if (auto dpsGroup = dyn_cast<plan::InlineClosedGroupOp>(op)) {
-      if (failed(outlineTensorRTRegion(rewriter, dpsGroup)))
+    if (auto group = dyn_cast<plan::InlineClosedGroupOp>(op)) {
+      if (failed(outlineTensorRTRegion(rewriter, group)))
+        return WalkResult::interrupt();
+      return WalkResult::advance();
```
christopherbate, line 308: Don't use `else` after `return` 

https://llvm.org/docs/CodingStandards.html#don-t-use-else-after-a-return

mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/OutlineClusters.cpp
```diff
@@ -264,6 +266,13 @@ static LogicalResult outlineTensorRTRegion(RewriterBase &rewriter,
   return success();
 }
 
+static LogicalResult outlineTensorRTRegion(RewriterBase &rewriter,
+                                           plan::InlineClosedGroupNonDPSOp op) {
+  llvm_unreachable("outlinining non-dps inline closed group ops to tensorrt "
```
christopherbate, line 271: when `llvm_unreachable` is used, it results in a crash/abort, which is not desired. Error should be propagated to callers so that the pass can signal failure. Just `return op->emitError(.....)` should be sufficient here.

mlir-tensorrt/compiler/include/mlir-tensorrt/Dialect/Plan/Transforms/Passes.td
```diff
@@ -248,6 +248,9 @@ def StablehloClusteringPass : Pass<"stablehlo-clustering", "::mlir::ModuleOp"> {
     Option<"entrypoint", "entrypoint", "std::string", "\"\"",
       "the name of the entrypoint function; if empty then the clustering runs"
       " on all functions">,
+    Option<"useNonDPSCallConv",
+      "use-non-dps-call-conv", "bool", "false",
+      "allow tensorrt based output allocations using output allocator">,
```
christopherbate, line 253: Should be "allow backend clusters to directly allocate outputs"

mlir-tensorrt/compiler/include/mlir-tensorrt/Dialect/Plan/Transforms/Passes.td
```diff
@@ -332,7 +335,10 @@ def CreateClosedRegionsPass : Pass<"plan-create-closed-regions", "::mlir::Module
     Option<"testPreWalkOrder", "test-pre-walk-order", "bool", "false",
       "(used only in testing) specifies to outline regions by walking in "
       " pre-order; used for verifying results are not sensitive "
-      "to traversal order">
+      "to traversal order">,
+    Option<"useNonDPSCallConv", "use-non-dps-call-conv", "bool", 
+           /*default=*/"false", 
+           "Allow TensorRT-based output allocations using output allocator">
```
christopherbate, line 341: Should be "allow backend clusters to directly allocate outputs"

mlir-tensorrt/compiler/include/mlir-tensorrt/Dialect/Plan/Transforms/Passes.td
```diff
@@ -428,6 +434,13 @@ def PlanAllocTensorsPass : Pass<"plan-alloc-tensors",
     "::mlir::bufferization::BufferizationDialect",
     "::mlir::plan::PlanDialect"
   ];
+
+  let options = [
+    Option<"useNonDPSCallConv", "use-non-dps-call-conv", "bool", 
+           /*default=*/"false", 
+           "Allow TensorRT-based output allocations using output allocator">
```
christopherbate, line 441: Should be "allow backend clusters to directly allocate outputs"

mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/CreateClosedRegions.cpp
```diff
@@ -351,70 +351,11 @@ static void remapAnalysisState(DataFlowSolver &solver, ValueRange originals,
   }
 }
```
christopherbate, line 352: You have lots of changes to this pass, so there the test `test/Dialect/Plan/create-closed-regions.mlir` should also be updated/expanded.

mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/Passes.cpp
```diff
@@ -73,6 +76,11 @@ struct ClusteringPipelineCliOpts
     : public PassPipelineOptions<ClusteringPipelineCliOpts> {
   Option<std::string> entrypoint{*this, "entrypoint", llvm::cl::init(""),
                                  llvm::cl::desc("name of entrypoint function")};
+  Option<bool> useNonDPSCallConv{
+      *this, "use-non-dps-call-conv",
+      llvm::cl::desc(
+          "allow tensorrt based output allocations using output allocator"),
```
christopherbate, line 82: Please correct doc comment throughout similar to other comments for this option.

christopherbate (changes requested):
Overall, it looks good, but it is missing updates to tests for the two main passes that were changed (`alloc-tensors` and `create-closed-regions`).

In addition, there should be new tests in 

- `test/Dialect/Plan/invalid.mlir` 
-  `test/Dialect/Plan/roundtrip.mlir` 

Specifically for the new operation that was added.

jhalakpatel:
> Overall, it looks good, but it is missing updates to tests for the two main passes that were changed (`alloc-tensors` and `create-closed-regions`).
> 
> In addition, there should be new tests in
> 
> * `test/Dialect/Plan/invalid.mlir`
> * `test/Dialect/Plan/roundtrip.mlir`
> 
> Specifically for the new operation that was added.

@christopherbate I have addressed your comments. Please review again.

mlir-tensorrt/compiler/include/mlir-tensorrt/Dialect/Plan/IR/PlanOps.td
```diff
@@ -255,16 +273,75 @@ def Plan_InlineClosedGroupOp : Plan_GroupOpBase<"inline_closed_group", [
...
+  DeclareOpInterfaceMethods<RegionBranchOpInterface,
+    ["getEntrySuccessorOperands"]>,
+  DeclareOpInterfaceMethods<OpAsmOpInterface,
+    ["getAsmBlockArgumentNames"]>
+]> {
+  let description = [{
+  The `plan.inline_closed_alloc_group` operation is a variant of the
+  `plan.inline_closed_group` operation that does not use destination-passing style
+  (DPS). It is isolated from above and explicitly captures input operands,
+  but unlike its DPS counterpart, it does not capture destination operands.
```
christopherbate, line 295:
```suggestion
  but unlike its DPS counterpart, it does not capture destination operands because its results must be lowered to allocation(s). The allocations may  or may not be of a size that can only be computed inside of the region.
```

mlir-tensorrt/compiler/include/mlir-tensorrt/Dialect/Plan/IR/PlanOps.td
```diff
@@ -255,16 +273,75 @@ def Plan_InlineClosedGroupOp : Plan_GroupOpBase<"inline_closed_group", [
...
+
+  Consider the following simple program containing operations with dynamically shaped operands:
+
+  ```mlir
+  %0 = ... : tensor<?xf32> // A dynamically shaped operand
+  %1 = ... : index         // A dynamic calculation of %0's extent
 
+  %2 = plan.inline_closed_alloc_group target(#plan.cluster_target<tensorrt>)
+    inputs(%0, %1 : tensor<?xf32>, index)
+    in_attrs [#plan.bounds<shape, , >, #plan.bounds<none>] -> tensor<?xf32> {
```
christopherbate, line 312: This syntax looks incorrect: `#plan.bounds<shape, , >`, did you mean to fill in something there?

mlir-tensorrt/compiler/lib/Dialect/Plan/IR/PlanOps.cpp
```diff
@@ -465,6 +467,68 @@ void InlineClosedGroupOp::build(OpBuilder &b, OperationState &state,
...
 }
 
+//===----------------------------------------------------------------------===//
+// InlineClosedAllocGroupOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult InlineClosedAllocGroupOp::verify() {
+  Operation *op = getOperation();
+  // Check for res_attrs
+  if (op->hasAttr("res_attrs")) {
```
christopherbate, line 477: Nit: remove braces on single-statement bodies of `if` statements 

https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements

mlir-tensorrt/compiler/lib/Dialect/Plan/IR/PlanOps.cpp
```diff
@@ -465,6 +467,68 @@ void InlineClosedGroupOp::build(OpBuilder &b, OperationState &state,
...
+OperandRange
+InlineClosedAllocGroupOp::getEntrySuccessorOperands(RegionBranchPoint point) {
+  return getOperands();
+}
+
+void InlineClosedAllocGroupOp::getAsmBlockArgumentNames(
+    Region &region, OpAsmSetValueNameFn setNameFn) {
+  assert(region.front().getNumArguments() == getInputs().size() &&
+         "expected one block arg for each input argument");
+  unsigned numInputs = getInputs().size();
```
christopherbate, line 504: Not really a need for the assertion here. The code here isn't going to crash if the assertion fails. `numInputs` will be an unused variable when assertions are turned off.

I believe that `region.front().getArguments()` can also be replaced with just `region.getArguments()`. A Region's arguments are the arguments of the entry block.

mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/CreateClosedRegions.cpp
```diff
@@ -351,70 +351,11 @@ static void remapAnalysisState(DataFlowSolver &solver, ValueRange originals,
...
-  }
-
-  // Since we are about to replace values that may be inputs to other regions
-  // ops, we need to update the solver to populate the replacement TensorKind
-  // information.
-  remapAnalysisState(solver, op->getResults(), replacements);
-  rewriter.replaceOp(op, replacements);
+static FailureOr<SmallVector<BoundsAttr>>
+getInputAttributes(RewriterBase &rewriter, DataFlowSolver &solver, Location loc,
+                   SmallVector<Value> const &inputs) {
```
christopherbate, line 356:
```suggestion
                  ValueRange inputs) {
```

In general, use `ValueRange` and `TypeRange` to represent const range over values and types, respectively. For other cases, use `ArrayRef<T>` rather than `const SmallVector<T> &`. For mutable SmallVector, use `SmallVectorImpl<T> &` to allow for different implicit size parameters. See note here https://llvm.org/docs/ProgrammersManual.html#llvm-adt-smallvector-h.

mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/CreateClosedRegions.cpp
```diff
@@ -425,30 +366,16 @@ static LogicalResult createClosedGroupOp(RewriterBase &rewriter,
     const TensorKindLattice *tensorKindLattice =
         solver.lookupState<TensorKindLattice>(input);
     if (!tensorKindLattice || tensorKindLattice->getValue().isUninitialized())
-      return closedGroupOp->emitOpError("input operand #")
-             << idx << " of type " << input.getType()
+      return emitError(loc)
+             << ("input operand #") << idx << " of type " << input.getType()
```
christopherbate, line 370: Nit: parentheses can be removed here

mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/CreateClosedRegions.cpp
```diff
@@ -503,11 +430,162 @@ static LogicalResult createClosedGroupOp(RewriterBase &rewriter,
...
+
+static LogicalResult materializeDestinationOperands(
+    RewriterBase &rewriter, plan::InlineGroupOp op, DataFlowSolver &solver,
+    SmallVector<DestinationOperandMaterializationResult> &destinationOperands) {
+  destinationOperands.reserve(op.getNumResults());
+  for (OpResult res : op->getOpResults()) {
+    FailureOr<DestinationOperandMaterializationResult> destResult =
+        materializeDestinationOperand(rewriter, res.getLoc(), op,
+                                      res.getResultNumber(), solver);
+    if (failed(destResult)) {
```
christopherbate, line 445: Nit: remove curly braces for single-statement bodies of `if` statements 
https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements

mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/CreateClosedRegions.cpp
```diff
@@ -503,11 +430,162 @@ static LogicalResult createClosedGroupOp(RewriterBase &rewriter,
...
+    }
+    destinationOperands.push_back(*destResult);
+  }
   return success();
 }
 
+// Helper function to create DPS closed group op.
+static LogicalResult createInlineClosedGroupOp(
+    RewriterBase &rewriter, plan::InlineGroupOp op, DataFlowSolver &solver,
+    const SmallVector<Value> &inputs,
```
christopherbate, line 458:
```suggestion
   ValueRange inputs,
```

mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/CreateClosedRegions.cpp
```diff
@@ -503,11 +430,162 @@ static LogicalResult createClosedGroupOp(RewriterBase &rewriter,
...
+    const SmallVector<DestinationOperandMaterializationResult>
```
christopherbate, line 459: Please use

```
    ArrayRef<DestinationOperandMaterializationResult>
        destinationOperands) {
```

mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/CreateClosedRegions.cpp
```diff
@@ -503,11 +430,162 @@ static LogicalResult createClosedGroupOp(RewriterBase &rewriter,
...
+  rewriter.inlineBlockBefore(
+      &op.getRegion().front(), &closedGroupOp.getRegion().front(),
+      closedGroupOp.getRegion().front().end(),
+      closedGroupOp.getRegion().getArguments().take_front(
+          op.getRegion().getNumArguments()));
+
+  SmallVector<Value> replacements;
+  replacements.reserve(destinationOperands.size());
+  for (auto [newResult, destOperand, originalType] :
+       llvm::zip(closedGroupOp->getResults(), destinationOperands,
```
christopherbate, line 477: Use `zip_equal` if all lists are supposed to be the same size (and it will assert as such). Otherwise the range is limited by the smallest element.

mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/CreateClosedRegions.cpp
```diff
@@ -503,11 +430,162 @@ static LogicalResult createClosedGroupOp(RewriterBase &rewriter,
...
+        BoundsKind::Shape, ArrayRef(dest.constantShapeLowerBound),
+        ArrayRef(dest.constantShapeUpperBound));
+    if (!boundsAttr)
+      return failure();
+    resultAttrs.push_back(boundsAttr);
+  }
+  closedGroupOp.setResAttrsAttr(resultAttrs);
+
+  // Create the closed region input profilw attrs.
+  auto inputAttr = getInputAttributes(rewriter, solver, closedGroupOp->getLoc(),
```
christopherbate, line 513: Replace `auto` with actual typename. `auto` is only encouraged when type is obvious (e.g. due to template parameter on RHS of `=`) or very long/complicated (e.g. specialized iterator types).

See https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable

mlir-tensorrt/test/Dialect/Plan/create-closed-regions.mlir
```diff
@@ -59,6 +69,21 @@ func.func @test_simple_shape_bound(%arg0: tensor<?x10xf32> {tensorrt.shape_profi
...
+//       CHECK-ALLOC:      inputs(%[[arg0]], %[[dim]], %[[c10]] : tensor<?x10xf32>, index, index)
+//       CHECK-ALLOC:      in_attrs [#plan.bounds<shape, [1, 10], [40, 10]>, #plan.bounds<none>, #plan.bounds<none>]
+//       CHECK-ALLOC:      -> tensor<?x10xf32> {
+//       CHECK-ALLOC:     ^bb0(%[[in:.+]]: tensor<?x10xf32>, %[[in_1:.+]]: index, %[[in_2:.+]]: index):
+//       CHECK-ALLOC:       %[[v3:.+]] = stablehlo.exponential %[[in]] : tensor<?x10xf32>
+//       CHECK-ALLOC:       %[[v4:.+]] = with_shape %[[v3]](%[[in_1]], %[[in_2]]) :
+//       CHECK-ALLOC:       yield %[[v4]] : tensor<?x10xf32>
+//       CHECK-ALLOC:     return %[[v2]] : tensor<?x10xf32>
+
 // -----
```
christopherbate, line 87: Include at least a `CHECK-ALLOC-LABEL` for the next test. Otherwise you don't know whether the `CHECK-ALLOC` statements in this last test are actually bleeding through the end of the test and matching the result of some other test segment. It is a common source of false positives that go undetected in large test files with multiple FileCheck prefixes in use.

mlir-tensorrt/test/Dialect/Plan/plan-alloc-tensors.mlir
```diff
@@ -232,6 +303,12 @@ func.func @test_return_arg(%arg0: tensor<10xf32>) -> tensor<10xf32> {
 //       CHECK:     %[[v0:.+]] = bufferization.materialize_in_destination %[[arg0]] in %[[arg1]]
 //       CHECK:     return %[[v0]]
 
+// CHECK-ALLOC-LABEL: @test_return_arg
+//  CHECK-ALLOC-SAME: (%[[arg0:.+]]: tensor<10xf32>)
+//       CHECK-ALLOC:     return %[[arg0]]
```
christopherbate, line 308:
```suggestion
//  CHECK-ALLOC-NEXT:     return %[[arg0]]

```

christopherbate (approved):
Some additional more minor comments, otherwise LGTM.

mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/CreateClosedRegions.cpp
jhalakpatel: Thanks a lot! This is super helpful.

PR #287, opened 2024-10-18
[Dialect/TensorRT] Update tensorrrt dialect to use non-dps calling convention

jhalakpatel:
Add non-DPS variant of `tensorrt::CallOp` i.e. `tensorrt::CallAllocOp`. Add conversion from call alloc op to runtime dialect enqueue alloc op.

### DIFF
--- a/mlir-tensorrt/compiler/lib/Conversion/TensorRTToTensorRTRuntime/TensorRTToTensorRTRuntime.cpp
+++ b/mlir-tensorrt/compiler/lib/Conversion/TensorRTToTensorRTRuntime/TensorRTToTensorRTRuntime.cpp
@@ -57,6 +57,78 @@ struct RewriteConstants : public OpRewritePattern<tensorrt::ConstantOp> {
   }
 };
 
+// Helper function to convert both CallOp and CallAllocOp
+static LogicalResult
+convertCallOp(Operation *op, IRRewriter &rewriter,
+              SymbolTableCollection &symbolTable, DataFlowSolver &solver,
+              ModuleOp module,
+              SmallVectorImpl<tensorrt::TensorRTModuleOp> &trtModules) {
+  MLIRContext *ctx = rewriter.getContext();
+  Location loc = op->getLoc();
+  ValueRange inputs;
+  ValueRange outputs;
+  func::FuncOp trtFunc;
+
+  if (auto callOp = dyn_cast<tensorrt::CallOp>(op)) {
+    trtFunc = callOp.getFuncCallee(symbolTable);
+    inputs = callOp.getInputs();
+    outputs = callOp.getOutputs();
+  } else if (auto callAllocOp = dyn_cast<tensorrt::CallAllocOp>(op)) {
+    trtFunc = callAllocOp.getFuncCallee(symbolTable);
+    inputs = callAllocOp.getInputs();
+    // CallAllocOp doesn't have outputs as operands
+  } else {
+    llvm_unreachable("unexpected type of operation. Only callOp and "
+                     "callAllocOp are supported.");
+    return failure();
+  }
+
+  solver.load<dataflow::DeadCodeAnalysis>();
+  solver.load<dataflow::SparseConstantPropagation>();
+  solver.load<TensorKindAnalysis>(symbolTable);
+  if (failed(solver.initializeAndRun(trtFunc)))
+    return trtFunc.emitError() << "failed to run TensorKindAnalysis";
+
+  // Check which tensors should be host tensors.
+  SmallVector<int64_t> hostTensorArgs;
+  for (auto [idx, arg] : llvm::enumerate(trtFunc.getArguments())) {
+    const TensorKindLattice *kind = solver.lookupState<TensorKindLattice>(arg);
+    RankedTensorType rtt = cast<RankedTensorType>(arg.getType());
+    // To be conservative, we only do this if type is i32 and num elements
+    // <= 8.
+    if (kind && !kind->getValue().isUninitialized() &&
+        kind->getValue().isHostVisible() &&
+        rtt.getElementType().isInteger(32) && rtt.getNumElements() <= 8)
+      hostTensorArgs.push_back(idx);
+  }
+
+  rewriter.setInsertionPoint(op);
+
+  Value executionContext = rewriter.create<trtrt::CompileOp>(
+      loc,
+      SymbolRefAttr::get(rewriter.getStringAttr(*trtModules.front().getName()),
+                         {FlatSymbolRefAttr::get(trtFunc)}));
+  Value stream = rewriter.create<cuda::GetGlobalStreamOp>(loc, 0);
+
+  Operation *enqueueOp;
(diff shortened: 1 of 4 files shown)


mlir-tensorrt/compiler/lib/Conversion/TensorRTToTensorRTRuntime/TensorRTToTensorRTRuntime.cpp
```diff
@@ -57,6 +57,80 @@ struct RewriteConstants : public OpRewritePattern<tensorrt::ConstantOp> {
   }
 };
 
+// Helper function to convert both CallOp and CallAllocOp
+static LogicalResult
+convertCallOp(Operation *op, MLIRContext *ctx, IRRewriter &rewriter,
```
christopherbate, line 62: Don't include `MLIRContext` in the argument list.... the context can be grabbed from any of the other arguments

mlir-tensorrt/compiler/lib/Conversion/TensorRTToTensorRTRuntime/TensorRTToTensorRTRuntime.cpp
```diff
@@ -57,6 +57,80 @@ struct RewriteConstants : public OpRewritePattern<tensorrt::ConstantOp> {
...
+    llvm_unreachable("unexpected type of operation. Only callOp and "
+                     "callAllocOp are supported.");
+    return failure();
+  }
+
+  solver.load<dataflow::DeadCodeAnalysis>();
+  solver.load<dataflow::SparseConstantPropagation>();
+  solver.load<TensorKindAnalysis>(symbolTable);
+  if (failed(solver.initializeAndRun(trtFunc))) {
+    trtFunc.emitError() << "failed to run TensorKindAnalysis";
```
christopherbate, line 89: These two can be combined into one line `return trtFunc.emitError()...`

mlir-tensorrt/compiler/lib/Conversion/TensorRTToTensorRTRuntime/TensorRTToTensorRTRuntime.cpp
```diff
@@ -57,6 +57,80 @@ struct RewriteConstants : public OpRewritePattern<tensorrt::ConstantOp> {
...
+            ? DenseI64ArrayAttr{}
+            : DenseI64ArrayAttr::get(ctx, hostTensorArgs));
+  } else {
+    enqueueOp = rewriter.create<trtrt::EnqueueAllocOp>(
+        loc, op->getResultTypes(), executionContext, stream, inputs,
+        /*host_tensors_args=*/hostTensorArgs.empty()
+            ? DenseI64ArrayAttr{}
+            : DenseI64ArrayAttr::get(ctx, hostTensorArgs));
+  }
+  rewriter.setInsertionPointAfter(enqueueOp);
```
christopherbate, line 128: Changing the insertion point here shouldn't have any effect. After creation of op, the insertion point is already after the op.

mlir-tensorrt/compiler/lib/Conversion/TensorRTToTensorRTRuntime/TensorRTToTensorRTRuntime.cpp
```diff
@@ -72,6 +146,8 @@ class ConvertTensorRTToRuntimePass
     }
 
     IRRewriter rewriter(ctx);
+    SymbolTableCollection symbolTable;
+    DataFlowSolver solver;
```
christopherbate, line 150: You're doing something different there, keeping a single solver and repeatedly re-initializing it? I've never seen that pattern before, not sure if it will actually work as intended.

christopherbate (changes requested):

mlir-tensorrt/compiler/lib/Conversion/TensorRTToTensorRTRuntime/TensorRTToTensorRTRuntime.cpp
jhalakpatel: I did not realized that we need should not keep a single solver while iterating through different operations. I have reverted my change.

christopherbate: Yeah what needs to be done to keep a solver is to initialize it once. Then when you do replace/rewrite, you need to make corresponding small adjustments/remaps to the solver state. We do this in the Plan dialect transforms, and ideally we can do that here as well, but it's not especially high priority.

mlir-tensorrt/tensorrt/include/mlir-tensorrt-dialect/TensorRT/IR/TensorRTOps.td
```diff
@@ -71,7 +71,9 @@ def TensorRT_TensorRTModuleOp : TensorRT_Op<"module", [
 
   let extraClassDeclaration = [{
     /// Return the name of this module if present.
-    std::optional<StringRef> getName() { return getSymName(); }
+    std::optional<StringRef> getName() const {
+      return (*this)->getAttrOfType<StringAttr>("sym_name").getValue();
```
christopherbate, line 75: Why was this changed? Generally you don't want to rely on accessing properties through the string literal name.

mlir-tensorrt/tensorrt/lib/TensorRT/IR/TensorRT.cpp
```diff
@@ -192,6 +192,39 @@ LogicalResult CallOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
...
+
+LogicalResult
+CallAllocOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
+  func::FuncOp kernel = getFuncCallee(symbolTable);
+  if (!kernel)
+    return emitOpError() << "no valid kernel found with symbol name "
+                         << getCallee();
+  FunctionType funcType = kernel.getFunctionType();
+
+  // TODO: For dynamic shapes, we allow passing a larger linear bufer.
```
christopherbate, line 214: Remove TODO? This TODO is talking about DPS, which we don't use here.

christopherbate (approved):
Minor comments otherwise LGTM

mlir-tensorrt/tensorrt/include/mlir-tensorrt-dialect/TensorRT/IR/TensorRTOps.td
jhalakpatel: It was required because we were passing trtModule why ArrayRef which I think is equivalent to passing by `const &`.
Accessing methods such as `getSymName` require the calling object to be `non const`.

I have reverted this change and instead now passing trtModule as `SmallVectorImpl<tensorrt::TensorRTModuleOp>& trtModules)`

christopherbate:
@jhalakpatel Looks like it was merged without some of the changes you said you made. Fix-forward is fine, but current state is not good.

PR #288, opened 2024-10-18
[Runtime/Lua]  Support multiple results in executeFunctionWithLuaBackend

jhalakpatel:


christopherbate:
IIUC, we should be able to add tests here?

jhalakpatel:
> IIUC, we should be able to add tests here?

can you point me to an example of such tests? IIUC, we need complete IR translation to work to reach a point where function actually returns a memref value to actually the test the code being added. I will be adding several integration tests in the later PR once core changes are merged.

mlir-tensorrt/executor/include/mlir-executor/Runtime/Backend/Lua/LuaRuntime.h
```diff
@@ -93,12 +93,14 @@ StatusOr<int64_t> runExecutorExecutable(
...
-/// output (destination args). Returns any results.
+/// output (destination args). Returns optional results supporting both DPS and
+/// non-DPS style calling convention.
 StatusOr<llvm::SmallVector<std::unique_ptr<RuntimeValue>>>
 executeFunctionWithLuaBackend(LuaRuntimeSession &session, std::string_view name,
                               llvm::ArrayRef<RuntimeValue *> inputArgs,
                               llvm::ArrayRef<RuntimeValue *> outputArgs,
-                              std::optional<CudaStream> stream = {});
+                              std::optional<CudaStream> stream = {},
+                              std::optional<RuntimeClient *> client = {});
```
christopherbate, line 103: You'll have to explain what this argument is for/when to provide it in the documentation comment.

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/LuaRuntime.cpp
```diff
@@ -502,12 +502,138 @@ static Status validateArgsTypesAgainstFuncArgs(const RuntimeValue *runArg,
   return getOkStatus();
 }
 
+[[maybe_unused]] static constexpr int MEMREF_FIXED_FIELDS =
```
christopherbate, line 505: For constant integers, our convention is to use `int64_t` and naming as follows:

```
static constexpr int64_t kMemRefFixedFields = ...
```

jhalakpatel: This is not present in the final code anymore.

shelkesagar29:
For failing test, 
1. Lets add new test to a folder named `Lua`.
2. `Lua` folder should be compiled only if `MLIR_EXECUTOR_ENABLE_GPU_INTEGRATION_TESTS ` is ON
```
if(MLIR_EXECUTOR_ENABLE_GPU_INTEGRATION_TESTS)
    add_subdirectory(Lua)
endif()
```

INT4 tests don't need GPU

shelkesagar29:
Also, commit message doesn't follow canonical format. Can you please fix it?

shelkesagar29 (approved):

PR #289, opened 2024-10-18
Enable end to end non-DPS testing

jhalakpatel:
Implement python binding changes to allow execute function return
multiple returns. Update tests to use non-DPS style calling convention.

Also, enable end to end lowering by enabling conversion of closed alloc group op to tensorrt dialect.

Miscellaneous fixes:
1. Add missing handling of `CallAllocOp` in EliminateShapeOps pass.
2. Skip non ranked tensor type function arguments while collecting host  tensor arguments.
3. Temporarily add a pass to remove clone operation in MemRefToExecutor dialect conversion.
4. Relax memref creation for empty shape tensors.
5. Fix memref life returned from Lua function results. This required session allocator to track returned memref.
6. Return error status instead of silently erroring out.

### DIFF
--- a/mlir-tensorrt/compiler/lib/Conversion/TensorRTRuntimeToExecutor/TensorRTRuntimeToExecutor.cpp
+++ b/mlir-tensorrt/compiler/lib/Conversion/TensorRTRuntimeToExecutor/TensorRTRuntimeToExecutor.cpp
@@ -379,20 +379,22 @@ struct ConvertEnqueueAllocToCall
 
     // Create output memrefs from output descriptors
     SmallVector<Value> results;
+    // Initialize output descriptor offset to skip number of results.
+    // `outputDescOffset` is used to retrieve rank, ptr, shapes, and strides per
+    // result.
+    outputDescOffset = 1;
     for (unsigned i = 0; i < op->getNumResults(); ++i) {
       unsigned rank = cast<MemRefType>(op->getResult(i).getType()).getRank();
-      unsigned offset =
-          1 +
-          i * (2 * rank + 2); // num res, (i * (rank, ptr, [shape], [stride]))
-
       Value rankOffset = b.create<executor::GetOffsetOp>(
           b.getI64Type(), structType,
-          ArrayRef<OpFoldResult>{this->createIndexConstant(b, 0),
-                                 rewriter.getI64IntegerAttr(offset++)});
+          ArrayRef<OpFoldResult>{
+              this->createIndexConstant(b, 0),
+              rewriter.getI64IntegerAttr(outputDescOffset++)});
       Value devicePtrOffset = b.create<executor::GetOffsetOp>(
           b.getI64Type(), structType,
-          ArrayRef<OpFoldResult>{this->createIndexConstant(b, 0),
-                                 rewriter.getI64IntegerAttr(offset++)});
+          ArrayRef<OpFoldResult>{
+              this->createIndexConstant(b, 0),
+              rewriter.getI64IntegerAttr(outputDescOffset++)});
 
       [[maybe_unused]] Value rankValue = b.create<executor::LoadOp>(
           b.getI64Type(), outputDescriptors, rankOffset);
@@ -406,8 +408,9 @@ struct ConvertEnqueueAllocToCall
       for (unsigned r = 0; r < rank; ++r) {
         Value shapeOffset = b.create<executor::GetOffsetOp>(
             b.getI64Type(), structType,
-            ArrayRef<OpFoldResult>{this->createIndexConstant(b, 0),
-                                   rewriter.getI64IntegerAttr(offset++)});
+            ArrayRef<OpFoldResult>{
+                this->createIndexConstant(b, 0),
+                rewriter.getI64IntegerAttr(outputDescOffset++)});
         Value shape = b.create<executor::LoadOp>(
             b.getI64Type(), outputDescriptors, shapeOffset);
         shapes.push_back(shape);
@@ -416,8 +419,9 @@ struct ConvertEnqueueAllocToCall
       for (unsigned r = 0; r < rank; ++r) {
         Value strideOffset = b.create<executor::GetOffsetOp>(
             b.getI64Type(), structType,
-            ArrayRef<OpFoldResult>{this->createIndexConstant(b, 0),
-                                   rewriter.getI64IntegerAttr(offset++)});
+            ArrayRef<OpFoldResult>{
+                this->createIndexConstant(b, 0),
+                rewriter.getI64IntegerAttr(outputDescOffset++)});
         Value shape = b.create<executor::LoadOp>(
             b.getI64Type(), outputDescriptors, strideOffset);
         shapes.push_back(shape);
--- a/mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/EliminateShapeOps.cpp
+++ b/mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/EliminateShapeOps.cpp
@@ -65,17 +65,26 @@ struct RemoveWithValuesRewriter : public OpRewritePattern<plan::WithValuesOp> {
 } // namespace
 
 /// Get a map from `tensorrt.func` functions to associated `tensorrt.call`
-/// operations.
-static llvm::DenseMap<func::FuncOp, SmallVector<tensorrt::CallOp>>
+/// and `tensorrt.call_alloc` operations.
+static llvm::DenseMap<func::FuncOp, SmallVector<Operation *>>
 getTensorRTFunctionCallMap(ModuleOp op, SymbolTableCollection &collection) {
-  llvm::DenseMap<func::FuncOp, SmallVector<tensorrt::CallOp>> map;
-  op->walk([&](tensorrt::CallOp callOp) {
-    func::FuncOp func = callOp.getFuncCallee(collection);
-    if (map.contains(func)) {
-      map[func].push_back(callOp);
+  llvm::DenseMap<func::FuncOp, SmallVector<Operation *>> map;
+  op->walk([&](Operation *callOp) {
+    if (!isa<tensorrt::CallOp, tensorrt::CallAllocOp>(callOp))
       return;
+
+    func::FuncOp func;
+    if (auto call = dyn_cast<tensorrt::CallOp>(callOp)) {
+      func = call.getFuncCallee(collection);
+    } else {
+      auto callAlloc = dyn_cast<tensorrt::CallAllocOp>(callOp);
+      func = callAlloc.getFuncCallee(collection);
     }
-    map.insert(std::make_pair(func, SmallVector<tensorrt::CallOp>{callOp}));
+
+    if (map.count(func))
+      map[func].push_back(callOp);
+    else
+      map.insert({func, SmallVector<Operation *>{callOp}});
   });
   return map;
 }
@@ -84,7 +93,7 @@ getTensorRTFunctionCallMap(ModuleOp op, SymbolTableCollection &collection) {
 /// `tensorrt.call` operations.
 static LogicalResult removeUnusedArgs(SymbolTableCollection &collection,
                                       ModuleOp op, func::FuncOp funcOp,
-                                      ArrayRef<tensorrt::CallOp> callOps) {
+                                      ArrayRef<Operation *> callOps) {
   llvm::SmallBitVector unusedArgs(funcOp.getNumArguments(), 0);
   for (BlockArgument arg : funcOp.getArguments()) {
     if (arg.use_empty())
@@ -99,10 +108,16 @@ static LogicalResult removeUnusedArgs(SymbolTableCollection &collection,
     funcOp.eraseArgument(i);
 
     // Update the call ops.
-    for (tensorrt::CallOp callOp : callOps)
-      callOp.getInputsMutable().erase(i);
+    for (Operation *callOp : callOps) {
+      if (auto call = dyn_cast<tensorrt::CallOp>(callOp))
+        call.getInputsMutable().erase(i);
+      else if (auto callAlloc = dyn_cast<tensorrt::CallAllocOp>(callOp))
+        callAlloc.getInputsMutable().erase(i);
+      else
+        return emitError(funcOp->getLoc())
+               << "Unexpected operation type in callOps";
+    }
   }
-
   return success();
 }
 
--- a/mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/OutlineClusters.cpp
+++ b/mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/OutlineClusters.cpp
@@ -156,28 +156,32 @@ getTensorRTShapeProfile(plan::BoundsAttr attr, Value v) {
   return getProfileAttr(attr.getMinShape(), attr.getMaxShape());
 }
 
-static LogicalResult outlineTensorRTRegion(RewriterBase &rewriter,
-                                           plan::InlineClosedGroupOp op) {
-  tensorrt::TensorRTModuleOp trtModuleOp = getOrCreateTensorRTModuleOp(op);
-  auto funcArgTypes = llvm::to_vector(TypeRange(op.getInputs()));
-  FailureOr<FunctionOpInterface> func = createOutlinedFunc(
-      rewriter, op.getLoc(), op, trtModuleOp, "tensorrt_cluster",
-      "cluster.tensorrt", TypeRange(op.getInputs()),
-      op.getYield()->getOperandTypes());
-  if (failed(func))
-    return failure();
-  assert(func->getFunctionBody().getBlocks().size() == 1 &&
-         "expected body with one block");
-  func->setPublic();
+template <typename OpType>
+static auto createCallOp(RewriterBase &rewriter, OpType op,
+                         tensorrt::TensorRTModuleOp trtModuleOp,
+                         FunctionOpInterface func) {
+  static_assert(
+      std::is_same_v<OpType, plan::InlineClosedGroupOp> ||
+          std::is_same_v<OpType, plan::InlineClosedAllocGroupOp>,
+      "OpType must be either InlineClosedGroupOp or InlineClosedAllocGroupOp");
+  if constexpr (std::is_same_v<OpType, plan::InlineClosedGroupOp>)
+    return rewriter.create<tensorrt::CallOp>(
+        op.getLoc(), op.getResultTypes(), op.getInputs(), op.getOuts(),
+        SymbolRefAttr::get(trtModuleOp.getNameAttr(),
+                           {FlatSymbolRefAttr::get(func)}));
+  else if constexpr (std::is_same_v<OpType, plan::InlineClosedAllocGroupOp>)
+    return rewriter.create<tensorrt::CallAllocOp>(
+        op.getLoc(), op.getResultTypes(), op.getInputs(),
+        SymbolRefAttr::get(trtModuleOp.getNameAttr(),
+                           {FlatSymbolRefAttr::get(func)}));
+}
 
-  rewriter.setInsertionPoint(op);
-  auto callOp = rewriter.create<tensorrt::CallOp>(
-      op.getLoc(), op.getResultTypes(), op.getInputs(), op.getOuts(),
-      SymbolRefAttr::get(trtModuleOp.getNameAttr(),
-                         {FlatSymbolRefAttr::get(*func)}));
-
+template <typename OpType>
+static LogicalResult populateFunctionAttributes(RewriterBase &rewriter,
+                                                OpType op,
+                                                FunctionOpInterface *func) {
   // Populate the function arguments attributes.
-  for (unsigned i = 0; i < (*func).getNumArguments(); i++) {
+  for (unsigned i = 0; i < func->getNumArguments(); i++) {
     BoundsAttr srcAttr = cast<BoundsAttr>(op.getInputAttrs()[i]);
     // We may have scalar (index|signless int)-typed values since we haven't
     // eliminated `plan.(with_shape|with_values)` ops yet.
@@ -202,30 +206,57 @@ static LogicalResult outlineTensorRTRegion(RewriterBase &rewriter,
     func->setArgAttr(i, mlir::getHostTensorArgAttrName(),
                      rewriter.getUnitAttr());
   }
-  // Populate the function result attributes.
-  for (unsigned i = 0; i < (*func).getNumResults(); i++) {
-    BoundsAttr srcAttr = cast<BoundsAttr>(op.getResAttrs()[i]);
-    if (srcAttr.isNone())
-      continue;
-    FailureOr<tensorrt::ShapeProfileAttr> boundsAttr =
-        getTensorRTShapeProfile(srcAttr, op.getResults()[i]);
-    if (failed(boundsAttr))
-      return op->emitOpError("failed to create TensorRT shape profile "
-                             "attribute from Plan BoundsAttr for result #")
-             << i << " (" << srcAttr << ")";
-    if (srcAttr.isShapeBound()) {
+  // Populate the function result attributes for DPS call op.
+  if constexpr (std::is_same_v<OpType, plan::InlineClosedGroupOp>) {
+    for (unsigned i = 0; i < func->getNumResults(); i++) {
+      BoundsAttr srcAttr = cast<BoundsAttr>(op.getResAttrs()[i]);
+      if (srcAttr.isNone())
+        continue;
+      FailureOr<tensorrt::ShapeProfileAttr> boundsAttr =
+          getTensorRTShapeProfile(srcAttr, op.getResults()[i]);
+      if (failed(boundsAttr))
+        return op->emitOpError("failed to create TensorRT shape profile "
+                               "attribute from Plan BoundsAttr for result #")
+               << i << " (" << srcAttr << ")";
+      if (srcAttr.isShapeBound()) {
+        func->setResultAttr(
+            i, tensorrt::TensorRTDialect::getShapeProfileArgAttrName(),
+            *boundsAttr);
+        continue;
+      }
+      assert(srcAttr.isValueBound() && "expected value bound or shape bound");
       func->setResultAttr(
-          i, tensorrt::TensorRTDialect::getShapeProfileArgAttrName(),
+          i, tensorrt::TensorRTDialect::getShapeTensorValueBoundsArgAttrName(),
           *boundsAttr);
-      continue;
+      func->setResultAttr(i, mlir::getHostTensorArgAttrName(),
+                          rewriter.getUnitAttr());
     }
-    assert(srcAttr.isValueBound() && "expected value bound or shape bound");
-    func->setResultAttr(
-        i, tensorrt::TensorRTDialect::getShapeTensorValueBoundsArgAttrName(),
-        *boundsAttr);
-    func->setResultAttr(i, mlir::getHostTensorArgAttrName(),
-                        rewriter.getUnitAttr());
   }
+  return success();
+}
+
+template <typename OpType>
+static LogicalResult outlineTensorRTRegion(RewriterBase &rewriter, OpType op) {
+  tensorrt::TensorRTModuleOp trtModuleOp = getOrCreateTensorRTModuleOp(op);
+  auto funcArgTypes = llvm::to_vector(TypeRange(op.getInputs()));
+
+  FailureOr<FunctionOpInterface> func = createOutlinedFunc(
+      rewriter, op.getLoc(), op, trtModuleOp, "tensorrt_cluster",
+      "cluster.tensorrt", TypeRange(op.getInputs()),
+      op.getYield()->getOperandTypes());
+
+  if (failed(func))
+    return failure();
+
+  assert(func->getFunctionBody().getBlocks().size() == 1 &&
+         "expected body with one block");
+  func->setPublic();
+
+  rewriter.setInsertionPoint(op);
+  auto callOp = createCallOp<OpType>(rewriter, op, trtModuleOp, *func);
+
+  if (failed(populateFunctionAttributes(rewriter, op, &(*func))))
+    return failure();
 
   // Populate the function entry block.
   rewriter.eraseBlock(&func->getFunctionBody().front());
@@ -234,14 +265,14 @@ static LogicalResult outlineTensorRTRegion(RewriterBase &rewriter,
   // ops to the `tensorrt.module` op. This is needed since `tensorrt.module` op
   // has its own symbol table.
   SymbolTableCollection symbolTable;
-  for (auto compositeOp : op.getBody().getOps<stablehlo::CompositeOp>()) {
+  for (auto compositeOp :
+       op.getBody().template getOps<stablehlo::CompositeOp>()) {
     auto decompositionFunc = dyn_cast_if_present<func::FuncOp>(
-        symbolTable.lookupSymbolIn(op->getParentOfType<ModuleOp>(),
+        symbolTable.lookupSymbolIn(op->template getParentOfType<ModuleOp>(),
                                    compositeOp.getDecompositionAttr()));
     if (!decompositionFunc)
       return emitError(compositeOp.getLoc())
-             << "failed to lookup stablehlo.composite decomposition "
-                "function: "
+             << "failed to lookup stablehlo.composite decomposition function: "
              << compositeOp.getDecompositionAttr();
     rewriter.moveOpAfter(decompositionFunc, func->getOperation());
   }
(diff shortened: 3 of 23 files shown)


mlir-tensorrt/executor/include/mlir-executor-c/Runtime/Runtime.h
```diff
@@ -374,16 +385,25 @@ static inline bool mtrtRuntimeSessionIsNull(MTRT_RuntimeSession session) {
...
 /// A stream may optionally be specified, otherwise pass the result of
 /// `mtrtStreamGetNull()`.
+///
+/// The `results` array must point to an array with at least the number of
+/// elements returned by mtrtRuntimeSessionGetNumResults for the given function.
 MLIR_CAPI_EXPORTED MTRT_Status mtrtRuntimeSessionExecuteFunction(
     MTRT_RuntimeSession session, MTRT_StringView name,
     const MTRT_RuntimeValue *inArgs, size_t numInArgs,
-    const MTRT_RuntimeValue *outArgs, size_t numOutArgs, MTRT_Stream stream);
+    const MTRT_RuntimeValue *outArgs, size_t numOutArgs,
```
jhalakpatel, line 402: As per @christopherbate in the other PR: It might be better to expose an API entirely for the function signature.

mlir-tensorrt/compiler/lib/Conversion/TensorRTToTensorRTRuntime/TensorRTToTensorRTRuntime.cpp
```diff
@@ -93,6 +93,8 @@ convertCallOp(Operation *op, IRRewriter &rewriter,
   SmallVector<int64_t> hostTensorArgs;
   for (auto [idx, arg] : llvm::enumerate(trtFunc.getArguments())) {
     const TensorKindLattice *kind = solver.lookupState<TensorKindLattice>(arg);
+    if (!isa<RankedTensorType>(arg.getType()))
```
jhalakpatel, line 96: This is required since function argument could be a stream or context.

mlir-tensorrt/executor/include/mlir-executor-c/Runtime/Runtime.h
jhalakpatel: I would like to defer this if this is not a blocker.

mlir-tensorrt/executor/include/mlir-executor/Runtime/Backend/Lua/LuaRuntime.h
```diff
@@ -104,13 +104,6 @@ executeFunctionWithLuaBackend(LuaRuntimeSession &session, std::string_view name,
                               std::optional<CudaStream> stream = {},
                               std::optional<RuntimeClient *> client = {});
 
-// Parses the results of a function call, handling both scalar and MemRef return
-// types
-StatusOr<llvm::SmallVector<std::unique_ptr<RuntimeValue>>>
```
jhalakpatel, line 109: This is not required to be present in the header since we do not access this function directly during testing.

mlir-tensorrt/python/bindings/Runtime/RuntimePyBind.cpp
```diff
@@ -950,22 +959,43 @@ PYBIND11_MODULE(_api, m) {
...
-                stream ? *stream : mtrtStreamGetNull());
+                resultsGeneric.data(), stream ? *stream : mtrtStreamGetNull(),
+                client);
             THROW_IF_MTRT_ERROR(s);
-          },
-          py::arg("name"), py::arg("in_args"), py::arg("out_args"),
-          py::arg("stream") = py::none());
 
+            std::vector<py::object> resultPyObject;
+            if (numResults > 0) {
```
jhalakpatel, line 987: Return a list of py::object instead of a single py::object when numResults == 1 to simplify call sites.

mlir-tensorrt/python/bindings/Runtime/RuntimePyBind.cpp
```diff
@@ -950,22 +959,43 @@ PYBIND11_MODULE(_api, m) {
       .def(
           "execute_function",
           [](PyRuntimeSession &self, std::string name,
-             std::vector<py::object> inArgs, std::vector<py::object> outArgs,
-             std::optional<MTRT_Stream> stream) {
+             std::vector<py::object> inArgs,
+             std::optional<std::vector<py::object>> outArgs,
+             std::optional<MTRT_Stream> stream, PyRuntimeClient &client) {
```
jhalakpatel, line 964: I have tried making `PyRuntimeClient` optional but for some reason I am getting argument mismatch error while calling the function.
```
TypeError: execute_function(): incompatible function arguments. The following argument types are supported:
    1. (self: mlir_tensorrt.runtime._mlir_libs._api.RuntimeSession, name: str, in_args: List[object], out_args: Optional[List[object]] = None, stream: Optional[MTRT_Stream] = None, client: Optional[MTRT_RuntimeClient] = None) -> List[object]

Invoked with: <mlir_tensorrt.runtime._mlir_libs._api.RuntimeSession object at 0x7f86f29f88f0>, 'main'; kwargs: in_args=[<mlir_tensorrt.runtime._mlir_libs._api.MemRefValue object at 0x7f86f29e9f70>], stream=<mlir_tensorrt.runtime._mlir_libs._api.Stream object at 0x7f86f2c682b0>, client=<mlir_tensorrt.runtime._mlir_libs._api.RuntimeClient object at 0x7f86f2c68270>
```

mlir-tensorrt/executor/lib/Conversion/MemRefToExecutor.cpp
```diff
@@ -548,6 +549,21 @@ void executor::populateMemRefToExecutorPatterns(
 }
 
 namespace {
+
+class RemoveNoOpClonePattern : public OpRewritePattern<bufferization::CloneOp> {
+public:
```
christopherbate, line 554: This is still illegal, we can't keep this here.

mlir-tensorrt/executor/test/lib/BufferizationTestPass.cpp
```diff
@@ -53,22 +53,32 @@ class ExecutorBufferizationTestPass
     }
   }
 };
+
+struct PlanBufferizationPipelineCliOpts
```
christopherbate, line 57: Plan bufferization is not part of the `executor` project. It looks like all the changes here can be removed since `enableNonDPSReturns` is not actually used anywhere here.

mlir-tensorrt/python/bindings/Runtime/RuntimePyBind.cpp
christopherbate: did you add `= py::none()` as the default value of `py::arg("client")`

jhalakpatel: Yes. I did.

mlir-tensorrt/tensorrt/lib/Target/TensorRTEncodingOpInterface/NetworkEncoder.cpp
```diff
@@ -566,11 +566,12 @@ static void serializeSplatElements(DenseIntOrFPElementsAttr values,
     packed |= ((value & 0x0F) << 4);
     // Fill `data` vector with `packed`
     std::fill_n(reinterpret_cast<uint8_t *>(data.data()), data.size(), packed);
-    return;
+    return llvm::success();
   }
-
-  llvm_unreachable("unsupported data type to convert MLIR splat attribute to "
```
christopherbate, line 572: This makes no sense. `llvm_reachable` will cause an abort, and program will never proceed to `llvm::errs`. just do 

```
return emitError(UnknownLoc:get(values.getContext()) << "my error message";
```

mlir-tensorrt/executor/lib/Conversion/MemRefToExecutor.cpp
jhalakpatel: Agreed. Please let me know how this can be better implemented.

mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/EliminateShapeOps.cpp
```diff
@@ -99,8 +108,17 @@ static LogicalResult removeUnusedArgs(SymbolTableCollection &collection,
...
     // Update the call ops.
-    for (tensorrt::CallOp callOp : callOps)
-      callOp.getInputsMutable().erase(i);
+    for (Operation *callOp : callOps) {
+      if (auto call = dyn_cast<tensorrt::CallOp>(callOp))
+        call.getInputsMutable().erase(i);
+      else if (auto callAlloc = dyn_cast<tensorrt::CallAllocOp>(callOp))
+        callAlloc.getInputsMutable().erase(i);
+      else {
+        llvm::errs() << "Unexpected operation type in callOps\n";
```
christopherbate, line 117: This is not how we handle errors. Use

```
return emitError(loc) << "my error message";
```

mlir-tensorrt/tensorrt/lib/Target/TensorRTEncodingOpInterface/NetworkEncoder.cpp
jhalakpatel: Yes, I replaced `llvm_unreachable` with `llvm::errs` and `llvm::failure`. I will use `emitError` instead.

FYI, several other places could benefit from such a change in `NetworkEncoder.cpp`. I can this as a follow-up PR.

mlir-tensorrt/python/bindings/Runtime/RuntimePyBind.cpp
jhalakpatel: It looks like using `PyRuntimeClient*` works instead of using `std::optional<PyRuntimeClient>`.

mlir-tensorrt/executor/lib/Conversion/MemRefToExecutor.cpp
jhalakpatel: @christopherbate Please re-review as I have addressed your comment.

mlir-tensorrt/compiler/lib/Conversion/TensorRTRuntimeToExecutor/TensorRTRuntimeToExecutor.cpp
```diff
@@ -379,12 +379,9 @@ struct ConvertEnqueueAllocToCall
 
     // Create output memrefs from output descriptors
     SmallVector<Value> results;
+    unsigned offset = 1; // Skip num of results
```
christopherbate, line 382: Can you give a better description of what this is for? Is this variable even used? should it be `const`?

mlir-tensorrt/compiler/lib/Conversion/TensorRTToTensorRTRuntime/TensorRTToTensorRTRuntime.cpp
christopherbate: A `func.func` that is the callee of `tensorrt.call|call_alloc` should only have tensor arguments. You have observed otherwise?

mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/EliminateShapeOps.cpp
```diff
@@ -65,17 +65,26 @@ struct RemoveWithValuesRewriter : public OpRewritePattern<plan::WithValuesOp> {
...
-    }
-    map.insert(std::make_pair(func, SmallVector<tensorrt::CallOp>{callOp}));
+
+    func::FuncOp func;
+    if (auto call = dyn_cast<tensorrt::CallOp>(callOp))
+      func = call.getFuncCallee(collection);
+    else if (auto callAlloc = dyn_cast<tensorrt::CallAllocOp>(callOp))
+      func = callAlloc.getFuncCallee(collection);
+    else
+      return;
```
christopherbate, line 82: Should be 

```
if(...) {
  func = 
} else {
  auto callAlloc = cast<tensorrrt::CallOp>(callOp)
  func =  ...
}
```

you already enforced the required invariant above at the start of the walk.

mlir-tensorrt/compiler/lib/Conversion/TensorRTRuntimeToExecutor/TensorRTRuntimeToExecutor.cpp
jhalakpatel: Sure. I can rename to `outputDesOffset`. It basically gets incremented in the code below. I found this issue while trying to debug a Tripy test where function has multiple returns values with different rank.

mlir-tensorrt/compiler/lib/Conversion/TensorRTToTensorRTRuntime/TensorRTToTensorRTRuntime.cpp
jhalakpatel: Actually. Weirdly, we do not require this check anymore. This was from my initial PR, so it looks like this is not required anymore.

mlir-tensorrt/test/Conversion/TensorRTRuntimeToExecutor/tensorrt-runtime-to-executor.mlir
```diff
@@ -133,4 +133,93 @@ func.func @main(%arg0: memref<1x3x256x256xf32, #executor.memory_type<device>>) -
...
+    return
+  }
+}
+
+func.func @main(%arg0: memref<?x?x?xf32, #executor.memory_type<device>> {tensorrt.shape_profile = #executor.dim_bounds<min = [1, 1, 2], max = [1, 4, 2]>}, %arg1: memref<?x?xi32, #executor.memory_type<device>> {tensorrt.shape_profile = #executor.dim_bounds<min = [1, 1], max = [1, 4]>}) -> (memref<?x?x256xf32, #executor.memory_type<device>> {tensorrt.shape_profile = #executor.dim_bounds<min = [1, 2, 256], max = [1, 5, 256]>}, memref<?x256x64x64xf32, #executor.memory_type<device>> {tensorrt.shape_profile = #executor.dim_bounds<min = [1, 256, 64, 64], max = [1, 256, 64, 64]>}) attributes {executor.function_metadata = #executor.func_meta<[memref<?x?x?xf32, #executor.memory_type<device>> {#executor.dim_bounds<min = [1, 1, 2], max = [1, 4, 2]>}, memref<?x?xi32, #executor.memory_type<device>> {#executor.dim_bounds<min = [1, 1], max = [1, 4]>}], [memref<?x?x256xf32, #executor.memory_type<device>> {#executor.dim_bounds<min = [1, 2, 256], max = [1, 5, 256]>}, memref<?x256x64x64xf32, #executor.memory_type<device>> {#executor.dim_bounds<min = [1, 256, 64, 64], max = [1, 256, 64, 64]>}], num_output_args = 0>} {
```
christopherbate, line 146: Tests should strive to contain` the *minimum* amount of IR/code to exercise the pass we are testing here. 

So all the `tensorrt.shape_profile`, etc should be removed. Think about it this way: If we update or change the IR textual assembly format for `#executor.dim_bounds`, we would need to update this test, but that component is not even needed here.

Similarly, the `bufferization.clone` should be removed.

Since this particular test does not need to test conversion of `trtrt.compile` or `cuda.get_global_stream`, you could just add additional arguments `%argX: !trtrt.context` and `%argY: !cuda.stream` to further simplify the test and the amount of `CHECK` statements that need to be written below.

mlir-tensorrt/test/Conversion/TensorRTRuntimeToExecutor/tensorrt-runtime-to-executor.mlir
```diff
@@ -133,4 +133,93 @@ func.func @main(%arg0: memref<1x3x256x256xf32, #executor.memory_type<device>>) -
...
+// CHECK-DAG: %[[c0:.*]] = executor.constant 0 : i64
+// CHECK-DAG: %[[c1:.*]] = executor.constant 1 : i64
+// CHECK: %[[v6:.*]] = builtin.unrealized_conversion_cast %[[arg0]] : memref<?x?x?xf32, #executor.memory_type<device>> to !executor.table<!executor.ptr<device>, !executor.ptr<device>, i64, i64, i64, i64, i64, i64, i64>
+// CHECK: %[[v7:.*]] = builtin.unrealized_conversion_cast %[[arg1]] : memref<?x?xi32, #executor.memory_type<device>> to !executor.table<!executor.ptr<device>, !executor.ptr<device>, i64, i64, i64, i64, i64>
+// CHECK: %[[v8:.*]] = executor.get_global @my_func_exec_ctx : !executor.opaque<"trtrt_context">
+// CHECK: %[[v9:.*]] = cuda.get_global_stream 0
+// CHECK: %[[v10:.*]] = builtin.unrealized_conversion_cast %[[v9]] : !cuda.stream to !executor.ptr<host>
+// CHECK: %[[v11:.*]] = executor.alloca %[[c1]] x !executor.table<i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64> : (i64) -> !executor.ptr<host>
+// CHECK: %[[v12:.*]] = executor.getoffset[0, 0] : () -> i64, !executor.table<i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64>
```
christopherbate, line 189: Nit: for these sorts of CHECK statements in very verbose IR, I tend to try to knock off as much extraneous detail as possible that doesn't really add anything to the test quality. Try to do a simple regex search/replace that eliminates everything after the `:` in all these check statements. That removes a lot of textual detail that isn't really important to check here and makes the test easier to maintain.

mlir-tensorrt/python/bindings/Runtime/RuntimePyBind.cpp
```diff
@@ -950,22 +959,45 @@ PYBIND11_MODULE(_api, m) {
...
+                client ? MTRT_RuntimeClient(*client)
+                       : mtrtRuntimeClientGetNull());
             THROW_IF_MTRT_ERROR(s);
-          },
-          py::arg("name"), py::arg("in_args"), py::arg("out_args"),
-          py::arg("stream") = py::none());
 
+            std::vector<py::object> resultPyObject;
+            if (numResults > 0) {
+              for (const auto &arg : resultsGeneric)
```
christopherbate, line 990: hmm, good idea

mlir-tensorrt/python/bindings/Runtime/RuntimePyBind.cpp
```diff
@@ -600,6 +600,15 @@ static MTRT_RuntimeValue convertArgType(py::object obj) {
   throw std::runtime_error("argument must be MemRef or scalar");
 }
 
+/// Convert Runtime value to PyMemRefValue or PyScalarValue object.
+static py::object convertGenericArgToPyObject(MTRT_RuntimeValue value) {
+  if (mtrtRuntimeValueIsMemRef(value))
+    return py::cast<PyMemRefValue>(mtrtRuntimeValueDynCastToMemRef(value));
+  if (mtrtRuntimeValueIsScalar(value))
+    return py::cast<PyScalarValue>(mtrtRuntimeValueDynCastToScalar(value));
+  return py::none();
```
christopherbate, line 609: Seems like we should throw an exception here?

mlir-tensorrt/executor/lib/Runtime/API/API.cpp
```diff
@@ -725,9 +731,16 @@ StatusOr<std::unique_ptr<MemRefValue>> MemRefValue::create(
   if (!::getFootprintInBytes(shape, strides, bitsPerElement).isOk())
     return getInvalidArgStatus(
         "only memrefs with non-negative strides are allowed");
-  if (!ptr)
+
+  auto is_empty_tensor = [](const llvm::ArrayRef<int64_t> &shape) -> bool {
```
christopherbate, line 735: Pass `llvm::ArrayRef` by value, not by reference. `const` also doesn't add anything of value when passing ArrayRef, so the convention that's always used is to pass a plain non-const `ArrayRef<T>` by value.

mlir-tensorrt/executor/lib/Runtime/API/API.cpp
```diff
@@ -367,13 +367,14 @@ RuntimeSession::RuntimeSession(RuntimeSessionOptions options,
...
 AllocTracker::~AllocTracker() {
+  MTRT_DBGF("Destroying alloc tracker %p", static_cast<void *>(this));
   MTRT_DBGF("checking %u allocations", map.size());
   llvm::SmallVector<PointerInfo> ptrsToFree;
   ptrsToFree.reserve(map.size());
   for (const auto &[ptrVal, metadata] : map) {
     if (metadata->info.isInternallyManaged() &&
         metadata->externalReferenceCount.load() == 0) {
-      MTRT_DBGF("still live: 0x%lx type %d size %lu", ptrVal,
+      MTRT_DBGF("still live: 0x%lx type %d size %lx", ptrVal,
```
christopherbate, line 377: Why do you want to print the size here in hex? Usually we print sizes as `%lu`

mlir-tensorrt/executor/lib/Conversion/MemRefToExecutor.cpp
```diff
@@ -24,6 +24,7 @@
 #include "mlir-executor/Conversion/ConvertToExecutorCommon.h"
 #include "mlir-executor/Conversion/Passes.h"
 #include "mlir-executor/Executor/IR/Executor.h"
+#include "mlir/Dialect/Bufferization/IR/Bufferization.h"
```
christopherbate, line 27: Nit: remove unnecessary inclusion

https://llvm.org/docs/CodingStandards.html#include-as-little-as-possible

mlir-tensorrt/executor/lib/CAPI/Runtime/Runtime.cpp
```diff
@@ -731,19 +742,36 @@ MTRT_Status mtrtRuntimeSessionExecuteFunction(
...
 
   return mtrtStatusGetOk();
 }
+
+MTRT_Status mtrtRuntimeSessionGetNumResults(MTRT_RuntimeSession session,
+                                            MTRT_StringView name,
+                                            int64_t *numResults) {
+  LuaRuntimeSession *cppSession =
+      static_cast<LuaRuntimeSession *>(unwrap(session));
+  *numResults = cppSession->getExecutable()
```
christopherbate, line 768: Probably should check that the function exists and return an error if it doesn't. Otherwise, we will crash here.

mlir-tensorrt/executor/include/mlir-executor-c/Runtime/Runtime.h
```diff
@@ -391,16 +402,25 @@ static inline bool mtrtRuntimeSessionIsNull(MTRT_RuntimeSession session) {
...
+    MTRT_RuntimeValue *results, MTRT_Stream stream, MTRT_RuntimeClient client);
+
+MLIR_CAPI_EXPORTED MTRT_Status mtrtRuntimeSessionGetNumResults(
```
christopherbate, line 422: Nit:  add documentation comment, e.g. explain that `name` refers to the name of an exported function in the executable

mlir-tensorrt/executor/include/mlir-executor-c/Runtime/Runtime.h
```diff
@@ -53,7 +53,7 @@ extern "C" {
 /// caller must be sure to delete errors via mtrtStatusDestroy.
 //===----------------------------------------------------------------------===//
 
-typedef struct MTRT_RuntimeClient MTRT_Runtimeclient;
+struct MTRT_RuntimeClient; // Forward declaration
```
christopherbate, line 56: This is a C header, not C++. 

Without the typedef, we would fail to compile when including this header in a C library (e.g. we would need to rename all the types below to `struct MTRT_RuntimeClient` instead of just `MTRT_RuntimeClient`. The `typedef` is a requirement.

mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/OutlineClusters.cpp
```diff
@@ -268,8 +268,82 @@ static LogicalResult outlineTensorRTRegion(RewriterBase &rewriter,
...
+        i, tensorrt::TensorRTDialect::getShapeTensorValueBoundsArgAttrName(),
+        *boundAttr);
+    func->setArgAttr(i, mlir::getHostTensorArgAttrName(),
+                     rewriter.getUnitAttr());
+  }
+
+  // Populate the function entry block.
+  rewriter.eraseBlock(&func->getFunctionBody().front());
+
+  // Move private decomposition funcs associated with all `stablehlo.composite`
```
christopherbate, line 320: There's a lot of code duplication here from the non-DPS function above (e.g. lines 323-334 and 291-315). Those can be outlined to helper functions to share the code.

mlir-tensorrt/executor/include/mlir-executor-c/Runtime/Runtime.h
```diff
@@ -215,6 +215,11 @@ static inline bool mtrtRuntimeClientIsNull(MTRT_RuntimeClient client) {
   return !client.ptr;
 }
 
+/// Returns null client.
+static inline MTRT_RuntimeClient mtrtRuntimeClientGetNull() {
+  return MTRT_RuntimeClient{nullptr};
```
christopherbate, line 220: Use C `NULL`, not `nullptr`, for same reasons as above. It looks like we have a few other uses of `nullptr` in the file unrelated to your change which we should correct as well. You can leave them as-is or update them, up to you.

mlir-tensorrt/test/python/IntegrationTests/test_non_dps_cconv.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+def test_empty_shape_tensor():
+    exe = compile_executable(empty_shape_tensor)
+    session_options = runtime.RuntimeSessionOptions(num_devices=1, device_id=0)
+    session = runtime.RuntimeSession(session_options, exe)
+    results = session.execute_function("main", in_args=[], stream=stream, client=client)
+    stream.sync()
+    print(np.asarray(results[0]))
+
```
christopherbate, line 204: Would be good to get some test coverage for scalar results and mixed memref/scalar results as well.

christopherbate:
Overall, looks good except for minor comments. Great to see it working end-to-end, and of course much kudos to you for getting this working!

mlir-tensorrt/tensorrt/lib/Target/TensorRTEncodingOpInterface/NetworkEncoder.cpp
christopherbate: Ah, ok, I didn't realize when reviewing that these changes were disconnected from the rest of the MR. We do have an open issue internally for this. Thanks for taking care of it!

mlir-tensorrt/test/Conversion/TensorRTRuntimeToExecutor/tensorrt-runtime-to-executor.mlir
jhalakpatel: Excellent point. I have simplified the IR now.

mlir-tensorrt/test/Conversion/TensorRTRuntimeToExecutor/tensorrt-runtime-to-executor.mlir
jhalakpatel: Make sense.

mlir-tensorrt/executor/lib/Runtime/API/API.cpp
jhalakpatel: That was an oversight. Fixed it.

mlir-tensorrt/executor/lib/CAPI/Runtime/Runtime.cpp
jhalakpatel: Good point. Let me update the `getFunction` to return status and return here.

mlir-tensorrt/executor/include/mlir-executor-c/Runtime/Runtime.h
jhalakpatel: any idea why we use `MTRT_Runtimeclient` instead of `MTRT_RuntimeClient`? Is it just a typo?

mlir-tensorrt/executor/include/mlir-executor-c/Runtime/Runtime.h
jhalakpatel: Sure. I will fix other sites as well.

mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/OutlineClusters.cpp
jhalakpatel: Sure. Let me use template to reduce deduplicate code.

mlir-tensorrt/test/python/IntegrationTests/test_non_dps_cconv.py
jhalakpatel: I have added tests to cover the scalar and mixed results.

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/LuaRuntime.cpp
```diff
@@ -650,9 +650,6 @@ parseResults(const sol::protected_function_result &pfr,
     if (!memref.isOk())
       return memref.getStatus();
 
-    // Increment external reference count since we are returning a memref
-    allocator.incrementExternalCount(info.ptr);
```
jhalakpatel: We do not need this. Reference count should only be incremented if a explicitly create a DL pack tensor on a memref. 
Here memref is internally owned and tracked.

jhalakpatel:
@christopherbate were you able to merge changes from this PR?

christopherbate:
Sorry, fell behind while you were gone and this wasn't blocking TriPy team. I will merge this week.

mlir-tensorrt/compiler/lib/Conversion/TensorRTRuntimeToExecutor/TensorRTRuntimeToExecutor.cpp
```diff
@@ -379,20 +379,22 @@ struct ConvertEnqueueAllocToCall
 
     // Create output memrefs from output descriptors
     SmallVector<Value> results;
+    // Initialize output descriptor offset to skip number of results.
```
christopherbate, line 382: these changes are merged

mlir-tensorrt/test/Conversion/TensorRTRuntimeToExecutor/tensorrt-runtime-to-executor.mlir
```diff
@@ -102,35 +102,93 @@ func.func @main(%arg0: memref<1x3x256x256xf32, #executor.memory_type<device>>) -
 // CHECK: %[[v8:.*]] = cuda.stream.create : !cuda.stream
 // CHECK: %[[v9:.*]] = builtin.unrealized_conversion_cast %[[v8]] : !cuda.stream to !executor.ptr<host>
 // CHECK: %[[v10:.*]] = executor.alloca %[[c1]] x !executor.table<i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64> : (i64) -> !executor.ptr<host>
-// CHECK: %[[v11:.*]] = executor.getoffset[0, 0] : () -> i64, !executor.table<i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64>
-// CHECK: executor.store %[[c1]] to %[[v10]] + %[[v11]] : i64, !executor.ptr<host>, i64
-// CHECK: %[[v12:.*]] = executor.getoffset[0, 1] : () -> i64, !executor.table<i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64>
-// CHECK: executor.store %[[c4]] to %[[v10]] + %[[v12]] : i64, !executor.ptr<host>, i64
+// CHECK: %[[v11:.*]] = executor.getoffset[0, 0]
+// CHECK: executor.store %[[c1]] to %[[v10]] + %[[v11]]
```
christopherbate, line 106: changes in this file are merged

mlir-tensorrt/compiler/lib/Dialect/Plan/Transforms/EliminateShapeOps.cpp
```diff
@@ -65,17 +65,26 @@ struct RemoveWithValuesRewriter : public OpRewritePattern<plan::WithValuesOp> {
 } // namespace
```
christopherbate, line 65: Changes here were merged  with some additional tests

jhalakpatel:
Closing as opened a new PR: https://github.com/NVIDIA/TensorRT-Incubator/pull/488. I lost permissions to directly commit to the project so I am moving to fork based development for now.

PR #290, opened 2024-10-19
[Python/test] Remove torch dependency from MLIR-TensorRT

jhalakpatel:


jhalakpatel:
@shelkesagar29 Let me know if `torch` is required to test unit stride support? 
`#       CHECK: Memref stride: [1, 4, 1]` stride from torch is basically incorrect as frameworks do not guarantee correct stride for 1 or 0 dim. I think numpy test is sufficient to test unit stride in this case.

shelkesagar29:
> @shelkesagar29 Let me know if `torch` is required to test unit stride support? `# CHECK: Memref stride: [1, 4, 1]` stride from torch is basically incorrect as frameworks do not guarantee correct stride for 1 or 0 dim. I think numpy test is sufficient to test unit stride in this case.

This test was added by Akhil. Can you ask him what was the context behind using torch and not Numpy?

shelkesagar29:
LGTM except above comment. Feel free to merge it after discussion with Akhil.

akhilg-nv:
This test validates two things:
1. We preserve the stride information for a tensor coming from another framework
2. The compiler check allows a tensor with non-canonical (contiguous, row-major) stride for unit dimension.

As seen by the change, using numpy will result in the stride information being preserved "correctly." Torch is used to test this specific case where the first dimension's stride will be 1 instead of 12. 

Perhaps an alternative to remove torch dependency would be to validate the stride from numpy, and then manually set the first dimension stride to 1 to verify that this still works. But I'm not sure if torch is used elsewhere - when adding this test, Chris mentioned that I should be able to use torch without any additional set-up.

jhalakpatel:
Closing this MR since the resolution is to update the container with torch installation.

PR #291, opened 2024-10-21
[Python/Runtime] Add bindings to enable llvm global debug flag from runtime api

jhalakpatel:
Allow enabling global llvm::DebugFlag and setting current debug types from runtime python bindings.

Usage:
```
runtime.GlobalDebug.flag = True
runtime.GlobalDebug.set_types(["runtime", "allocator"])
```

Runtime debug dumps:
```
/workspaces/TensorRT-Incubator/mlir-tensorrt/executor/lib/Runtime/API/API.cpp:457 AllocTracker is now tracking 0x5b6000200 size=96 space=device ownership=internal
[runtime] copying 0x00005C9147A03470 to 0x00000005B6000200 size=96 bytes asynchronously via pinned staging buffer at 0x00000005B5000000
/workspaces/TensorRT-Incubator/mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp:133 freeing execution context at 101779074191656
```

### DIFF
--- a/mlir-tensorrt/executor/include/mlir-executor-c/Runtime/Runtime.h
+++ b/mlir-tensorrt/executor/include/mlir-executor-c/Runtime/Runtime.h
@@ -56,6 +56,22 @@ extern "C" {
 typedef struct MTRT_RuntimeClient MTRT_Runtimeclient;
 
 //===----------------------------------------------------------------------===//
+// MTRT_GlobalDebug
+//===----------------------------------------------------------------------===//
+
+/// Enables or disables global debug flag.
+MTRT_Status mtrtEnableGlobalDebug(bool enable);
+
+/// Retrieves the current state of the global debug flag.
+MTRT_Status mtrtIsGlobalDebugEnabled(bool *enable);
+
+/// Sets the current global debug type.
+MTRT_Status mtrtSetGlobalDebugType(const char *type);
+
+/// Sets multiple global debug types.
+MTRT_Status mtrtSetGlobalDebugTypes(const char **types, size_t n);
+
+//===----------------------------------------------------------------------===//
 // MTRT_Stream
 //===----------------------------------------------------------------------===//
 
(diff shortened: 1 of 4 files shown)


mlir-tensorrt/executor/lib/CAPI/Runtime/Runtime.cpp
```diff
@@ -146,6 +147,48 @@ static MTRT_Status wrap(const Status &status) {
 // MTRT_Stream
 //===----------------------------------------------------------------------===//
 
+static void mlirEnableGlobalDebug(bool enable) { llvm::DebugFlag = enable; }
```
jhalakpatel, line 150: @christopherbate I have copied the implementation from https://github.com/llvm/llvm-project/blob/main/mlir/lib/CAPI/Debug/Debug.cpp#L18. I tested it but for some reason it, it is still not dumping anything.
This is how I am using the API: https://github.com/NVIDIA/TensorRT-Incubator/pull/291/files#diff-765df7b86eaf815fdd1a6331a3d7e7c177214cc01aa96fa8040c87850d403806R77

mlir-tensorrt/executor/include/mlir-executor-c/Runtime/Runtime.h
```diff
@@ -55,6 +55,18 @@ extern "C" {
 
 typedef struct MTRT_RuntimeClient MTRT_Runtimeclient;
 
+//===----------------------------------------------------------------------===//
+// MTRT_GlobalDebug
+//===----------------------------------------------------------------------===//
+
+MLIR_CAPI_EXPORTED MTRT_Status mtrtEnableGlobalDebug(bool enable);
```
christopherbate, line 62: Please add documentation comments

mlir-tensorrt/executor/lib/CAPI/Runtime/Runtime.cpp
christopherbate: That makes sense.... However there's no point adding these functions `mlirXYZ` here. That's very confusing. Just  inline al these static functions you have defined.

christopherbate (approved):
Added a few comments otherwise LGTM

PR #292, opened 2024-10-21
Avoids passing stream when creating host memrefs to not trip an assertion

pranavm-nvidia:
MLIR-TRT includes an assertion that disallows a stream from being provided when creating a host memref. This change adds a new `create_memref` helper which will avoid passing the stream in such cases. This also unifies all the call-sites to use this new helper.

### DIFF
--- a/tripy/tripy/backend/mlir/executor.py
+++ b/tripy/tripy/backend/mlir/executor.py
@@ -20,7 +20,7 @@ from typing import List
 import mlir_tensorrt.runtime.api as runtime
 
 from tripy.backend.api.stream import default_stream
-from tripy.backend.mlir.memref import create_empty_memref
+from tripy.backend.mlir.memref import create_memref
 from tripy.backend.mlir.utils import MLIRRuntimeClient, convert_runtime_dtype_to_tripy_dtype
 from tripy.backend.utils import TensorInfo
 from tripy.common import datatype, device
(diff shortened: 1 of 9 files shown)


tripy/tripy/backend/mlir/executor.py
```diff
@@ -134,12 +136,8 @@ def execute(self, output_devices: List[device], inputs: List["Tensor"] = []) ->
 
         # Allocate output memory and store buffer pointers.
         outputs = [
-            create_empty_memref(
-                shape=info.shape,
-                dtype=info.dtype,
-                device=info.device,
-                stream=self.stream._active_cuda_stream,
-                use_cache=False,
```
pranavm-nvidia, line 142: @jhalakpatel do we actually need to disable the cache here?

jhalakpatel: Yes, otherwise for two outputs with the same signature, it would use the same buffer. I ran into an issue with one of the tests where both the outputs would end up sharing the same buffer.

pranavm-nvidia: Isn't that fine for empty memrefs?

jhalakpatel: I think the use of `create_empty_memref` was confusing here since `info.shape` is not meant to be "empty" i.e. as a `0` in of the dim. Instead, it should have been an initialized memref.

I think for "empty" tensors it would be OK to drop the `use_cache`. May be earlier logic was not caching irrespective of whether the shape is empty or not, that is why I ran into the bug.

jhalakpatel (approved):

tripy/tripy/backend/mlir/executor.py
pranavm-nvidia: Makes sense. Right now we are only caching for empty tensors, so it should be fine to reuse the same memrefs.

yizhuoz004 (approved):

PR #299, opened 2024-10-24
Add ndim; support start_dim negative

parthchadha:


tripy/tripy/frontend/tensor.py
```diff
@@ -162,6 +162,10 @@ def dtype(self):
     def rank(self):
         return self.trace_tensor.rank
 
+    @property
+    def ndim(self):
```
pranavm-nvidia, line 166: Should we remove `rank` then?

parthchadha: We should keep both IMO.

PR #301, opened 2024-10-24
Replaces `convert_inputs_to_tensors` decorator with a better decorator

pranavm-nvidia:
- Replaces `convert_inputs_to_tensors` decorator with `convert_to_tensors`,
     which will convert any `TensorLike` arguments by default and sync data types
     according to any specified data type constraints. This means that information
     does not need to be duplicated in the decorator like it used to be.

- Disallows unsafe downcasting in the new `convert_to_tensors` decorator.
     Previously, you could have:
     ```py
     x = tp.Tensor([1], dtype=tp.int32)
     x * 3.99
     ```
     which would yield a result equal to:
     ```py
     tp.Tensor([3])
     ```
     due to truncation when casting.

- Fixes several type annotations for binary elementwise ops.

- Removes NestedNumberSequence
    We only want to convert Python numbers to Tensors automatically.
    It doesn't make sense to write, for example, `x + [3]`.

### DIFF
--- a/tripy/tests/frontend/test_shape.py
+++ b/tripy/tests/frontend/test_shape.py
@@ -511,7 +511,7 @@ class TestShape:
     def test_binary_elementwise_broadcast_rejected(self, values):
         with raises(
             tp.TripyException,
-            match=r"\_\_mul\_\_ expects tensor arguments to have matching class types, but got mixed `tp\.Tensor` and `tp\.Shape` arguments\.",
+            match=r"Error processing shape inputs in operator BinaryElementwise Further information: Binary elementwise operators do not accept combinations of Shape and Tensor arguments.",
         ):
             tp.Shape(values).multiply(tp.Tensor([values, values]))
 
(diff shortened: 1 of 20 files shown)


yizhuoz004 (approved):

parthchadha (approved):

tripy/tests/frontend/test_shape.py
```diff
@@ -511,7 +511,7 @@ def test_quantize_rejected(self, values):
     def test_binary_elementwise_broadcast_rejected(self, values):
         with raises(
             tp.TripyException,
-            match=r"\_\_mul\_\_ expects tensor arguments to have matching class types, but got mixed `tp\.Tensor` and `tp\.Shape` arguments\.",
+            match=r"Error processing shape inputs in operator BinaryElementwise Further information: Binary elementwise operators do not accept combinations of Shape and Tensor arguments.",
```
jhalakpatel, line 514:
```suggestion
            match=r"Error processing shape inputs in operator BinaryElementwise. Further information: Binary elementwise operators do not accept combinations of Shape and Tensor arguments.",
```

pranavm-nvidia: @jhalakpatel I'm going to replace the part of the code that emits this error as part of the shape refactoring. Will fix the error messages then.

PR #305, opened 2024-10-25
Fix compile when no args are provided

parthchadha:


tripy/tests/backend/api/conftest.py
```diff
@@ -28,6 +31,10 @@ def returns_nothing(a):
     return
 
 
+def accepts_nothing():
+    return Tensor([1])
```
pranavm-nvidia, line 35: Can we just inline this into the test if it's only used in one place?

tripy/tripy/utils/utils.py
```diff
@@ -436,16 +436,18 @@ def get_positional_arg_names(func, *args):
     # None instead.
     signature = inspect.signature(func)
     arg_names = []
+    varargs_name = None
```
pranavm-nvidia, line 439: Why not just initialize `name = None`?

pranavm-nvidia: Actually I guess it's clearer this way.

pranavm-nvidia (approved):

PR #306, opened 2024-10-25
Removes `Shape` tensors

pranavm-nvidia:
In Tripy, we want shapes of tensors to behave like lists. This makes it
easier and more natural to manipulate shapes. However, we previously achieved
this through a `Shape` abstraction that was actually a `Tensor` which behaved
like a list in some, but not all, ways, which made the API very confusing.

To solve this, we remove the `Shape` abstraction and instead represent
shapes of tensors as lists of `DimensionSize`s (which are similar to `ShapeScalar`s).
These lists can be manipulated freely and are then converted back to tensors
by the `convert_to_tensors` decorator, which takes over the functionality of
`convert_shape_inputs`. This also has the advantage that operations on shapes
are not materialized in the computation graph - only operations on the individual
dimensions are.

Also includes various miscellaneous fixes:
- Updates `expand` to disallow `-1`s in prepended dimensions.
- Updates `resize` to use function overloading to accept either output shape or scales.
- Simplifies logic in `flatten`, `repeat`, `unsqueeze`, `full`, and `pad`.
- Greatly simplifies logic in `reshape`.
- Disallows empty lists in `concatenate`, similar to what `stack` does.
- Moves `squeeze` into its own file.
- Fixes to `get_arg_candidate_column_offsets` to work with `skip_num_stack_entries`.
- Updates `Storage` op to copy list inputs so that changes made to a list after it is used to construct a tensor don't affect the created tensor.


tripy/tripy/frontend/trace/ops/base.py
```diff
@@ -73,80 +72,27 @@ def build(cls, inputs: List["Tensor"], *args, num_outputs=1, **kwargs) -> Union[
         of returning a list of output tensors.
         """
 
-        from tripy.frontend.shape import Shape, ShapeScalar
         from tripy.frontend.tensor import Tensor
+        from tripy.frontend.dimension_size import DimensionSize
+
+        # Operations that operate on only DimensionSize inputs will always yield a DimensionSize.
+        # For any mixed operations, DimensionSize must be casted up to Tensor.
+        TensorType = DimensionSize if all(isinstance(inp, DimensionSize) for inp in inputs) else Tensor
```
pranavm-nvidia, line 80: Still not completely sure of how to implement the logic to determine whether the output should be a DimensionSize or a Tensor. There are two parts to this:
1. What should the decorator convert inputs to? If it always makes Tensors, the operations will never see DimensionSizes and will therefore never be able to indicate the output type accurately. 
2. What type should the op generate?

tripy/tripy/frontend/trace/ops/reshape.py
```diff
@@ -15,40 +15,30 @@
...
 from tripy.frontend.trace.ops import utils as op_utils
 from tripy.frontend.trace.ops.base import BaseTraceOp
+from tripy.types import ShapeLike
 
 
 @dataclass(repr=False)
 class Reshape(BaseTraceOp):
 
     output_rank: int
-    output_len: Optional[int] = None  # only used to help with infer_len for a shape input
```
pranavm-nvidia, line 31: Need to bring this back for this PR and remove it in a separate PR.

tripy/tripy/frontend/utils.py
```diff
@@ -226,20 +168,49 @@ def add_arg(arg):
...
                 if name in targets and not isinstance(arg, Tensor):
+                    if name in shape_likes:
+                        # NOTE: We cannot use the reshape API here since it would lead to an infinite loop
+                        # when it tries to convert its shape argument via `convert_to_tensors`
+                        dims = [
+                            Reshape.build([dim, Tensor([1])], 1) if isinstance(dim, DimensionSize) else Tensor([dim])
+                            for dim in arg
+                        ]
+                        arg = concatenate(dims, dim=0)
+                        # TODO (pranavm): Remove this WAR for segfault:
```
pranavm-nvidia, line 179: Need to fix a segfault in MLIR-TRT before merging this PR.

tripy/tripy/frontend/trace/ops/reshape.py
pranavm-nvidia: Actually this is not needed since `Reshape` can never produce a Shape now.

tripy/tests/flat_ir/ops/test_iota.py
```diff
@@ -49,9 +49,11 @@ def test_mlir(self, flat_ir):
...
             module {
                 func.func @main() -> tensor<?x?xf32> {
-                    %c = stablehlo.constant dense<[2, 3]> : tensor<2xi32>
-                    %0 = stablehlo.dynamic_iota %c, dim = 0 : (tensor<2xi32>) -> tensor<?x?xf32>
-                    return %0 : tensor<?x?xf32>
+                    %c = stablehlo.constant dense<2> : tensor<1xi32>
+                    %c_0 = stablehlo.constant dense<3> : tensor<1xi32>
+                    %0 = stablehlo.concatenate %c, %c_0, dim = 0 : (tensor<1xi32>, tensor<1xi32>) -> tensor<2xi32>
+                    %1 = stablehlo.dynamic_iota %0, dim = 0 : (tensor<2xi32>) -> tensor<?x?xf32>
+                    return %1 : tensor<?x?xf32>
```
yizhuoz004, line 56: What is the reason for this change?

tripy/tripy/frontend/trace/ops/slice.py
```diff
@@ -42,46 +40,8 @@ def infer_rank(self):
     def infer_len(self):
         # Only infer if we have concrete values to use. Note that the result is only a Shape if these are *slices*,
         # not single indices, so a slice is the only case that needs to be considered
-        if self.shape_slice is not None:
```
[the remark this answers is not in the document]
pranavm-nvidia: Since shapes are just lists now, slicing shapes does not trigger this code path.

tripy/tripy/frontend/trace/ops/base.py
yizhuoz004: Shall we only support a few ops to return `DimensionSize`? E.g. only binary and unary elementwise. I did not see other use cases. To return `DimensionSize`, the op must follow quite strict rules, e.g. I/O rank must be 0, dtype must be int32, it may be easier to just white-list the ops and let them override some function to provide the type information.

yizhuoz004:
Since tensor.shape is just a list now, how do we support dynamic slicing? for example
```
d0 = a.shape[0] # d0 is a DimensionSize
b.shape[d0]
```

tripy/docs/post0_developer_guides/how-to-add-new-ops.md
```diff
@@ -200,11 +196,10 @@ import tripy.frontend.utils as frontend_utils
 # If we needed to provide any special autodoc options, we could use the `autodoc_options` parameter.
 @export.public_api(document_under="tensor_operations")
 
-# The `convert_shape_inputs` decorator converts the specified function arguments into `tripy.Shape`s,
-# which would allow for using Python numbers and sequences. The `convert_to_tensors` decorator more generally converts
-# function arguments into Tripy tensors and is also commonly used in the codebase.
-@frontend_utils.convert_shape_inputs(["shape"])
-def theta(shape: Tuple[int], dim: int = 0, dtype: datatype.dtype = datatype.float32) -> "tripy.Tensor":
+# The `convert_to_tensors` decorator converts automatically converts compatible arguments,
+# like `TensorLike` or `ShapeLike`s into tensors.
```
slyubomirsky, line 200:
```suggestion
# The `convert_to_tensors` decorator automatically converts compatible arguments,
# like `TensorLike` or `ShapeLike`s, into tensors.
```

pranavm-nvidia:
> Since tensor.shape is just a list now, how do we support dynamic slicing? for example
> 
> ```
> d0 = a.shape[0] # d0 is a DimensionSize
> b.shape[d0]
> ```


If we support `b.shape[d0]`, then it seems natural to support `b.shape[:d0]` also . This would break our frontend rank/len inference because we would have to know the value of the shape elements instead of just the number of elements. Practically speaking, I can't really imagine a case where you'd need to index a shape based on the value of another shape. I expect that you would be able to transform the shape value (i.e. `d0`) directly in such cases.

tripy/tests/flat_ir/ops/test_iota.py
slyubomirsky: Perhaps a consequence of having lists of dimensions rather than single shape tensors?

tripy/tripy/frontend/dimension_size.py
```diff
@@ -0,0 +1,73 @@
...
+                raise_error(
+                    f"Scalar shape tensors must be of rank 0, but input tensor is rank {data.rank}", details=[data]
+                )
+            if data.dtype is not None and data.dtype != int32:
+                raise_error(
+                    f"Scalar shape tensor must have int32 member, but input tensor has data type {data.dtype}",
+                    details=[data],
+                )
+
+            # the shape of data should correspond to the given rank
```
slyubomirsky, line 54:
```suggestion
```
I don't think the comment refers to anything now.

tripy/tripy/frontend/trace/ops/base.py
pranavm-nvidia: I agree. I'm stuck on the mechanism for whitelisting though. What we would need is for the frontend APIs to somehow have knowledge of the original input types before they are converted to tensors. With that in place, we could just have `convert_to_tensors` always convert to the `Tensor` base class - since `Tensor`/`DimensionSize` are just frontend constructs, we can actually consume `Tensor`s and produce `DimensionSize`. 

One approach is to attach the original type as a private field to the tensor produced by `convert_to_tensors`:
```py
@convert_to_tensors()
def func(a: TensorLike, b: TensorLike):
    return_type = Tensor
    if a._original_type is DimensionSize and b._original_type is DimensionSize:
        return_type = DimensionSize
```
But I'm wondering if there is a better/cleaner approach than this.

tripy/tests/flat_ir/ops/test_iota.py
jhalakpatel: I wonder if the same IR bloat will be present once we move to the tensorrt dialect.

tripy/tripy/frontend/trace/ops/base.py
slyubomirsky: Is there any reason not to convert to `DimensionSize` if the criteria are met?

tripy/tests/flat_ir/ops/test_iota.py
yizhuoz004: I did not understand the difference, the original one should just work.

yizhuoz004:
> > Since tensor.shape is just a list now, how do we support dynamic slicing? for example
> > ```
> > d0 = a.shape[0] # d0 is a DimensionSize
> > b.shape[d0]
> > ```
> 
> If we support `b.shape[d0]`, then it seems natural to support `b.shape[:d0]` also . This would break our frontend rank/len inference because we would have to know the value of the shape elements instead of just the number of elements. Practically speaking, I can't really imagine a case where you'd need to index a shape based on the value of another shape. I expect that you would be able to transform the shape value (i.e. `d0`) directly in such cases.

I agree. We still support dynamic slicing if the input is transformed into a tensor.

tripy/tests/test_function_registry.py
```diff
@@ -474,7 +474,7 @@ def func(n: Sequence[Union[int, float]]) -> int:
     "typ, expected",
     [
         (tp.types.TensorLike, "Union[tripy.Tensor, numbers.Number]"),
-        (tp.types.ShapeLike, "Union[tripy.Shape, Sequence[Union[int, tripy.ShapeScalar]]]"),
+        (tp.types.ShapeLike, "Sequence[Union[int, tripy.DimensionSize]]"),
```
jhalakpatel, line 477: shouldn't `tripy.DimensionsSize` be an instance of `TensorLike` since it inherits from the `Tensor` class?

tripy/tripy/frontend/trace/ops/slice.py
```diff
@@ -42,46 +40,8 @@ def infer_rank(self):
     def infer_len(self):
         # Only infer if we have concrete values to use. Note that the result is only a Shape if these are *slices*,
         # not single indices, so a slice is the only case that needs to be considered
```
slyubomirsky, line 42:
```suggestion
```
Since this gets rid of `Shape` as a class, the comment is no longer applicable either.

tripy/tests/flat_ir/ops/test_iota.py
pranavm-nvidia: Yes, it's a consequence of having lists of dimensions. `Shape([2, 3])` will materialize the two element tensor, whereas now we are doing a `concat(DimensionSize([2]), DimensionSize([3]))`.

tripy/tripy/frontend/utils.py
```diff
@@ -570,30 +386,33 @@ def topological_sort(ops: List[Union["BaseTraceOp", BaseFlatIROp]]) -> List[Unio
...
-                    ],
-                )
-            return new_dim
-
-        args, kwargs = utils.modify_arg("dim", process_dim, func, *args, **kwargs)
-        return func(*args, **kwargs)
-
-    return wrapper
+def process_dim():
+    def process_dim_impl(func):
```
slyubomirsky, line 390: Was there a specific reason to add this step?

tripy/tripy/frontend/trace/ops/reshape.py
```diff
@@ -58,26 +48,19 @@ def infer_rank(self):
         else:
             self.outputs[0].rank = self.output_rank
 
-    @frontend_utils.make_function
+    # @frontend_utils.make_function
```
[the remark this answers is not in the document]
pranavm-nvidia: I don't think so since it's just a single FlatIR op. Will remove it.

tripy/tests/test_function_registry.py
pranavm-nvidia: Yes, it is, since `TensorLike` includes `Tensor` (and therefore all children of `Tensor`).

tripy/tripy/frontend/dimension_size.py
```diff
@@ -0,0 +1,73 @@
...
+            super().__init__(data=None, dtype=int32, name=name, device=data.device)
+            # share the underlying data
+            self.trace_tensor = data.trace_tensor
+            self.stack_info = data.stack_info
+        else:
+            # NOTE: We do not use isinstance here because bool is a subclass of int.
```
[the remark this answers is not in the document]
pranavm-nvidia: If we did then you could have `DimensionSize`s in the frontend with boolean values. Then we have to decide what it would mean to do `reshape(x, [True, False, True, False])` :).

tripy/tripy/utils/ast.py
```diff
@@ -141,10 +137,6 @@ def index_into_expr(node: ast.expr, index: int) -> ast.expr:
             else:
                 arg_node = index_into_expr(node.slice, index)
 
-        # if the resulting arg node is a list or tuple literal, we might be able to narrow it down with a list index
-        if list_index is not None and isinstance(arg_node, (ast.List, ast.Tuple)) and list_index < len(arg_node.elts):
-            arg_node = arg_node.elts[list_index]
```
slyubomirsky, line 146: What was the reason for getting rid of this? Did it cause problems somewhere?

pranavm-nvidia: It was dead code

tripy/tripy/frontend/trace/ops/base.py
pranavm-nvidia: Since `DimensionSize` -> `Tensor` is a one-way conversion, we should try to keep things as `DimensionSize`s where possible. The problem is that the logic that I've currently implemented isn't completely right - not all Python integers actually represent dimension sizes and not all ops produce dimension sizes even if all their inputs are dimension sizes.

yizhuoz004: Is it possible to wrap another decorator to decide the return type? e.g.

```py
@infer_return_type
@convert_to_tensors()
def func(...)
```

tripy/tripy/utils/ast.py
slyubomirsky: Ah, I see, the code above handles it too.

tripy/tripy/frontend/dimension_size.py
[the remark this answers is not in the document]
slyubomirsky: Good point, probably not a case where treating Bools as numbers would be advisable.

tripy/tripy/frontend/trace/ops/base.py
slyubomirsky: > not all Python integers actually represent dimension sizes

We could present `DimensionSize` as meaning that the value _could_ be used to represent a dimension size, but maybe that's too clever by half. I think using a method or perhaps another decorator to indicate a policy for wrapping might be a sensible approach if isn't a way to avoid the issue.

pranavm-nvidia: Yes, an outer decorator is a good idea. Will update the PR accordingly.

tripy/tripy/frontend/utils.py
```diff
@@ -19,43 +19,40 @@
...
+    from tripy.common.datatype import int32
+    from tripy.frontend.dimension_size import DimensionSize
     from tripy.frontend.tensor import Tensor
-    from tripy.frontend.trace.ops.cast import cast
+    from tripy.frontend.trace.ops.concatenate import concatenate
+    from tripy.frontend.trace.ops.reshape import Reshape
+
+    if not arg:
+        return Tensor([], dtype=int32)
+
```
yizhuoz004, line 39: Might worth adding a simple optimization:
```py
if not arg or all(isinstance(e, int) for e in arg):
    return Tensor(arg, dtype=int32)
```

pranavm-nvidia: @yizhuoz004 Thanks, I implemented a generalized version of this approach which will accumulate contiguous sequences of integers instead of creating individual tensors for each.

tripy/tests/flat_ir/ops/test_iota.py
pranavm-nvidia: This is resolved by the fix for: https://github.com/NVIDIA/TensorRT-Incubator/pull/306#discussion_r1821340540
The IR should now be the same as before.

yizhuoz004 (approved):

tripy/tripy/frontend/trace/ops/base.py
pranavm-nvidia: Updated this as we discussed so that we always try casting to DimensionSize where possible.

tripy/tripy/frontend/trace/ops/base.py
```diff
@@ -72,21 +73,30 @@ def build(cls, inputs: List["Tensor"], *args, num_outputs=1, **kwargs) -> Union[
...
         inp_trace_tensors = [inp.trace_tensor for inp in inputs]
         out_trace_tensors = [out.trace_tensor for out in outputs]
         cls.build_internal(inp_trace_tensors, out_trace_tensors, *args, **kwargs)
 
+        # Operations that operate on only DimensionSize inputs will always yield a DimensionSize.
+        # For any mixed operations, DimensionSize must be casted up to Tensor.
+        all_inputs_are_dimension_size = all(isinstance(inp, DimensionSize) for inp in inputs)
+        for index, out in enumerate(outputs):
+            if always_cast_to_dimension_size or (
+                all_inputs_are_dimension_size and out.dtype == int32 and out.shape == []
```
yizhuoz004, line 93: Will `out.shape` trigger eval here?

tripy/tripy/frontend/trace/ops/base.py
```diff
@@ -72,21 +73,30 @@ def build(cls, inputs: List["Tensor"], *args, num_outputs=1, **kwargs) -> Union[
...
         # the Tensor constructor!
-        outputs = [TensorType(None) for _ in range(num_outputs)]
+        outputs = [Tensor(None) for _ in range(num_outputs)]
 
         inp_trace_tensors = [inp.trace_tensor for inp in inputs]
         out_trace_tensors = [out.trace_tensor for out in outputs]
         cls.build_internal(inp_trace_tensors, out_trace_tensors, *args, **kwargs)
 
+        # Operations that operate on only DimensionSize inputs will always yield a DimensionSize.
+        # For any mixed operations, DimensionSize must be casted up to Tensor.
```
yizhuoz004, line 89: Shall we add simple unit tests to guard this logic?

parthchadha (approved):

tripy/tripy/frontend/trace/ops/base.py
slyubomirsky: I think not since these would be trace tensors.

slyubomirsky (approved):

tripy/tripy/frontend/trace/ops/base.py
pranavm-nvidia: @yizhuoz004 I did add a test under `tripy/tests/frontend/trace/ops/test_binary_elementwise.py`. I could add a separate test with a mock operation, but I don't think it provides any additional coverage.

tripy/tripy/frontend/trace/ops/base.py
pranavm-nvidia: Oh good catch, yes it would. I meant to do `out.trace_tensor.shape`

pranavm-nvidia: Fixed.

PR #309, opened 2024-10-28
[tripy] Sequential module

farazkh80:
Sequential similar to [torch.nn.sequential ](https://pytorch.org/docs/stable/generated/torch.nn.Sequential.html)

Things done

- [x] new frontend module called `tp.Sequential`
- [x] Unit tests under `tests/frontend/module/test_sequential.py`
- [x] integration tests and comparison with torch under

Supports nested tp.Sequential and most list operations other than modifications. `tp.Sequential` can not be modified after creation.

### DIFF
--- a/tripy/tripy/frontend/module/sequential.py
+++ b/tripy/tripy/frontend/module/sequential.py
@@ -5,6 +5,7 @@ from tripy import export
 from tripy.frontend.module.parameter import Parameter
 from tripy.frontend.module import Module
 
+
 @export.public_api(document_under="modules/sequential.rst")
 @dataclass
 class Sequential(Module):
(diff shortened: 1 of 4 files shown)


tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,284 @@
...
+
+            model = tp.Sequential(tp.Linear(1, 3), tp.Linear(3, 2))
+            for layer in model:
+                print(layer)
+        """
+        return iter(self._modules.values())
+
+    def __getitem__(self, idx: Union[int, str]) -> Module:
+        r"""
+        Accesses a module by index (int) or name (str).
```
pranavm-nvidia, line 100: Is `int` here valid syntax? I think the interpreter would reject something like `module.0`.

slyubomirsky: I think it would be fine. `__getitem__` is array-style access.

pranavm-nvidia: Ah my mistake, I misread it as `__getattr__`.

farazkh80:
closing sin e I have a feature branch MR now

PR #312, opened 2024-10-28
NFC: Add MLIR-TensorRT 'CMakePresets.json' and update build documentation

christopherbate:
This change adds a CMake presets file to MLIR-TensorRT, which helps to
simplify the initial CMake setup considerably.

The build documentation is updated to use this feature, and some additional
notes are added to describe how to control the TensorRT version that
is used for build & test.


shelkesagar29 (approved):
LGTM

PR #313, opened 2024-10-28
[TensorRT/transforms] Fix TileLikeBroadcastToSlice to set output shapes if known

parthchadha:
Fix TileLikeBroadcastToSlice to set output shapes if known.

### DIFF
--- a/mlir-tensorrt/tensorrt/lib/TensorRT/Transforms/ExpandOps.cpp
+++ b/mlir-tensorrt/tensorrt/lib/TensorRT/Transforms/ExpandOps.cpp
@@ -299,30 +299,30 @@ struct TileLikeBroadcastToSlice : public OpRewritePattern<BroadcastOp> {
       input = reshapeBroadcastInput(rewriter, op.getLoc(), op);
     TensorType inputType = input.getType();
 
-    // Create static_start and static_stride attributes
-    auto static_start = rewriter.getDenseI32ArrayAttr(
+    // Create staticStart and staticStride attributes
+    auto staticStart = rewriter.getDenseI32ArrayAttr(
         SmallVector<int32_t>(inputType.getRank(), 0));
-    auto static_stride = rewriter.getDenseI32ArrayAttr(
+    auto staticStride = rewriter.getDenseI32ArrayAttr(
         SmallVector<int32_t>(inputType.getRank(), 1));
 
     if (resultType.hasStaticShape()) {
       // For static shapes, use DenseI32ArrayAttr for size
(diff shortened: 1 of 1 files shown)


mlir-tensorrt/executor/lib/Runtime/API/API.cpp
```diff
@@ -414,6 +414,9 @@ void AllocTracker::incrementExternalCount(uintptr_t ptr) {
 }
 
 void AllocTracker::decrementExternalCount(uintptr_t ptr) {
+  if (!llvm::is_contained(map, ptr))
+    return;
+
   assert(llvm::is_contained(map, ptr) &&
```
jhalakpatel, line 420: This assertion will always be true now. @parthchadha Do we have a unit test to repro this issue? May be there is a better fix?

parthchadha: Very hard to construct a unit test, it fails 100% time in SAM2 example though.

mlir-tensorrt/tensorrt/lib/TensorRT/Transforms/ExpandOps.cpp
```diff
@@ -297,21 +299,31 @@ struct TileLikeBroadcastToSlice : public OpRewritePattern<BroadcastOp> {
       input = reshapeBroadcastInput(rewriter, op.getLoc(), op);
     TensorType inputType = input.getType();
 
-    SmallVector<int32_t> start(inputType.getRank(), 0);
-    SmallVector<int32_t> stride(start.size(), 1);
+    // Create static_start and static_stride attributes
+    auto static_start = rewriter.getDenseI32ArrayAttr(
```
jhalakpatel, line 303: Why do we need to explicitly convert to `DenseI32ArrayAttr`? It looks like the previous code did not require that.

parthchadha: The code switches to using another way of constructing `SliceOp` to make it consistent with dynamic shapes code path (check the various `build` definitions in `TensorRTOps.h.inc`.

parthchadha:
> Each change needs to go in a separate MR with accompanying tests. If you need help with formulating a test, let me know.

I have remove the API.cpp change from this MR and the current MR has the change with its relevant unit test.

christopherbate (commented):
Forgot to click "submit" my review that I wrote earlier, please disregard any out-of-date comments.

jhalakpatel (approved):

PR #314, opened 2024-10-28
Update IOutputAllocator usage to be compatible with TensorRT 9.2 APIs

jhalakpatel:
`IOutputAllocator` needs to be supported for both TRT 9.2 and TRT 10.x release. Add support for using TRT 9.2 API.

### DIFF
--- a/mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
+++ b/mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
@@ -26,6 +26,7 @@
 #include "mlir-executor/Runtime/Backend/Utils/NvtxUtils.h"
 #include "mlir-executor/Support/Allocators.h"
 #include "mlir-executor/Support/Status.h"
+#include "mlir-tensorrt-dialect/Utils/TensorRTVersion.h"
 #include <memory>
 
 #if defined(__GNUC__) || defined(__clang__)
@@ -130,6 +131,7 @@ public:
     mOutputSize = size;
   }
 
+#if MLIR_TRT_COMPILE_TIME_TENSORRT_VERSION_GTE(10, 0, 0)
   void *reallocateOutputAsync(const char *name, void *memory, uint64_t size,
                               uint64_t alignment,
                               cudaStream_t stream) override {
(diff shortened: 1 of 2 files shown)


mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -155,6 +164,30 @@ class OutputAllocatorImpl : public nvinfer1::IOutputAllocator {
     }
     return reinterpret_cast<void *>(mOutputPtr);
   }
+#else
+  void *reallocateOutput(const char *name, void *memory, uint64_t size,
```
shelkesagar29, line 168: Does this work with TRT 8.6 too?

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -37,6 +37,14 @@
 #pragma GCC diagnostic pop
 #endif
 
+/// Evaluates to true if the version of TensorRT that we are compiling against
+/// is greater than or equal to the given version number.
+#define TENSORRT_MODULE_VERSION_GTE(major, minor, patch)                       \
```
shelkesagar29, line 42: Can we use `MLIR_TRT_COMPILE_TIME_TENSORRT_VERSION_GTE` that is already available?

jhalakpatel: I was running into some compilation issues while including `mlir-tensorrt/tensorrt/include/mlir-tensorrt-dialect/Utils/TensorRTVersion.h` here.
I did not investigate the compilation issue further, but does that dependency make sense in the `mlir-tensorrt/executor`? If yes, I can update the Makefile to add the dependency.

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
jhalakpatel: Yes, should work with TRT 8.6 as well.

mlir-tensorrt/executor/test/IntegrationTests/lua_results.mlir
```diff
@@ -0,0 +1,24 @@
+// RUN: rm %t.mlir || true
```
jhalakpatel, line 1: Please ignore this file. I will remove it soon.

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
shelkesagar29: This should depend only on `MLIRTRTTensorRTDynamicLoader`, I think. Adding dependency is okay.

jhalakpatel: I think I needed to explicitly include a header directory. Link time dependency is not required.

shelkesagar29 (approved):

PR #318, opened 2024-10-29
[NFC][mlir-tensorrt] Fix minor issues

shelkesagar29:
This PR makes the following changes,
- Add code owners for `mir-tensorrt` part to help review.
- Add torch as dependency in `requirements-dev.txt`. Torch based test is used in Python API.
- Fix an error in README.

### DIFF
--- a/mlir-tensorrt/python/requirements-dev.txt
+++ b/mlir-tensorrt/python/requirements-dev.txt
@@ -1,6 +1,7 @@
 # Torch is required by `optimum`. This will let PIP select the CPU version.
 # The GPU version installs `nvidia-` pip packages which confuse JAX.
 -r requirements.txt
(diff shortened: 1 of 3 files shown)


christopherbate (approved):

mlir-tensorrt/python/requirements-dev.txt
```diff
@@ -13,3 +13,4 @@ nvidia-ml-py
 black
 build
 psutil
+torch
```
christopherbate, line 16: this should be more specific in terms of features/version. Internally we have 

```
--extra-index-url https://download.pytorch.org/whl/cpu
torch==2.1.0+cpu
```

shelkesagar29: It should be installing cpu version by default. I made it explicit.

Issue #295, opened 2024-10-22
Sequential module support


farazkh80:
Currently, `tripy` supports sequential modules if defined as a list of modules. However, this is not discoverable for users. A better way, is to add `Sequential` to front end api and to the documentation.

farazkh80:
WIP PR https://github.com/NVIDIA/TensorRT-Incubator/pull/309

farazkh80:
Closing since now supported

PR #321, opened 2024-10-29
[tripy] Sequential module feature branch

farazkh80:
Sequential similar to [torch.nn.sequential ](https://pytorch.org/docs/stable/generated/torch.nn.Sequential.html) for addressing https://github.com/NVIDIA/TensorRT-Incubator/issues/295

Things done

- [x] new frontend module called `tp.Sequential`
- [x] Unit tests under `tests/frontend/module/test_sequential.py`
- [x] integration tests and comparison with torch under

Supports nested tp.Sequential and most list operations other than modifications. `tp.Sequential` can not be modified after creation.


tripy/tests/frontend/module/conftest.py
```diff
@@ -139,3 +139,12 @@ def mixed_network():
...
     yield ComplexNetwork()
+
+@pytest.fixture
+def sequential_network():
+    yield tp.Sequential(tp.Linear(1, 3), tp.Linear(3, 2))
+
+
+@pytest.fixture
+def nested_sequential_network():
+    yield tp.Sequential(tp.Linear(2, 4), tp.Sequential(tp.Linear(4, 3), tp.Linear(3, 1)))
```
pranavm-nvidia, line 150: Could we define this in the test file? `conftest.py` is nice when we want to share fixtures but it otherwise just makes tests harder to read IMO.

tripy/tests/integration/test_sequential.py
```diff
@@ -0,0 +1,90 @@
...
+            torch.nn.Linear(1, 3, dtype=torch.float32), torch.nn.Linear(3, 2, dtype=torch.float32)
+        )
+        tp_model = tp.Sequential(tp.Linear(1, 3, dtype=tp.float32), tp.Linear(3, 2, dtype=tp.float32))
+
+        tp_model[0].weight = tp.Parameter(torch_model[0].weight.detach())
+        tp_model[0].bias = tp.Parameter(torch_model[0].bias.detach())
+        tp_model[1].weight = tp.Parameter(torch_model[1].weight.detach())
+        tp_model[1].bias = tp.Parameter(torch_model[1].bias.detach())
+
+        input_tensor = torch.tensor([[1.0]], dtype=torch.float32).to('cuda')
```
pranavm-nvidia, line 33: Let's just set the device in the constructor so we don't need to do the copy

tripy/tests/integration/test_sequential.py
```diff
@@ -0,0 +1,90 @@
...
+        tp_input = tp.Tensor(input_tensor, dtype=tp.float32)
+
+        tp_output = tp_model(tp_input)
+
+        torch_model.to('cuda').eval()
```
pranavm-nvidia, line 38: Can we just set the device in the `torch.nn.Linear` constructor so we don't incur a copy here?

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,169 @@
...
+@dataclass
+class Sequential(Module):
+    r"""
+    A module to stack multiple layers or modules in a sequential order. The `Sequential`
+    container can accept either a list of modules or a dictionary of named modules. Modules are
+    added in the order they are passed, and each is called sequentially during the forward pass.
+    """
+
+    modules: Union[Module, Dict[str, Module]]
+    r"""The modules to include in the sequence"""
```
pranavm-nvidia, line 19: Should this be publicly accessible/documented? If not, we can drop these two lines.

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,169 @@
...
+        r"""
+        Args:
+            *modules (Union[Module, Dict[str, Module]]): The modules to include in the sequence.
+                Can be passed as individual positional arguments or as a single dictionary of named modules.
+
+        .. code-block:: python
+            :linenos:
+            :caption: Example
+
+            import tripy as tp
```
pranavm-nvidia, line 31:
```suggestion
```

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,169 @@
...
+
+            # Sequential with layers passed as arguments
+            model = tp.Sequential(tp.Linear(1, 3), tp.Linear(3, 2))
+
+            # Sequential with named dictionary of layers
+            model_named = tp.Sequential({'layer1': tp.Linear(1, 3), 'layer2': tp.Linear(3, 2)})
```
pranavm-nvidia, line 37: Could you make this a separate example and also show the inputs/outputs?

pranavm-nvidia: You can look at the `Conv` module for how to include multiple examples in the docstring.

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,169 @@
...
+
+            x = tp.Tensor([1.0])
+            output = model(x)
+        """
+        super().__init__()
+        self.modules = {}
+
+        if len(modules) == 1 and isinstance(modules[0], dict):
```
pranavm-nvidia, line 45: Should we disallow mixing modules/dicts in the constructor?

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,169 @@
...
+            for name, module in modules[0].items():
+                self.modules[name] = module
```
pranavm-nvidia, line 47: Could probably just do: `self.modules = copy.copy(modules[0])`.

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        for module in self.modules.values():
+            x = module(x)
+        return x
+
+    def __getattr__(self, name) -> Any:
+        """
+        Custom __getattr__ to search both in `modules` dictionary and in other attributes. This is for handling
+        `module = operator.attrgetter(child_name)(module)` calls in tripy/frontend/module/module.py:load_state_dict
+        """
```
pranavm-nvidia, line 70: Could you either switch this docstring to a comment or explicitly exclude this method from the public documentation by setting `autodoc_options` in the `public_api` decorator (see `tripy/frontend/tensor.py` for an example)?

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,169 @@
...
+        # Check if `name` is a key in the modules dictionary
+        if "modules" in self.__dict__ and name in self.modules:
+            return self.modules[name]
+
+        # Fallback to regular attribute access if not found in modules
+        try:
+            return super().__getattr__(name)
+        except AttributeError:
+            raise AttributeError(f"'{self.__class__.__name__}' object has no attribute or module named '{name}'")
```
pranavm-nvidia, line 79: I think we can just let the original attribute error go through here:
```suggestion
        return super().__getattr__(name)
```

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def __len__(self) -> int:
+        r"""
+        Returns the total number of modules in the sequence.
+
+        Returns:
+            int: The number of modules in the sequence.
```
pranavm-nvidia, line 86:
```suggestion
            The number of modules in the sequence.
```

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,169 @@
...
+
+        .. code-block:: python
+            :linenos:
+            :caption: Example
+
+            model = tp.Sequential(tp.Linear(1, 64), tp.Linear(64, 128))
+            assert len(model) == 2
```
pranavm-nvidia, line 93: Assertions are stripped from public docs, so we should do something like this to show the behavior:
```suggestion
            # doc: print-locals model length
            
            model = tp.Sequential(tp.Linear(1, 64), tp.Linear(64, 128))
            length = len(model)
            assert length == 2
```

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def __getitem__(self, idx: Union[int, str]) -> Module:
+        r"""
+        Accesses a module by index (int) or name (str).
+
+        Args:
+            idx (Union[int, str]): The index or name of the module to retrieve.
+
+        Returns:
+            Module: The module at the specified index or name.
```
pranavm-nvidia, line 122:
```suggestion
            idx: The index or name of the module to retrieve.

        Returns:
            The module at the specified index or name.
```

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,169 @@
...
+
+        return self.modules[key]
+
+    def named_children(self) -> Iterator[Tuple[str, "Module"]]:
+        r"""
+        Returns an iterator over all child modules in this `Sequential` container.
+        Each child module is represented by its name and the module object itself.
+
+        Returns:
+            Iterator[Tuple[str, Module]]: An iterator over tuples containing
```
pranavm-nvidia, line 150:
```suggestion
            An iterator over tuples containing
```

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,169 @@
...
+        else:
+            for idx, module in enumerate(modules):
+                self.modules[str(idx)] = module
+
+    def __call__(self, x: "tripy.Tensor") -> "tripy.Tensor":
+        r"""
+        Defines the forward pass by applying each module in the container sequentially to input `x`.
+
+        Args:
+            x (Tensor): The input tensor to pass through the sequence of modules.
```
pranavm-nvidia, line 57:
```suggestion
            x: The input tensor to pass through the sequence of modules.
```

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,169 @@
...
+
+        Returns:
+            Tensor: The output tensor after passing through each module in sequence.
```
pranavm-nvidia, line 60:
```suggestion
            The output tensor after passing through each module in sequence.
```

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,169 @@
...
+    added in the order they are passed, and each is called sequentially during the forward pass.
+    """
+
+    modules: Union[Module, Dict[str, Module]]
+    r"""The modules to include in the sequence"""
+
+    def __init__(self, *modules: Union[Module, Dict[str, Module]]) -> None:
+        r"""
+        Args:
+            *modules (Union[Module, Dict[str, Module]]): The modules to include in the sequence.
```
pranavm-nvidia, line 24:
```suggestion
            *modules: The modules to include in the sequence.
```

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,169 @@
...
+                print(f"{name}: {type(child).__name__}")
+
+        """
+        for name, module in self.modules.items():
+            if isinstance(module, Module):
+                yield name, module
+            elif isinstance(module, Sequential):
+                # Traverse deeper into Sequential containers within Sequential
+                for child_name, child in module.named_children():
+                    yield f"{name}.{child_name}", child
```
pranavm-nvidia, line 169: `named_children` should only return the first-order children of the module. It is called recursively when required.

tripy/tripy/frontend/module/sequential.py
farazkh80: dropping, should not be accesible

tripy/tests/frontend/module/conftest.py
farazkh80: done

tripy/tests/integration/test_sequential.py
farazkh80: done

tripy/tests/integration/test_sequential.py
farazkh80: done

tripy/tripy/frontend/module/sequential.py
farazkh80: done

tripy/tripy/frontend/module/sequential.py
farazkh80: done

tripy/tripy/frontend/module/sequential.py
farazkh80: done

tripy/tripy/frontend/module/sequential.py
farazkh80: done

tripy/tripy/frontend/module/sequential.py
farazkh80: done

farazkh80:
Addressed PR changes, and added some more tests for `Dict` mode of Sequential @pranavm-nvidia

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,186 @@
...
+            input = module(input)
+        return input
+
+    def __getattr__(self, name) -> Any:
+        """
+        Custom __getattr__ to search both in `modules` dictionary and in other attributes. This is for handling
+        `module = operator.attrgetter(child_name)(module)` calls in tripy/frontend/module/module.py:load_state_dict
+        """
+        # Check if `name` is a key in the modules dictionary
+        if "modules" in self.__dict__ and name in self.modules:
```
pranavm-nvidia, line 94: Is the first part of the check necessary? Wouldn't the modules member always be defined?

tripy/tripy/frontend/module/sequential.py
pranavm-nvidia: After thinking about this some more, I think it's probably better to let the function registry's type checking handle this so we don't need to duplicate type checking logic. 

For that to work we will need to implement #338 and #341 but I think we can omit this check for now.

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,186 @@
...
+
+        .. code-block:: python
+            :linenos:
+            :caption: Example
+
+            model = tp.Sequential(tp.Linear(1, 3), tp.Linear(3, 2))
+            print(model[1])
+        """
+        if not isinstance(idx, (int, str)):
+            raise TypeError("Index must be an int or str.")
```
pranavm-nvidia, line 157: We can probably omit this check also since the function registry will eventually be able to handle this

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,186 @@
...
+
+        key = str(idx) if isinstance(idx, int) else idx
+
+        if key not in self.modules:
+            raise ValueError(f"Key {key} not found in modules.")
```
pranavm-nvidia, line 162: Let's use `raise_error`:
```suggestion
            raise_error(f"Key: '{key}' not found in modules.", [f"Note: Available keys were: {list(self.modules.keys())}"])
```

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,186 @@
...
+
+        return self.modules[key]
+
+    def named_children(self) -> Iterator[Tuple[str, "Module"]]:
```
pranavm-nvidia, line 166: Do we actually need to implement this? I think the base class implementation should just work

farazkh80: the base implementation with display them every children with `modules` prefix, which is sth we want to avoid.

pranavm-nvidia: Could you add a comment in the code explaining that?

farazkh80: added

tripy/tripy/frontend/module/sequential.py
farazkh80: remvoed

tripy/tripy/frontend/module/sequential.py
farazkh80: removed

tripy/tripy/frontend/module/sequential.py
farazkh80: okay, sounds good. removing

farazkh80: sounds good, removing!

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,185 @@
...
+
+from tripy import export
+from tripy.common.exception import raise_error
+from tripy.frontend.module import Module
+
+
+@export.public_api(
+    document_under="modules/sequential.rst",
+    autodoc_options=[
+        ":exclude-members: __getattr__",
```
pranavm-nvidia, line 27: Could you double check that this is working?

tripy/tripy/frontend/module/sequential.py
```diff
@@ -0,0 +1,185 @@
...
+        return input
+
+    def __getattr__(self, name) -> Any:
+        """
+        Custom __getattr__ to search both in `modules` dictionary and in other attributes. This is for handling
+        `module = operator.attrgetter(child_name)(module)` calls in tripy/frontend/module/module.py:load_state_dict
+        """
+        # Check if `name` is a key in the modules dictionary
+        if name in self.modules:
+            return self.modules[name]
```
pranavm-nvidia, line 92: Does this also work for integers? I'm wondering if it would be better to make `Sequential` a subclass of `dict` as well. That would also allow us to drop len/iter/getitem below.

farazkh80: hmmm I rather not, since dict has other methods that we don't necessarily want Sequential to have like, `__setattr__` after construction.

Also, subclass of dict and Module would be multiple inheritance and could be a problem in case Module and dict share any mutual methods which could introduce hidden bugs for Sequential.

farazkh80: Instead i can just cast name to str and add typing to name `name: Union[str, int]`

tripy/tripy/frontend/module/sequential.py
farazkh80: Actually, I will remove this, since this matters only when we have `special_members` ON.

tripy/tripy/frontend/module/sequential.py
pranavm-nvidia: But how can you call `__getattr__` with an integer? Doing something like `module.0` wouldn't be valid syntax

farazkh80: you are right. Should we just allow strings then?

pranavm-nvidia: That's why I was suggesting inheriting from `dict`; we have special handling that will bypass the `operator.attrgetter` for dictionaries. Should also work to only allow strings though, especially if we don't expect this to be user facing.

farazkh80: yeah I see. I added the string type to `name` and yes since it does not face users, I think it will work.

But I still don't like inheriting from dict, since we will inherit other methods that we don't necessarily need or want.

pranavm-nvidia (approved):

PR #293, opened 2024-10-21
Resent50 Notebook

farazkh80:

PR #322, opened 2024-10-29
[Tripy] Resnet50 example

farazkh80:
Resnet 50 for image classification example notebook implementation in `tripy`.

- [x] Resnet50 implementation, weight loading, compilation and demo with tripy
- [x] Added `pytest --nbval-lax` support for testing notebooks
- [x] Added and tested TRT JIT (link to posted [results](https://nvidia.slack.com/archives/C06LCFTTRH7/p1730745579886509) on slack) 
- [x] added an [issue](https://github.com/NVIDIA/TensorRT-Incubator/issues/335) for adaptive average pooling constraining dynamic shape support 

new PR on top of https://github.com/NVIDIA/TensorRT-Incubator/pull/293 since the fork's branch fails pipeline


farazkh80:
added testing using `pytest --nbval-lax` which runs each notebook cells and passes if all cells run without errors. @parthchadha @pranavm-nvidia

farazkh80:
PR is ready for review

- [x] Added `pytest --nbval-lax` support for testing notebooks
- [x] Added and tested TRT JIT (link to posted [results](https://nvidia.slack.com/archives/C06LCFTTRH7/p1730745579886509) on slack) 
- [x] added an [issue](https://github.com/NVIDIA/TensorRT-Incubator/issues/335) for adaptive average pooling constraining dynamic shape support

.github/workflows/tripy-l0.yml
```diff
@@ -70,6 +70,7 @@ jobs:
         options: --gpus all -v ${{ github.workspace }}/tripy:/tripy
         run: |
           pytest --cov=tripy/ --cov-config=.coveragerc tests/ -v -m "not l1 and not manual" -n 4 --durations=15 --ignore tests/performance
+          find . -name '*.ipynb' -print0 | xargs -0 pytest --nbval-lax
```
pranavm-nvidia, line 73: Any way to have this run as part of our normal pytest command? Seems like there's [another extension that can do this](https://pytest-notebook.readthedocs.io/en/latest/user_guide/tutorial_intro.html#pytest-file-collection).

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
+{
```
pranavm-nvidia, line 1: Directory name has a typo: `resent50` -> `resnet50`.

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# ResNet50 Image Classificiation with `tripy` - An Implementation and Demo\n",
```
pranavm-nvidia, line 7: I think we can shorten the title:
```suggestion
    "# ResNet50 Image Classificiation",
```
`tripy` and "Demo" are implied by the location of the notebook

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+    "\n",
+    "This notebook aims to demonstrate the implementation of a ResNet50 model using `tripy`. We'll explore the architecture, \n",
+    "load pretrained weights, and run predictions on a sample dataset, showcasing how `tripy` can be used effectively for inference of ResNet50 based image classification model.\n",
```
pranavm-nvidia, line 10: The last bit is redundant:
```suggestion
    "This notebook aims to demonstrate the implementation of a ResNet50 model using `tripy`. We'll explore the architecture, \n",
    "load pretrained weights, and run predictions on a sample dataset.\n",
```

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+    "\n",
+    "### Objectives:\n",
+    "1. Implement ResNet50 using `tripy` and load pretrained weights.\n",
+    "2. Run predictions on sample images and visualize results.\n"
```
pranavm-nvidia, line 14: I think we can drop the objectives section since we already explained what we're going to do above?
In general, let's try to keep things brief. Assume short attention spans :slightly_smiling_face:

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Setup\n",
+    "\n",
+    "install neccessary libraries, load pretrained resnet50 weights and the inference dataset."
+   ]
+  },
```
pranavm-nvidia, line 25: The important part here is the loading of the weights and example image:
```suggestion
  {
   "cell_type": "markdown",
   "metadata": {},
   "source": [
    "# Setup\n",
    "\n",
    "Load pretrained resnet50 weights and an example image:"
   ]
  },
```

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+    "!pip install datasets matplotlib pillow"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Load necessary libraries\n",
```
pranavm-nvidia, line 41:
```suggestion
```

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+    "from transformers import ResNetForImageClassification, AutoImageProcessor\n",
+    "from datasets import load_dataset\n",
+    "\n",
+    "# Load the pretrained ResNet50 model from Hugging Face\n",
+    "resnet_pretrained = ResNetForImageClassification.from_pretrained(\"microsoft/resnet-50\")\n",
+    "resnet_pretrained.eval()\n",
+    "resnet_pretrained = resnet_pretrained.to('cuda')\n",
+    "\n",
+    "# Load the image processor\n",
+    "processor = AutoImageProcessor.from_pretrained(\"microsoft/resnet-50\")\n",
```
pranavm-nvidia, line 52: Can we move this down to where it's used?

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Preprocess the image\n",
+    "inputs = processor(image, return_tensors=\"pt\")['pixel_values'].to('cuda')\n",
+    "\n",
+    "# Run the image through the model\n",
```
pranavm-nvidia, line 70:
```suggestion
    "# Run the original PyTorch model\n",
```

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+    "\n",
+    "# Load a sample image from the dataset\n",
+    "dataset = load_dataset(\"comet-team/coco-500\", split=\"train\")\n",
+    "idx = 103\n",
+    "image = dataset[idx]['Image']\n",
+    "image"
+   ]
+  },
+  {
+   "cell_type": "code",
```
pranavm-nvidia, line 62: I don't use notebooks a lot - is there some convention on how code is split between cells? Should each conceptually different thing, e.g. loading weights, running inference, be in its own cell?

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+    "import tripy as tp"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We will define:\n",
+    "\n",
+    "1. **ResNet Building Blocks**:\n",
```
pranavm-nvidia, line 101: Maybe we can include this information as docstrings/comments in the modules themselves? I think it would be helpful to have this high-level information where we can also look at the code.

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+    "        self.activation = activation\n",
+    "\n",
+    "    def __call__(self, x):\n",
+    "        x = self.convolution(x)\n",
+    "        x = self.normalization(x)\n",
+    "        if self.activation:\n",
+    "            x = tp.relu(x)\n",
+    "        return x\n",
+    "    \n",
+    "class ResNetShortCut(tp.Module):\n",
```
pranavm-nvidia, line 184: Should each module be its own cell?

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+    "    converting them into tp.Parameter before loading.\n",
+    "    \"\"\"\n",
+    "    model_dict = model.state_dict()  # Get the model's state dict\n",
+    "\n",
+    "    # Convert each weight to a tp.Parameter and update the model's state dict\n",
+    "    for k, v in pretrained_state_dict.items():\n",
+    "        if 'num_batches_tracked' in k:\n",
+    "            continue    # Skip num_batches_tracked since it is not needed for inference (tp.BatchNorm does not support)\n",
+    "        if 'classifier' in k:\n",
+    "            k = k.replace('1.', '') # Remove extra 1 (slight naming mismtach)\n",
```
pranavm-nvidia, line 358: Any idea why this mismatch is happening?

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+    "First we ensure all state dict keys match and then we load the pretrained weights."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def load_pretrained_weights_tp(model, pretrained_state_dict):\n",
```
pranavm-nvidia, line 346:
```suggestion
    "def load_pretrained_weights(model, pretrained_state_dict):\n",
```

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+    "    model_dict = model.state_dict()  # Get the model's state dict\n",
+    "\n",
+    "    # Convert each weight to a tp.Parameter and update the model's state dict\n",
+    "    for k, v in pretrained_state_dict.items():\n",
+    "        if 'num_batches_tracked' in k:\n",
+    "            continue    # Skip num_batches_tracked since it is not needed for inference (tp.BatchNorm does not support)\n",
+    "        if 'classifier' in k:\n",
+    "            k = k.replace('1.', '') # Remove extra 1 (slight naming mismtach)\n",
+    "\n",
+    "        assert k in set(model_dict.keys())\n",
```
pranavm-nvidia, line 360: nit: would be a bit more efficient to initialize this set outside the loop

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def load_pretrained_weights_tp(model, pretrained_state_dict):\n",
+    "    \"\"\"\n",
+    "    This function loads weights from a state_dict into the custom TPResNet model,\n",
+    "    converting them into tp.Parameter before loading.\n",
+    "    \"\"\"\n",
+    "    model_dict = model.state_dict()  # Get the model's state dict\n",
```
pranavm-nvidia, line 351: Is there a reason we can't just start with an empty dictionary?

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Run a dummy forward path\n",
+    "x = tp.ones(input_shape, dtype=tp.float32)\n",
+    "compiled_output = compiled_model(x)\n",
+    "compiled_output.shape"
```
pranavm-nvidia, line 397: What's the reason for this dummy forward pass?

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Prepare the image for Tripy\n",
+    "processed_image = processor(image, return_tensors=\"np\")['pixel_values']\n",
+    "tp_image = tp.Tensor(processed_image, dtype=tp.float32, device=tp.device(\"gpu\"))\n",
+    "tp_image.shape"
```
pranavm-nvidia, line 418: Do we not need a `print()` here?

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Get the predicted label\n",
+    "tp_predicted_label = torch.argmax(torch.from_dlpack(tp_logits), -1).tolist()[0]\n",
```
pranavm-nvidia, line 439: Can we use `tp.argmax`?

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+   "cell_type": "code",
+   "execution_count": 18,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def get_tp_pred(model, image, processor):\n",
+    "    processed_image = processor(image, return_tensors=\"np\")['pixel_values']\n",
+    "    tp_image = tp.Tensor(processed_image, dtype=tp.float32, device=tp.device(\"gpu\"))\n",
+    "    tp_logits = model(tp_image)\n",
+    "    return torch.argmax(torch.from_dlpack(tp_logits), -1).tolist()[0]\n",
```
pranavm-nvidia, line 454: This seems to be duplicating what we did above

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+    "\n",
+    "def get_exact_answer(model, image, processor):\n",
```
pranavm-nvidia, line 456: Can we rename to `get_torch_pred`?

tripy/examples/resent50/resnet50.ipynb
farazkh80: we can but @parthchadha suggested earlier to use torch for pre and post processing of input/outputs IIRC. Should I still change it to tp.argmax?

tripy/examples/resent50/resnet50.ipynb
farazkh80: just to demonstrate that compiled model works. Can remove it, since we actually do the demo down below.

tripy/examples/resent50/resnet50.ipynb
farazkh80: Yeah, it is better if we seperate them so each cell is concise. Unless code is highly correlated like running forward path on image and showing the plot are fine to put on the same cell. But I think for model we should have Modules separate from weight loading and inference

tripy/examples/resent50/resnet50.ipynb
farazkh80: yeah will include it as well.

tripy/examples/resent50/resnet50.ipynb
farazkh80: I think that's not neccessary, we could do one big cell, but that would be a bit ugly. I tried to keep relevant parts together on the same cell.

tripy/examples/resent50/resnet50.ipynb
farazkh80: because torch does sequential on `Flatten` and `Linear` for the classification head. We have sequential but can't mix them until the mixed container PR is in.

tripy/examples/resent50/resnet50.ipynb
farazkh80: Hmm, I can remove it if you just want to do asserts. It was more to demonstrate that the trophy model works the same as the torch model since they both generate the same label. But I guess I did not communicate that well. 

Question is,
1- do we want to keep the cross comparison with torch model in this notebook? or do we just use torch model weights and run the demo on Tripy and let the user figure out that the labels for the pictures are reasonable?
2- letting the user figure out maybe ambiguous though, since If I am a user, I would personally be curious to see if the tripy model performs the same as torch model? How can I see that without that label assertion/print? (this is completely subjective though, since it is a tripy example demo, it is probably not necessary...)

tripy/examples/resent50/resnet50.ipynb
farazkh80: you mean include tp.argmax in the model definition? so we don't return logits, we return `tp.argmax(logits)`?

tripy/examples/resent50/resnet50.ipynb
pranavm-nvidia: Makes sense. Can we do the mixed container PR first then so we can fix this before merging?

tripy/examples/resent50/resnet50.ipynb
pranavm-nvidia: It doesn't have to be part of the modules, but it should be part of what we compile. So for example, you could have something like:
```py
def infer(x):
    logits = model(x)
    return tp.argmax(logits)

compiled_model = tp.compile(infer, ...)
```

tripy/examples/resent50/resnet50.ipynb
pranavm-nvidia: The key point is that the model works as its expected to (i.e. is able to correctly classify images). Comparing against torch is just a proxy for that, and if we already know what labels we expect, we can check directly against those instead of introducing more complexity.

tripy/examples/resent50/resnet50.ipynb
farazkh80: removed

tripy/examples/resent50/resnet50.ipynb
farazkh80: done

tripy/examples/resent50/resnet50.ipynb
farazkh80: Thanks! Fixed

tripy/examples/resent50/resnet50.ipynb
farazkh80: I reduced number of cells now only 3 cells
- 1 cell for defining the core convolution + Norm layers
- 1 for building blocks
- 1 for actual backbone and classifier wrapper

tripy/examples/resent50/resnet50.ipynb
```diff
@@ -0,0 +1,529 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def load_pretrained_weights_tp(model, pretrained_state_dict):\n",
+    "    \"\"\"\n",
+    "    This function loads weights from a state_dict into the custom TPResNet model,\n",
+    "    converting them into tp.Parameter before loading.\n",
+    "    \"\"\"\n",
```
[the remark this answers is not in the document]
farazkh80: done

tripy/examples/resent50/resnet50.ipynb
farazkh80: We can! fixed now

tripy/examples/resent50/resnet50.ipynb
farazkh80: removed

tripy/examples/resent50/resnet50.ipynb
farazkh80: I see! that's great. Added it

tripy/examples/resent50/resnet50.ipynb
farazkh80: removed

tripy/examples/resent50/resnet50.ipynb
farazkh80: Sounds good, added hard coded labels!

tripy/examples/resent50/resnet50.ipynb
farazkh80: Merged mixed container, and fixed this now!

tripy/examples/resnet50/resnet50.ipynb
```diff
@@ -0,0 +1,383 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Load the image processor\n",
+    "processor = AutoImageProcessor.from_pretrained(\"microsoft/resnet-50\")\n",
+    "\n",
+    "# Run the pre-trained PyTorch model on image as a sanity check for weights\n",
```
pranavm-nvidia, line 64: I think we can probably drop the sanity check. It feels more like testing code than something we are trying to demonstrate.

.github/workflows/tripy-l0.yml
farazkh80: We can't really do one command, since we would need to specify another location to look for tests, `examples`. Moreover, notebook cells should be run in order so we can't use multiple workers. 

I can do this command instead
`pytest --nbval-lax examples/  -v -m "not l1 and not manual" -n 1 --durations=15`

tripy/examples/resnet50/resnet50.ipynb
farazkh80: Fair enough, I removed it!

tripy/examples/resnet50/resnet50.ipynb
```diff
@@ -0,0 +1,383 @@
...
+    "    def __call__(self, x):\n",
+    "        x = self.convolution(x)\n",
+    "        x = self.normalization(x)\n",
+    "        if self.activation:\n",
+    "            x = tp.relu(x)\n",
+    "        return x\n",
+    "    \n",
+    "class ResNetShortCut(tp.Module):\n",
+    "    def __init__(self, in_channels, out_channels, stride):\n",
+    "        \"\"\"1x1 convolution and batch normalization for input and output channel dimension matching in ResNetBottleNeckLayer.\n",
```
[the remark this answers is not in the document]
farazkh80: did a format with black and also added type annotations for resnet modules for clarity.

.github/workflows/tripy-l0.yml
```diff
@@ -69,7 +69,7 @@ jobs:
         image: ${{ env.l0_image }}
         options: --gpus all -v ${{ github.workspace }}/tripy:/tripy
         run: |
-          pytest --cov=tripy/ --cov-config=.coveragerc tests/ -v -m "not l1 and not manual" -n 4 --durations=15 --ignore tests/performance
+          pytest --nb-test-files --cov=tripy/ --cov-config=.coveragerc tests/ examples/ -v -m "not l1 and not manual" -n 4 --durations=15 --ignore tests/performance
```
pranavm-nvidia, line 72: We should drop this from L0 since it's so slow

tripy/examples/resnet50/resnet50.ipynb
```diff
@@ -0,0 +1,483 @@
...
+    "\n",
+    "fig, axes = plt.subplots(1, len(idxs), figsize=(15, 3))\n",
+    "\n",
+    "# Loop through each index and corresponding subplot\n",
+    "for ax, nid, label in zip(axes, idxs, labels):\n",
+    "    image = dataset[nid][\"Image\"]\n",
+    "    ax.imshow(image)\n",
+    "\n",
+    "    # Pre-process the image for inference\n",
+    "    processed_image = processor(image, return_tensors=\"np\")[\"pixel_values\"]\n",
```
pranavm-nvidia, line 427: Can we return torch GPU tensors here? Then we can avoid a host->device copy.

tripy/examples/resnet50/resnet50.ipynb
```diff
@@ -0,0 +1,483 @@
...
+    "    resnet_classifier,\n",
+    "    args=[tp.InputInfo(input_shape, dtype=tp.float32)],\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Tripy Demo\n",
```
pranavm-nvidia, line 386: We can drop "Tripy` from all the section titles. e.g. this section could be "Running Inference".

tripy/examples/resnet50/resnet50.ipynb
```diff
@@ -0,0 +1,483 @@
...
+    "    for k, v in pretrained_state_dict.items():\n",
+    "        if \"num_batches_tracked\" in k:\n",
+    "            continue  # Skip num_batches_tracked since it is not needed for\n",
+    "            # inference (tp.BatchNorm does not support)\n",
+    "\n",
+    "        assert k in model_state_dict_keys\n",
+    "        model_dict[k] = tp.Parameter(v.contiguous())\n",
+    "\n",
+    "    model.load_state_dict(model_dict)\n",
+    "    return model\n",
```
pranavm-nvidia, line 345: Since we're updating the model in-place, we don't need to return it here

tripy/examples/resnet50/resnet50.ipynb
```diff
@@ -0,0 +1,483 @@
...
+    "## Compile Tripy Model"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 10,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def resnet_classifier(input):\n",
```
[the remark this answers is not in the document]
farazkh80: added comment

tripy/examples/resnet50/resnet50.ipynb
farazkh80: Do you mean "if processor can return GPU tensors" or moving it after processor returns?

tripy/examples/resnet50/resnet50.ipynb
farazkh80: dropped

tripy/examples/resnet50/resnet50.ipynb
farazkh80: I am not sure if that's possible, because `processor` is not a torch model https://github.com/huggingface/transformers/blob/052e652d6d53c2b26ffde87e039b723949a53493/src/transformers/models/convnext/image_processing_convnext.py#L51 and the `preprocess` function does not take any device related args

tripy/examples/resnet50/resnet50.ipynb
farazkh80: done

tripy/examples/resnet50/resnet50.ipynb
pranavm-nvidia: I'm assuming it would still respect [set_default_device](https://pytorch.org/docs/stable/generated/torch.set_default_device.html#torch-set-default-device) though? IIUC it changes the global default for all torch calls

farazkh80: Ok yes, but they do a conversion to `np.array` [here](https://github.com/huggingface/transformers/blob/052e652d6d53c2b26ffde87e039b723949a53493/src/transformers/models/convnext/image_processing_convnext.py#L283) and then back to torch tensor using `torch.from_numpy` which does not respect `set_default_device`. I validated it like below

image

pranavm-nvidia: Ah ok, I guess we can't avoid the copy then

parthchadha:
Let's try to make this notebook a standalone notebook where tripy is installed and notebook can be executed in colab (this can be done in a separate PR).

yizhuoz004:
Would be nice to polish the notebook a bit with more comments at the beginning of each section, e.g.
- Implementation: Maybe change the title to be "Modeling" and add a comment like "Tripy's implementation of ResNet50, see reference: xxx"
- Load Weights to Model: where did the state_dict come from, what are necessary steps to transform the torch tensors.
- Compile Model: can we merge the argmax to `ResNetClassifier`?

farazkh80:
> Would be nice to polish the notebook a bit with more comments at the beginning of each section, e.g.
> 
> * Implementation: Maybe change the title to be "Modeling" and add a comment like "Tripy's implementation of ResNet50, see reference: xxx"
> * Load Weights to Model: where did the state_dict come from, what are necessary steps to transform the torch tensors.

Sounds good, will change some titles and add some comments for each section.

Regarding 

> * Compile Model: can we merge the argmax to `ResNetClassifier`?

including post processing of logits like `argmax` in the model implementation is a little odd and since we want to emphasize on "compiling any tripy related step weather it is pre or post processing", making the `infer function` which does `tp.argmax` after model forward makes more sense.

pranavm-nvidia:
+1 for keeping `argmax` out of the model.

yizhuoz004:
> > Would be nice to polish the notebook a bit with more comments at the beginning of each section, e.g.
> > 
> > * Implementation: Maybe change the title to be "Modeling" and add a comment like "Tripy's implementation of ResNet50, see reference: xxx"
> > * Load Weights to Model: where did the state_dict come from, what are necessary steps to transform the torch tensors.
> 
> Sounds good, will change some titles and add some comments for each section.
> 
> Regarding
> 
> > * Compile Model: can we merge the argmax to `ResNetClassifier`?
> 
> including post processing of logits like `argmax` in the model implementation is a little odd and since we want to emphasize on "compiling any tripy related step weather it is pre or post processing", making the `infer function` which does `tp.argmax` after model forward makes more sense.

I see your point. From a new user's perspective, it might be a little confusing that you are not compiling the model itself. Might be better if we just compile the model, do the argmax in torch, and also explain why (the eager mode pitfalls).

.github/workflows/tripy-l1.yml
```diff
@@ -36,3 +36,10 @@ jobs:
       run: |
         cd /tripy/
         pytest --cov=tripy/ --cov-config=.coveragerc tests/ -v -m "l1 and not manual" -n 4 --durations=15 --ignore tests/performance
+    
+    - name: notebook-test
+      run: |
+        unset PYTHONPATH
+        pip uninstall tripy -y
+        cd /tripy/
```
pranavm-nvidia, line 44: We want to use the release packages here right? Not the TOT build of Tripy.

farazkh80:
L1 passes now https://github.com/NVIDIA/TensorRT-Incubator/actions/runs/12042555141/job/33576419072

@pranavm-nvidia

.github/workflows/tripy-l1.yml
farazkh80: right, they are installed in the first code block of the notebook. Here I uninstall the from source tripy so I can install later in the notebook

tripy/setup.py
```diff
@@ -49,5 +50,10 @@ def generate_stubs(self):
             print(f"Error generating stubs: {e}")
 
 
+default_excludes = list(FlatLayoutPackageFinder.DEFAULT_EXCLUDE) + ["notebooks"]
+
 if __name__ == "__main__":
-    setup(cmdclass={"build_py": BuildPyCommand})
+    setup(
+        cmdclass={"build_py": BuildPyCommand},
+        packages=find_packages(exclude=default_excludes),  # Use the dynamically updated exclude list
```
pranavm-nvidia, line 58: Can you add a comment here explaining why we didn't use `include` in `pyproject.toml`?

tripy/notebooks/resnet50.ipynb
```diff
@@ -0,0 +1,493 @@
...
+    "        \"\"\"Consists of a ResNetShortCut layer for dimension matching and\n",
+    "        three ResNetConvLayers for residual learning.\"\"\"\n",
+    "        super(ResNetBottleNeckLayer, self).__init__()\n",
+    "\n",
+    "        self.shortcut = (\n",
+    "            ResNetShortCut(in_channels, out_channels, stride)\n",
+    "            if in_channels != out_channels or stride != (1, 1)\n",
+    "            else lambda x: x\n",
+    "        )\n",
+    "        self.layer = [\n",
```
pranavm-nvidia, line 168: Should we replace all the lists with `tp.Sequential`?

tripy/notebooks/resnet50.ipynb
```diff
@@ -0,0 +1,493 @@
...
+    "    def __init__(\n",
+    "        self,\n",
+    "        in_channels: int,\n",
+    "        out_channels: int,\n",
+    "        bottleneck_channels: int,\n",
+    "        stride: Tuple[int, int],\n",
+    "    ):\n",
+    "        \"\"\"Consists of a ResNetShortCut layer for dimension matching and\n",
+    "        three ResNetConvLayers for residual learning.\"\"\"\n",
+    "        super(ResNetBottleNeckLayer, self).__init__()\n",
```
pranavm-nvidia, line 161: These can all be shortened to: `super().__init__()`

.github/workflows/tripy-l1.yml
farazkh80: ok added tripy TOT install here too

tripy/setup.py
farazkh80: done

tripy/notebooks/resnet50.ipynb
farazkh80: I think not, because sequential is one big forward call. But usually you want to do

```python
for layer in layers:
    x= layer(x)
```
 
 instead of `x=layers(x)`. It is also good to have both to demonstrate functionality options.

tripy/notebooks/resnet50.ipynb
farazkh80: done

farazkh80:
Should be now @pranavm-nvidia  L1 passed as well

tripy/notebooks/resnet50.ipynb
farazkh80: switched to sequential, since it is the same input for all layers.

pranavm-nvidia (approved):

PR #323, opened 2024-10-30
Move internal changes

shelkesagar29:
This is a combination of the following commits:

```

[compiler/StableHloExt] Create general pattern to allow StableHlo ops to absorb generalizing `tensor.cast`
    
    This change generalizes a rewrite pattern in the `stablehlo-ext-constant-folding` pass
    and exposes the pattern for use in other passes. In many of the early plan segmentation
    passes where types are being refined based on shape and value bounds analyses, it is
    necessary to create `tensor.cast` operations since in general it is not legal to update
    types of SSA values in-place. It is therefore important that we have a robust set of
    patterns to eliminate these `tensor.cast` operations; otherwise they can cause problems
    in the segmentation (e.g. because we don't support converting `tensor.cast` to
    TensorRT like we do with StableHlo ops).

[StablehloToTensorRT] Support additional constant operation types in TensorRT conversion
    
    This change allows for converting `arith.constant` in addition to `stablehlo.constant`
    during the Stablehlo-to-TensorRT conversion. Since we are allowing this one
    additional operation, I didn't generate a whole new pass/pattern set and just
    added an instantiation of the templated constant converter that already exists.

[tensorrt] Fix logic in `tensorrt.convolution` verification routine
    
    The `tensorrt.convolution` verification routine incorrectly was comparing
    shape dimension values without checking if they are dynamic.

[plan] Add integer-range based optimizations to the `materialize-shape-calculations` pass
    
    This change adds additional simplification patterns based on the IntegerRangeAnalysis.
    This is a convenient way to re-use the logic provided in the analysis to refine
    operations like `tensor.dim` into constants where possible.

[plan] Add integer-range based optimizations to the `materialize-shape-calculations` pass
    
    This change adds additional simplification patterns based on the IntegerRangeAnalysis.
    This is a convenient way to re-use the logic provided in the analysis to refine
    operations like `tensor.dim` into constants where possible.

NFC: [compiler] Fix trailing whitespace in Plan tablegen files

NFC: remove use of 'is_pointwise' attribute during `stablehlo.composite` creation
    
    After making the `stablehlo.composite` shape materialization generic, we
    no longer need the `is_pointwise` attribute that was being attached to Q/DQ
    `stablehlo.composite` operations.

[Dialect/Plan] Make the 'stablehlo.composite' shape propagation completely generic
    
    Makes shape propagation in 'plan-materialize-shape-calculations' completely
    generic by simply cloning the composite implementation function body and
    let the shape machinery analyze the body operations. The extra implementation
    operations end up getting erased as long as the shape machinery can propagate through
    them (which means as long as there is no DDS). Otherwise, the extra materialization
    of intermediate values will be retained for shape calculation (which
    may be removed down the line if the composite op is replaced with its
    decomposition and CSE is performed).

[compiler] Fix use-after-free in the `plan-create-closed-regions` pass
    
    A recent change in the `plan-create-closed-regions` pass results in calling
    `op->getLoc()` after `op` has already been replaced/deleted. This change
    fixes that issue, resolving ASAN errors.

```

### DIFF
--- a/mlir-tensorrt/compiler/lib/Conversion/TensorRTToTensorRTRuntime/TensorRTToTensorRTRuntime.cpp
+++ b/mlir-tensorrt/compiler/lib/Conversion/TensorRTToTensorRTRuntime/TensorRTToTensorRTRuntime.cpp
@@ -57,78 +57,6 @@ struct RewriteConstants : public OpRewritePattern<tensorrt::ConstantOp> {
   }
 };
 
-// Helper function to convert both CallOp and CallAllocOp
-static LogicalResult
-convertCallOp(Operation *op, IRRewriter &rewriter,
-              SymbolTableCollection &symbolTable, DataFlowSolver &solver,
-              ModuleOp module,
-              SmallVectorImpl<tensorrt::TensorRTModuleOp> &trtModules) {
-  MLIRContext *ctx = rewriter.getContext();
-  Location loc = op->getLoc();
-  ValueRange inputs;
-  ValueRange outputs;
-  func::FuncOp trtFunc;
-
-  if (auto callOp = dyn_cast<tensorrt::CallOp>(op)) {
-    trtFunc = callOp.getFuncCallee(symbolTable);
-    inputs = callOp.getInputs();
-    outputs = callOp.getOutputs();
-  } else if (auto callAllocOp = dyn_cast<tensorrt::CallAllocOp>(op)) {
-    trtFunc = callAllocOp.getFuncCallee(symbolTable);
-    inputs = callAllocOp.getInputs();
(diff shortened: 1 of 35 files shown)


mlir-tensorrt/compiler/lib/Conversion/TensorRTToTensorRTRuntime/TensorRTToTensorRTRuntime.cpp
```diff
@@ -57,78 +57,6 @@ struct RewriteConstants : public OpRewritePattern<tensorrt::ConstantOp> {
   }
 };
```
christopherbate, line 59: We need to sync OSS to internal, it's dropping OSS changes

PR #324, opened 2024-10-30
[Runtime] Remove incorrect removal of previous entry in map in AllocT…

parthchadha:
…racker::track; track DLPackTensors in RuntimeClient such that deleters can be reset when RuntimeClient is destroyed.

This PR fixes two issues:
1. When you have a dlpack array allocated outside of `mlir-tensorrt` (ex: numpy) and you create memref view of this array, and then create another dlpack tensor from this memref, there was a bug in our existing implementation of `AllocTracker::track`. In `track()` we were unnecessarily reseting the value of already defined entry in map and thereby resting the `externalReferenceCount`. 
2. After fixing ^, and adding the test in `test_create_memref.py` it was found that the `AllocTracker` (from `RuntimeClient) can be destructed before the external allocated tensor is destructed. To fix this issue we now remove the `deleter` of the tracked `dlpack` tensors when `RuntimeClient` is destructed.

Thanks to @jhalakpatel for debugging this issue.


mlir-tensorrt/executor/include/mlir-executor/Runtime/API/API.h
```diff
@@ -907,6 +907,9 @@ class RuntimeSession {
 /// RuntimeClient and be destroyed/deallocated through the appropriate method.
 class RuntimeClient {
 public:
+  /// Add destructor declaration
```
christopherbate, line 910: Nit: comment isn't useful, can be removed

mlir-tensorrt/executor/include/mlir-executor/Runtime/API/API.h
```diff
@@ -969,6 +972,17 @@ class RuntimeClient {
     return pinnedMemoryAllocator;
   }
 
+  /// Track the DLPack tensors in RuntimeClient such that their deleters can be
+  /// reset when RuntimeClient is destructed.
+  void trackDLPackTensor(DLManagedTensor *tensor) {
+    dlPackTensors.insert(tensor);
```
christopherbate, line 978: Please move implementation to CPP file.

mlir-tensorrt/executor/lib/Runtime/API/API.cpp
```diff
@@ -463,8 +466,6 @@ void AllocTracker::track(PointerInfo info) {
     map.insert(std::make_pair(info.ptr, std::move(value)));
     return;
   }
-  untrack(info.ptr);
-  map.insert(std::make_pair(info.ptr, std::move(value)));
```
christopherbate, line 467: It's not clear to me why removing these lines is OK. The value struct that was previously in the map may be different than what is being passed into this function.

mlir-tensorrt/executor/include/mlir-executor/Runtime/API/API.h
```diff
@@ -977,6 +991,7 @@ class RuntimeClient {
   PinnedMemoryAllocator pinnedMemoryAllocator;
   AllocTracker allocTracker;
   ResourceTracker resourceTracker;
+  std::set<DLManagedTensor *> dlPackTensors;
```
christopherbate, line 994: Please use "llvm::DenseMap". Actually, if we do not need to do frequent lookups, then just tracking using a vector would be OK too. Depends on how often this map is going to be used. Check the notes in the LLVM programmer manual guide to guide the specific container choice.

mlir-tensorrt/executor/lib/Runtime/API/API.cpp
```diff
@@ -1067,6 +1072,13 @@ Status RuntimeClient::copyToHost(const MemRefValue &deviceMemRef,
 #endif
 }
 
+RuntimeClient::~RuntimeClient() {
+  // Reset the deleter for all tracked dlPackTensors
+  for (auto *tensor : dlPackTensors) {
+    tensor->deleter = nullptr;
```
christopherbate, line 1078: Deleting the deleter seems incorrect, since other memory not associated with the client gets freed there (eg strides and shape arrays). Instead, we should probably set the manager context handle to null here, then check for nullity in the deleter?

mlir-tensorrt/test/python/mlir_tensorrt_runtime/test_create_memref.py
```diff
@@ -496,6 +496,29 @@ def create_memref_from_dlpack(arr, module):
...
+    print(f"-- Memref shape: {memref.shape}")
+    print(f"-- Memref dtype: {memref.dtype}")
+    print(f"-- {module.__name__}.from_dlpack(): {dl}")
+    return memref, dl
+
+
+print("Test keeping np.array, dlpack, memref view and dlpack view alive.")
+arr = np.array([1, 2, 3, 4], dtype=np.int32)
+dl = arr.__dlpack__()
+memref1, dl1 = create_memref_dlpackview_from_dlpack(dl, np)
```
christopherbate, line 511: OK I think I understand... in Python we don't have control over the destruction order, so these could be destroyed prior to client. The test should be more explicit here then. Create a new client then explicitly delete it prior to deleting the memref object.

If this is indeed the issue, then we need a more robust fix to ensure the correct destruction order? I guess we somehow already side step this issue with non-dlpack memrefs

jhalakpatel:
> …racker::track; track DLPackTensors in RuntimeClient such that deleters can be reset when RuntimeClient is destroyed.

Looks like PR description got truncated. May be add a shorter commit message and a separate PR description.

mlir-tensorrt/test/python/mlir_tensorrt_runtime/test_create_memref.py
parthchadha: @christopherbate @jhalakpatel  I have updated the PR issue description with the 2 issues that this PR solves. 
I can fix the test to add more explicit destruction order. We do not really need to correct the destruction order since we are only bookkeeping the external allocated `dlpack` tensors and if `RuntimeClient` is destructed early it shouldn't affect the behavior of these tensors.

mlir-tensorrt/test/python/mlir_tensorrt_runtime/test_create_memref.py
```diff
@@ -496,6 +496,34 @@ def create_memref_from_dlpack(arr, module):
...
+    def create_memref_dlpackview_from_dlpack(dl, module):
+        memref = new_client.create_memref_view_from_dlpack(dl)
+        dl = module.from_dlpack(memref)
+        print(f"-- Memref shape: {memref.shape}")
+        print(f"-- Memref dtype: {memref.dtype}")
+        print(f"-- {module.__name__}.from_dlpack(): {dl}")
+        return memref, dl
+
+    memref1, dl1 = create_memref_dlpackview_from_dlpack(dl, np)
+    memref2, dl2 = create_memref_dlpackview_from_dlpack(dl, np)
```
christopherbate, line 511: before returning, still need to explicitly delete and invoke gc collect?

```
    del new_client
    gc.collect()
    return memref1, dl1, memref2, dle2
```

mlir-tensorrt/executor/lib/Runtime/API/API.cpp
parthchadha: Explained in the PR description.

christopherbate: But there's still a problem by dropping these lines then, which is explained in the comment above starting on line 449. 

I think what we want here is something like `if( existingInfo.isExternallyManged && info.isExternallyManged ) return)`, otherwise retain the original behavior.

mlir-tensorrt/executor/lib/Runtime/API/API.cpp
```diff
@@ -463,6 +466,22 @@ void AllocTracker::track(PointerInfo info) {
     map.insert(std::make_pair(info.ptr, std::move(value)));
     return;
   }
+
+  auto existingReferenceCount = map.at(info.ptr)->externalReferenceCount.load();
+
+  // If exisiting value in map is internally maanged and not yet released
+  // internally, we can just increment the reference count for the exisiting
```
jhalakpatel, line 473: This comment looks incomplete based on the current implementation. can we instead explain the motivation to increase the reference count rather than explain what code does which seems self-explanatory.

mlir-tensorrt/executor/lib/Runtime/API/API.cpp
```diff
@@ -463,6 +466,22 @@ void AllocTracker::track(PointerInfo info) {
...
+  if (get(info.ptr).isInternallyManaged() &&
+      !map.at(info.ptr)->releasedInternally && existingReferenceCount > 0 &&
+      info.isExternallyManaged()) {
+    ++map.at(info.ptr)->externalReferenceCount;
+    return;
+  }
+
+  // In the default case when there is an exisiting value in map, we reuse the
+  // exisiting reference count for the new pointer.
+  value->externalReferenceCount = existingReferenceCount;
```
jhalakpatel, line 484: are we expecting the incoming pointer reference count to be 0 here? Should we assert that invariant?

christopherbate (approved):
Copying discussion from slack:

OK It passes internal tests. After looking through how XLA handles these issues, I think we will ultimately re-work this in a two-step approach:
1) ensuring all MemRef value Python objects retain a handle to the RuntimeClient Python object. that will ensure that RuntimeClient is not garbage collected by Python when clients are still live.
2) ensuring all DLPack objects retain a handle to the Python MemRef value. This is tricky to implement, but XLA does a clever trick with how they allocate the DLManagedTensor object to achieve this.
1 and 2 together should enable removing the "dlpack tracking" functionality that you are adding to the client right now.
Anyway, none of this should block merging, so please go ahead.

PR #325, opened 2024-10-31
Removes useless tests

pranavm-nvidia:
Removes `test_str` and `test_mlir` from FlatIR tests. These tests are extremely fragile because the MLIR we generate can change very easily. We already have functional and performance tests to ensure there are no regressions, so there is no utility in ensuring we generate exactly the same MLIR.


parthchadha (approved):
Agreed, there was very little value but extremely high overhead that these tests were adding. Though, when we switch to TensorRT dialect it might be valuable to add them back since these tests can be really fast and act as real unit tests.

slyubomirsky:
I think it would be best to replace them with lightweight tests that validate very basic properties/invariants. Matching exact strings is far too brittle. Since we will switch dialects, eliminating the current ones makes sense.

slyubomirsky (approved):

PR #328, opened 2024-10-31
Makes rank inference happen entirely in the frontend

pranavm-nvidia:
- Removes FlatIR functions. The function deduplication logic is currently broken in some edge cases. Once we switch to the TRT dialect, functions will not be required anyway, so this commit disables functions for now to avoid functional issues.
    
- Removes `infer_len` and updates `infer_rank` to no longer require going all the way to MLIR. Instead, rank inference is done entirely in the frontend.

- Renames a test file to use correct casing (`test_binaryelementwise.py` -> `test_binary_elementwise.py`)

- Removes side effects of running `infer_rank` in `iota`, `transpose`, and `reduce` functions.


tripy/tripy/frontend/trace/ops/slice.py
```diff
@@ -30,17 +30,11 @@
 
 @dataclass(repr=False)
 class Slice(BaseTraceOp):
+    infer_rank = op_utils.InferRankPolicies.same_as_input()
```
slyubomirsky, line 33: Maybe I'm missing something obvious, but is this actually correct? If the slice is by a single index, then the result has a lower rank than the input.

```
x = Tensor([1]) # rank 1
y = x[0] # rank 0
```

tripy/tripy/frontend/trace/ops/iota.py
```diff
@@ -31,25 +32,7 @@ class Iota(BaseTraceOp):
...
-                assert out_shape[0] >= 0, f"incorrect shape computation {out_shape}"
-                self.output_rank = out_shape[0]
-            else:
-                self.output_rank = self.inputs[0].shape[0]
-
-        # Iota requires inputs[0] to be statically shaped
-        if self.inputs[0].shape is None:
-            self.inputs[0].shape = (self.output_rank,)
-
-        if self.dim < 0:
```
parthchadha, line 50: You might have adjust `self.dim` in some function now.

tripy/tripy/frontend/trace/ops/slice.py
pranavm-nvidia: Yes, correct, but we do the squeezing separately:
```py
    squeeze_dims = []
    for i, idx in enumerate(index):
        if isinstance(idx, (tuple, list)):
            raise NotImplementedError("Gather is not supported")
        if isinstance(idx, int):
            squeeze_dims.append(i)
    if squeeze_dims:
        out = squeeze(out, make_tuple(squeeze_dims))
```
The output of the `Slice` _trace_ operation, is always the same rank as the input, even though the frontend API can remove dimensions.

tripy/tripy/frontend/trace/ops/resize.py
```diff
@@ -33,8 +34,7 @@ class Resize(BaseTraceOp):
     scales: Optional[Sequence[float]]
     align_corners: bool
 
-    def infer_rank(self):
-        self.outputs[0].rank = self.inputs[0].rank
+    infer_rank = op_utils.InferRankPolicies.same_as_input()
```
parthchadha, line 37: Isn't output rank same as length of second input (shape tensor) for resize?

pranavm-nvidia: Resize can't change the rank of the input (only its shape) and the second input may not always be materialized at this point (e.g. when using scales). So it's just easier to get the rank from the first input.

tripy/tripy/frontend/trace/ops/utils.py
```diff
@@ -42,29 +42,43 @@ def is_minus_one(arg):
...
+class InferRankPolicies:
+    def same_as_input(idx=0):
+        def impl(self):
+            self.outputs[0].rank = self.inputs[idx].rank
+
+        return impl
+
+    def same_as_shape_of_shape_input(idx=0):
+        def impl(self):
+            # Expects that the first input is a shape tensor.
```
yizhuoz004, line 70:
```suggestion
            # Expects that the inputs[idx] is a shape tensor.
```

yizhuoz004 (approved):

yizhuoz004:
Shall we test on nanoGPT to see the compilation time?

tripy/tripy/frontend/trace/ops/utils.py
pranavm-nvidia: Actually don't even need the comment since the assertion says the exact same thing. Will remove.

pranavm-nvidia:
> Shall we test on nanoGPT to see the compilation time?

@yizhuoz004 Before all my shape refactoring changes, the time was:
```
Compilation took 31.865108418278396 seconds.
```
For the first shape refactor PR, I had to undo a few of the shape inference optimizations, so the time on TOT main right now is:
```
Compilation took 685.2806877028197 seconds.
```
After this change, which makes the shape inference optimizations unnecessary in the first place, the time is:
```
Compilation took 12.440084163099527 seconds.
```
So almost 3x better than where we started.

parthchadha (approved):

tripy/tripy/frontend/trace/ops/matmul.py
```diff
@@ -16,87 +16,32 @@
...
-                    raise ValueError("Ranks must be at least 1")
-
-                def is_vector(rank):
-                    return rank == 1
-
-                def is_matrix(rank):
-                    return rank >= 2
+            out_rank = 0
+        elif self.inputs[0].rank == 1 or self.inputs[1].rank == 1:
+            out_rank = 1
```
jhalakpatel, line 42: should we add an assertion check for unsupported input ranks?

pranavm-nvidia: Yes, there's already one [here](https://github.com/NVIDIA/TensorRT-Incubator/pull/328/files#diff-143063c187b705db2521def8d47c2e26f18d468dadae71de1ec02272b410efdfR186).

jhalakpatel (approved):

tripy/tripy/frontend/trace/ops/matmul.py
pranavm-nvidia: Oh that's handled before the conditional:
```
out_rank = max(self.inputs[0].rank, self.inputs[1].rank)
```
I could move it to an else block if preferred.

slyubomirsky (approved):

Issue #311, opened 2024-10-28
Make `preprocess_args` in `convert_to_tensors` work for variadic arguments


pranavm-nvidia:
We need to enable preprocessing for variadic arguments.

pranavm-nvidia:
Search for `TODO (#311)` in the code for details.

slyubomirsky:
Implemented.

PR #329, opened 2024-10-31
[Tripy] Handle variadic arguments in `preprocess_args` for the `convert_to_tensors` decorator

slyubomirsky:
Addresses issue #311.

### DIFF
--- a/tripy/tripy/frontend/utils.py
+++ b/tripy/tripy/frontend/utils.py
@@ -18,7 +18,7 @@
 import functools
 import inspect
 from collections import deque
-from typing import Callable, List, Optional, Set, Union
+from typing import Callable, Dict, List, Optional, Sequence, Set, Tuple, Union
 
 from tripy import utils
 from tripy.common.exception import raise_error
@@ -166,7 +166,8 @@ def convert_to_tensors(
         preprocess_args: A callback used to preprocess arguments before potential conversion. If provided,
             this is always called, regardless of whether the decorator actually needed to perform conversion.
             This will be called with all arguments that were passed to the decorated function and should
-            return a dictionary of all updated arguments.
+            return a dictionary of all updated arguments. For a variadic arg, the dictionary entry for the name
+            should have a list of all the updated values.
     """
 
     def impl(func):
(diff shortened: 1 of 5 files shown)


tripy/tripy/frontend/utils.py
```diff
@@ -191,14 +192,32 @@ def wrapper(*args, **kwargs):
...
+                # for the variadic arg to handle variadic args when preprocessing.
+                def find_variadic_name_and_start_idx(all_args: Sequence) -> Optional[Tuple[str, int]]:
+                    encountered_names = {}
+                    for i, (name, _) in enumerate(all_args):
+                        if name in encountered_names:
+                            return name, encountered_names[name]
+                        if name not in encountered_names:
+                            encountered_names[name] = i
+
+                    return None
```
pranavm-nvidia, line 207: We actually have this information already when we call `get_positional_arg_names` in `merge_function_arguments`. Could we just return it from there?

slyubomirsky: I added it into there. Maybe a little odd to throw in an extra return value but it avoids an extra loop.

tripy/tripy/utils/utils.py
```diff
@@ -447,22 +448,26 @@ def get_positional_arg_names(func, *args):
         arg_names.append(name)
 
     # For all variadic positional arguments, assign the name of the variadic group.
-    arg_names.extend([varargs_name] * (len(args) - len(arg_names)))
-    return list(zip(arg_names, args))
+    num_variadic_args = len(args) - len(arg_names)
+    variadic_start_idx = len(arg_names)
+    arg_names.extend([varargs_name] * num_variadic_args)
+    return list(zip(arg_names, args)), (varargs_name, variadic_start_idx) if num_variadic_args > 0 else None
```
pranavm-nvidia, line 455: I was thinking that we could do something more like:
```py
    variadic_start_idx = None
    for index, (name, param) in enumerate(signature.parameters.items()):
        if param.kind == inspect.Parameter.VAR_POSITIONAL:
            # Positional arguments cannot follow variadic positional arguments
            # (they would just be absorbed into the variadic argument).
            variadic_start_idx = index
            varargs_name = name
            break
```

slyubomirsky: Ah, even simpler

pranavm-nvidia (approved):

PR #330, opened 2024-10-31
Miscellaneous fixes

pranavm-nvidia:
- Updates plugin test to check for shapes

- Makes process_dim a regular function instead of a decorator. Making `process_dim` a decorator is overkill and complicates the stack infomation unnecessarily. This change makes it a regular function so that it can be used more broadly.

- Removes dead code

- Fixes a typo in the profiling section of the README

- Makes datatype constraints tests significantly faster by not evaluating inputs.

### DIFF
--- a/tripy/tests/README.md
+++ b/tripy/tests/README.md
@@ -36,7 +36,7 @@ tests together.
 For example, to profile L0 tests, run:
 
 ```bash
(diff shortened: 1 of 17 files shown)


yizhuoz004 (approved):

slyubomirsky (approved):
Very good to remove dead code.

Issue #310, opened 2024-10-28
Fix or remove  `skip_num_stack_entries` in `convert_to_tensors`


pranavm-nvidia:
When `skip_num_stack_entries` is set, the AST that we parse doesn't necessarily match the signature of the function that is actually being decorated. This can cause bugs because the arguments that we expect to see aren't the ones that we actually see. 

`skip_num_stack_entries` is mainly used in cases where we want to process the argument before it has been converted to a tensor/shape, and we can do this through the `preprocess_args` option in the decorator. In `slice`, we use it to get stack information for the individual slice index arguments, but we could perhaps do that manually.

slyubomirsky:
`slice` is the only use of it, so handling that separately might be the right way to go. I will see how feasible that would be.

pranavm-nvidia:
@slyubomirsky Should I close this issue now?

PR #333, opened 2024-11-03
[Tripy] Eliminate need for `skip_num_stack_entries` argument in `convert_to_tensors`

slyubomirsky:
Addresses issue #310. The only use of `skip_num_stack_entries` was for `slice_helper` and addressing this issue in a systematic manner would likely require building in many hacks and assumptions, so the approach here is just to manually override the stack information in that one function.

### DIFF
--- a/tripy/tripy/frontend/trace/ops/slice.py
+++ b/tripy/tripy/frontend/trace/ops/slice.py
@@ -258,4 +258,31 @@ def __getitem__(
 
 @frontend_utils.convert_to_tensors()
 def slice_helper(tensor, *slice_params: TensorLike):
+    from tripy.utils import get_arg_candidate_column_offsets
+
+    # The default behavior of convert_to_tensors will not add the correct column info to the slice params
+    # because this call occurs *inside* the overridden call to __getitem__, so we adjust the column info manually.
+    arg_names = ["tensor"] + ["slice_params"] * len(slice_params)
+    for arg_index, arg in enumerate(slice_params):
+        # Look for the call to __getitem__. We need to go one stack frame beyond to get to the *user* call of __getitem__.
+        frame_index = -1
+        for idx, source_info in enumerate(arg.stack_info):
+            if source_info._dispatch_target == "__getitem__":
+                frame_index = idx + 1
+                break
+
+        # convert_to_tensors should have taken care of this for us
+        assert frame_index >= 0, "No call to the __getitem__ dispatch found"
+
+        source_info = arg.stack_info[frame_index]
+
+        # Note: arg_index does not account for the positional arg, hence we add 1 for the index argument
+        candidates = get_arg_candidate_column_offsets(
+            source_info.code, 1 + arg_index, 1, "__getitem__", False, arg_names
+        )
+
+        # Now we can set the column range correctly
+        if len(candidates) == 1:
+            source_info.column_range = candidates[0]
+
     return Slice.build(inputs=[tensor, *slice_params])
--- a/tripy/tripy/frontend/trace/ops/slice.py
+++ b/tripy/tripy/frontend/trace/ops/slice.py
@@ -266,14 +266,15 @@ def slice_helper(tensor, *slice_params: TensorLike):
     # Look for the stack frame index to __getitem__. We need to go one stack frame beyond to get to the *user* call of __getitem__.
     # It will be the same for all the slice params
     frame_index = -1
-    if slice_params:
-        for idx, source_info in enumerate(slice_params[0].stack_info):
-            if source_info._dispatch_target == "__getitem__":
-                frame_index = idx + 1
-                break
+    assert slice_params
 
-        # convert_to_tensors should have taken care of this for us
-        assert frame_index >= 0, "No call to the __getitem__ dispatch found"
+    for idx, source_info in enumerate(slice_params[0].stack_info):
+        if source_info._dispatch_target == "__getitem__":
+            frame_index = idx + 1
+            break
+
+    # convert_to_tensors should have taken care of this for us
+    assert frame_index >= 0, "No call to the __getitem__ dispatch found"
 
     arg_names = ["tensor"] + ["slice_params"] * len(slice_params)
     for arg_index, arg in enumerate(slice_params):
--- a/tripy/tripy/frontend/trace/ops/slice.py
+++ b/tripy/tripy/frontend/trace/ops/slice.py
@@ -262,11 +262,12 @@ def slice_helper(tensor, *slice_params: TensorLike):
 
     # The default behavior of convert_to_tensors will not add the correct column info to the slice params
     # because this call occurs *inside* the overridden call to __getitem__, so we adjust the column info manually.
-    arg_names = ["tensor"] + ["slice_params"] * len(slice_params)
-    for arg_index, arg in enumerate(slice_params):
-        # Look for the call to __getitem__. We need to go one stack frame beyond to get to the *user* call of __getitem__.
-        frame_index = -1
-        for idx, source_info in enumerate(arg.stack_info):
+
+    # Look for the stack frame index to __getitem__. We need to go one stack frame beyond to get to the *user* call of __getitem__.
+    # It will be the same for all the slice params
+    frame_index = -1
+    if slice_params:
+        for idx, source_info in enumerate(slice_params[0].stack_info):
             if source_info._dispatch_target == "__getitem__":
                 frame_index = idx + 1
                 break
@@ -274,6 +275,8 @@ def slice_helper(tensor, *slice_params: TensorLike):
         # convert_to_tensors should have taken care of this for us
         assert frame_index >= 0, "No call to the __getitem__ dispatch found"
 
+    arg_names = ["tensor"] + ["slice_params"] * len(slice_params)
+    for arg_index, arg in enumerate(slice_params):
         source_info = arg.stack_info[frame_index]
 
         # Note: arg_index does not account for the positional arg, hence we add 1 for the index argument
--- a/tripy/tripy/frontend/trace/ops/slice.py
+++ b/tripy/tripy/frontend/trace/ops/slice.py
@@ -256,8 +256,6 @@ def __getitem__(
     return out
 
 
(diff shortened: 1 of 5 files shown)


tripy/tripy/frontend/utils.py
```diff
@@ -84,11 +85,23 @@ def _add_column_info(arg, arg_index, is_kwarg, num_positional, func_name, skip_n
 
     # Find the first caller of this function that is NOT the function registry.
     # Also save the last dispatch target we see.
+
+    # Start from the registry. It will always be present except for tests,
+    # since the decorator is intended only for overrides of magic functions.
+    # This check supports cases like slice_helper, where the decorated function
+    # is used *inside* the override and hence the wrapped call would come *before*
+    # the call from the registry.
+    REGISTRY_STACK_DEPTH = WRAPPER_STACK_DEPTH
```
pranavm-nvidia, line 94: Do we still need to find `WRAPPER_STACK_DEPTH` first?

slyubomirsky: It would only be needed in our test cases, unless we want to add those to the function registry.

pranavm-nvidia: I would slightly prefer adding hacks into the tests rather than into Tripy proper.

tripy/tripy/frontend/utils.py
```diff
@@ -84,11 +85,23 @@ def _add_column_info(arg, arg_index, is_kwarg, num_positional, func_name, skip_n
 
     # Find the first caller of this function that is NOT the function registry.
     # Also save the last dispatch target we see.
+
+    # Start from the registry. It will always be present except for tests,
```
pranavm-nvidia, line 89: One more place where it could be missing is methods of classes. If that becomes an issue, then maybe we could update the function registry to automatically discover and wrap methods too.

tripy/tripy/frontend/utils.py
slyubomirsky: I'll change the tests, in that case.

tripy/tripy/frontend/utils.py
slyubomirsky: Would there be an example of a class method for which we would use the decorator? I thought we were only using it for overrides of magic methods.

tripy/tripy/frontend/utils.py
slyubomirsky: Hm, it seems my assumption that it's for overrides is incorrect. `convert_to_tensors` is also used on `full`, `full_like`, `iota`, `quantize`, `dequantize`, `expand`, `reshape`, and `resize`, though these appear in a function registry via `export.public_api`, so the check still succeeds.

slyubomirsky: The code actually proves to be necessary in the following case: If `tensor_from_shape_like` is used by the decorator, that may call `concatenate`, which is in the function registry due to the use of `export.public_api`. Without the earlier loop, the inner call to `concatenate` is selected first and the stack entry is chosen. This arises in various test cases that use `tensor_initializers`, e.g., `test_unsqueeze`. I will see if that case can be excluded cleanly.

tripy/tripy/frontend/utils.py
pranavm-nvidia: Since it also handles conversion to shapes now, there could be cases in the future where we want to use it on a method. Could you add a note in the docstring of the decorator mentioning the assumptions it makes about the call stack?

tripy/tripy/frontend/utils.py
pranavm-nvidia: I think skipping stack entries is fundamentally broken. Unless the function signature is the same, it's difficult to reliably determine column information. Could we just update the slice implementation to add column information without the help of this decorator and then disallow skipping stack entries here?

slyubomirsky: Okay, I'll keep the old behavior of checking for the call to the decorator.

tripy/tripy/frontend/utils.py
slyubomirsky: Hopefully it won't be necessary if we remove these hacks per the above discussion, will note whatever assumptions ultimately remain.

tripy/tripy/frontend/utils.py
slyubomirsky: Updated, probably fewer hacks needed overall

tripy/tripy/frontend/trace/ops/slice.py
```diff
@@ -256,8 +256,33 @@ def clamp_bound(bound: Union[int, Tensor]) -> Union[int, Tensor]:
...
+    # The default behavior of convert_to_tensors will not add the correct column info to the slice params
+    # because this call occurs *inside* the overridden call to __getitem__, so we adjust the column info manually.
+    arg_names = ["tensor"] + ["slice_params"] * len(slice_params)
+    for arg_index, arg in enumerate(slice_params):
+        # Look for the call to __getitem__. We need to go one stack frame beyond to get to the *user* call of __getitem__.
+        frame_index = -1
+        for idx, source_info in enumerate(arg.stack_info):
+            if source_info._dispatch_target == "__getitem__":
+                frame_index = idx + 1
+                break
```
pranavm-nvidia, line 272: I think we can lift this out of the loop right? The frame for `__getitem__` should be the same for all the arguments to slice.

slyubomirsky: Oh I had overlooked that. That's correct.

tripy/tripy/frontend/trace/ops/slice.py
```diff
@@ -256,8 +256,36 @@ def clamp_bound(bound: Union[int, Tensor]) -> Union[int, Tensor]:
...
 def slice_helper(tensor, *slice_params: TensorLike):
+    from tripy.utils import get_arg_candidate_column_offsets
+
+    # The default behavior of convert_to_tensors will not add the correct column info to the slice params
+    # because this call occurs *inside* the overridden call to __getitem__, so we adjust the column info manually.
+
+    # Look for the stack frame index to __getitem__. We need to go one stack frame beyond to get to the *user* call of __getitem__.
+    # It will be the same for all the slice params
+    frame_index = -1
+    if slice_params:
```
pranavm-nvidia, line 269: What's the case where there are no `slice_params`?

tripy/tripy/frontend/trace/ops/slice.py
```diff
@@ -256,8 +256,36 @@ def clamp_bound(bound: Union[int, Tensor]) -> Union[int, Tensor]:
     return out
 
 
-# Because the helper is called inside another function, we need to skip one entry in the call stack to find
-# the original call to user code.
-@frontend_utils.convert_to_tensors(skip_num_stack_entries=1)
+@frontend_utils.convert_to_tensors()
 def slice_helper(tensor, *slice_params: TensorLike):
```
pranavm-nvidia, line 260: I'm wondering if it would be better to do the conversion to tensors and the stack info in the caller? The reason `skip_num_stack_entries` is unreliable is that the signature of the API can differ from that of the decorated function, which breaks our AST parsing. By still having the helper, it seems like we would have the same problem.

tripy/tripy/frontend/trace/ops/slice.py
slyubomirsky: I'll change it to an assert. I don't think it's possible for it to be empty but it's best to avoid an indexing error in case it somehow happens.

tripy/tripy/frontend/trace/ops/slice.py
slyubomirsky: Have you run into a case where that's actually happened? The error messages in the test cases generate correctly.

We could indeed get rid of the decorator altogether and handle everything directly in `__getitem__`. The reason for having the helper is that the decorator normally would not be able to handle `slice` inputs and we would have to preprocess them. In any case, we do need to know which inputs are `slice`s and which ones are `int`s to determine things like distinguishing between a slice of size 1 (returns a list of length 1) and a single index (returns a single entry).

slyubomirsky: In this case, the code that would be parsed would be the user's subscripted tensor, i.e., the subscript AST node, whose signature would not match that of our override for `__getitem__` regardless. We could pull in the portions of the code from the decorator that convert values into tensors, but it would not be much different in principle from what the decorator is already doing.

pranavm-nvidia: > Have you run into a case where that's actually happened? 

Yes, but not in slice. I made some fixes, but we used to have some APIs that used decorated helpers with different arguments than the public facing functions. The stack information would point to completely wrong things in those cases. I'm fine with keeping the decorator if it works fine for slice and as long as our testing for it is robust (which I think it is).

pranavm-nvidia (approved):

tripy/tripy/utils/ast.py
```diff
@@ -130,12 +130,13 @@ def index_into_expr(node: ast.expr, index: int) -> ast.expr:
                 return node
 
             # If we have multiple dimensions specified, then we have a tuple of slices.
-            # Indices are given in as a list of start, stop, step
+            # NOTE: We subtract num_positional from the index because the slice arguments would
+            # be passed as *variadic arguments* to slice_helper and so would come after the positional argument
             if isinstance(node.slice, ast.Tuple):
-                element = node.slice.elts[index // 3]
-                arg_node = index_into_expr(element, index % 3)
+                element = node.slice.elts[(index - num_positional) // 3]
```
pranavm-nvidia, line 136: Why was it working before without the offset?

slyubomirsky: I changed it in two places. We were subtracting one from index to make up for the positional arg when we called `index_into_expr` previously, but I think it makes more sense to include the positional arg in the function call and adjust inside `index_into_expr` instead.

tripy/tripy/frontend/trace/ops/slice.py
slyubomirsky: Thanks for the information. Definitely a good reason to have detailed tests for error cases.

PR #340, opened 2024-11-06
Convert state_dict param, add warnings for missing/unexpected keys

yizhuoz004:


tripy/tripy/frontend/module/module.py
```diff
@@ -203,6 +203,15 @@ def find_module(module: Union[Module, List, Dict], sub_strs: List[str]):
                     module = operator.attrgetter(child_name)(module)
             return module
 
+        expected_keys = set(self.state_dict())
```
pranavm-nvidia, line 206: Wouldn't this be a `Set[Tuple[str, Parameter]]`? Or does it only get the keys?

tripy/tripy/frontend/module/module.py
```diff
@@ -203,6 +203,15 @@ def find_module(module: Union[Module, List, Dict], sub_strs: List[str]):
...
+        provided_keys = set(state_dict)
+        missing_keys = expected_keys - provided_keys
+        unexpected_keys = provided_keys - expected_keys
+        if len(missing_keys) > 0:
+            logger.warning("Missing key(s) in state_dict: {}. ".format(", ".join(f'"{k}"' for k in missing_keys)))
+        if len(unexpected_keys) > 0:
+            logger.warning("Unexpected key(s) in state_dict: {}. ".format(", ".join(f'"{k}"' for k in unexpected_keys)))
```
pranavm-nvidia, line 213: Should we use `raise_error` instead?
```suggestion
        if missing_keys:
            logger.warning("Missing key(s) in state_dict: {}. ".format(", ".join(missing_keys)))
        if unexpected_keys:
            logger.warning("Unexpected key(s) in state_dict: {}. ".format(", ".join(unexpected_keys)))
```

tripy/tripy/frontend/module/module.py
yizhuoz004: It only get the keys, it's same as set(dictionary.keys())

tripy/tripy/frontend/module/module.py
yizhuoz004: Seems torch will raise an error, I'll use raise_error too.

yizhuoz004: Added a `strict` option, similar as torch. If strict is True (which is by default), error will be raised. Otherwise the `load_state_dict` function will only load the valid keys, and return the `missing_keys, unexpected_keys`.

tripy/tripy/frontend/module/module.py
```diff
@@ -156,13 +156,20 @@ def __init__(self):
 
         return state_dict
 
-    def load_state_dict(self, state_dict: Dict[str, Parameter]) -> None:
+    def load_state_dict(self, state_dict: Dict[str, Parameter], strict: bool = True) -> Tuple[Set[str]]:
```
pranavm-nvidia, line 159:
```suggestion
    def load_state_dict(self, state_dict: Dict[str, Parameter], strict: bool = True) -> Tuple[Set[str], Set[str]]:
```

tripy/tripy/frontend/module/module.py
```diff
@@ -156,13 +156,20 @@ def __init__(self):
...
         r"""
         Loads parameters from the provided ``state_dict`` into the current module.
         This will recurse over any nested child modules.
 
         Args:
             state_dict: A dictionary mapping names to parameters.
+            strict: If True, keys in ``state_dict`` must exactly match those in this module, if not,
```
pranavm-nvidia, line 166:
```suggestion
            strict: If True, keys in ``state_dict`` must exactly match those in this module. If not,
```

tripy/tripy/frontend/module/module.py
```diff
@@ -203,7 +210,19 @@ def find_module(module: Union[Module, List, Dict], sub_strs: List[str]):
...
             return module
 
+        expected_keys = set(self.state_dict())
+        provided_keys = set(state_dict)
+        missing_keys = expected_keys - provided_keys
+        unexpected_keys = provided_keys - expected_keys
+        if strict and (missing_keys or unexpected_keys):
+            raise_error(
+                "state_dict is incompatible.",
+                f"Got missing keys=[{', '.join(missing_keys)}], unexpected_keys=[{', '.join(unexpected_keys)}].",
```
pranavm-nvidia, line 220: We can make the error a little bit nicer. Also we don't need the `join` since sets will be printed that way anyway:
```suggestion
            details = []
            if missing_keys:
                details.append(f"Missing keys: {missing_keys}\n")
            if unexpected_keys:
                details.append(f"Unexpected keys: {unexpected_keys}")
            raise_error(
                "state_dict is incompatible.",
                details,
```

pranavm-nvidia (approved):

Issue #341, opened 2024-11-06
Function Registry should do type checking on variadic arguments


pranavm-nvidia:
Currently, the function registry doesn't perform type checking on variadic arguments. We should update it so it can handle variadic arguments as well.

slyubomirsky:
Implemented/fixed.

PR #345, opened 2024-11-07
[Tripy] Check type annotations for variadic arguments in the function registry

slyubomirsky:
Addresses issue #341.

### DIFF
--- a/tripy/tests/test_function_registry.py
+++ b/tripy/tests/test_function_registry.py
@@ -255,12 +255,22 @@ class TestFunctionRegistry:
 
     def test_variadic_positional_args(self, registry):
         @registry("test")
-        def func(*args: List[Any]):
+        def func(*args: Any):
             return sum(args)
 
         assert registry["test"](1.0) == 1.0
         assert registry["test"](1.0, 2.0, 3.0) == 6.0
         assert registry["test"]() == 0
+
+    def test_variadic_positional_and_keyword_args(self, registry):
+        # ensure the interaction succeeds
+        @registry("test")
+        def func(a: int, *args: int, b: int, c: int):
+            return a + sum(args) + b + c
+
+        assert registry["test"](3, b=1, c=2) == 6
+        assert registry["test"](3, 4, b=1, c=2) == 10
+        assert registry["test"](3, 4, 5, b=1, c=2) == 15
 
     def test_variadic_keyword_args(self, registry):
         @registry("test")
--- a/tripy/tests/test_function_registry.py
+++ b/tripy/tests/test_function_registry.py
@@ -265,12 +265,12 @@ class TestFunctionRegistry:
     def test_variadic_positional_and_keyword_args(self, registry):
         # ensure the interaction succeeds
         @registry("test")
-        def func(a: int, *args: int, b: int, c: int):
-            return a + sum(args) + b + c
+        def func(a: int, *args: int, b: float, c: str):
+            return a + sum(args) + int(b) + len(c)
 
-        assert registry["test"](3, b=1, c=2) == 6
-        assert registry["test"](3, 4, b=1, c=2) == 10
-        assert registry["test"](3, 4, 5, b=1, c=2) == 15
+        assert registry["test"](3, b=1.0, c="ab") == 6
+        assert registry["test"](3, 4, b=1.0, c="ab") == 10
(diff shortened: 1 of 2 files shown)


tripy/tests/test_function_registry.py
```diff
@@ -469,6 +471,30 @@ def func(n: Sequence[Union[int, float]]) -> int:
...
+            Could not find an implementation for function: 'test'.
+                Candidate overloads were:
+
+                --> \x1b\[38;5;3m{__file__}\x1b\[0m:[0-9]+ in \x1b\[38;5;6mfunc\(\)\x1b\[0m
+                      \|
+                  [0-9]+ \|         def func\(a: int, \*args: int\) \-> int:
+                  [0-9]+ \|     \.\.\.
+                      \|\s
+
+                Not a valid overload because: For parameter: 'args', expected an instance of type: 'int' but got argument of type: 'str'\.
```
pranavm-nvidia, line 492: Maybe we should just check this part of the message? We have lots of other tests that check how the candidate overloads are displayed.

tripy/tripy/function_registry.py
```diff
@@ -197,14 +197,20 @@ def matches_type(name: str, annotation: type, arg: Any) -> bool:
...
+        )
+        if not has_variadic_arg and len(args) > len(annotations):
             return Result.err(
                 [f"Function expects {len(annotations)} parameters, but {len(args)} arguments were provided."],
             )
 
-        for (name, annotation), arg in zip(annotations.items(), args):
+        positional_args_to_check = (
+            zip(annotation_items, args) if not has_variadic_arg else zip(annotation_items[:-1], args)
+        )
```
pranavm-nvidia, line 212: What if we just repeat the annotation for variadic arguments here? Then we wouldn't need a separate loop or special handling

slyubomirsky: It would check the first variadic argument only (that was the existing behavior, I tested it). We would still need separate logic to check the rest.

tripy/tripy/function_registry.py
```diff
@@ -197,14 +197,20 @@ def matches_type(name: str, annotation: type, arg: Any) -> bool:
         annotations = self._get_annotations()
 
         # Check if we have too many positional arguments. We can only do this if there isn't a variadic positional argument.
-        if not any(annotation.kind == inspect.Parameter.VAR_POSITIONAL for annotation in annotations.values()) and len(
-            args
-        ) > len(annotations):
+        # The variadic argument, if present, is necessarily the last.
```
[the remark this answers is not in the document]
pranavm-nvidia: Is that true for `annotations` as well though? I thought we merged all the annotations together. I know that the arguments themselves are separated into args/kwargs.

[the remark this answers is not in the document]
slyubomirsky: Yes, keyword args following variaidic args will be listed under `kwargs`. I will add a test case to demonstrate it.

tripy/tests/test_function_registry.py
slyubomirsky: Sure, easy enough.

tripy/tripy/function_registry.py
slyubomirsky: Oh, I misunderstood the comment, you meant copying over the final annotation that many times. Yes, that would result in less code.

tripy/tripy/function_registry.py
```diff
@@ -197,14 +199,30 @@ def matches_type(name: str, annotation: type, arg: Any) -> bool:
...
             return Result.err(
                 [f"Function expects {len(annotations)} parameters, but {len(args)} arguments were provided."],
             )
 
-        for (name, annotation), arg in zip(annotations.items(), args):
+        # If there is a variadic positional arg, we can copy the final annotation for the remaining args.
+        # Keyword-only args (only possible with a variadic arg) will appear in kwargs and don't need to be checked here.
+        if variadic_idx is not None:
+            positional_args_to_check = chain(
+                zip(annotation_items[:variadic_idx], args),
```
pranavm-nvidia, line 219: We could actually just express it like this:
```suggestion
                zip(annotation_items, args),
```
`len(args)` will always be less than or equal to `len(annotation_items[:variadic_idx])` so we don't gain anything by slicing it in advance.

pranavm-nvidia (approved):

Issue #302, opened 2024-10-25
`tripy.Module` does not support list of layers/ops of mixed types


farazkh80:
Currently, the following code (a simple classification head) 

```python
class ResNetClassifier(tp.Module):
    def __init__(self, num_classes=1000):
        super(ResNetClassifier, self).__init__()
        self.resnet = ResNetModel()
        self.classifier = [
               lambda x: tp.flatten(x, start_dim=1), 
              tp.Linear(2048, 1000))
        ]

    def __call__(self, x):
        features = self.resnet(x)
        output = self.classifier(features)
        return output
  
model = ResNetClassifier()
```

will show following warning
```
[W] A container of mixed types will not be registered with this module's state_dict().
Note: container was set here: 

--> /tmp/ipykernel_30063/1723333162.py:15 in __init__()
```

Upon discussion with @pranavm-nvidia , we should actually drop that constraint. we don't think it would be that hard to support containers of mixed types.

farazkh80:
closing since https://github.com/NVIDIA/TensorRT-Incubator/pull/365 merged to address this issue.

PR #365, opened 2024-11-13
[Tripy] Mixed module list/dict for `tp.Module`

farazkh80:
PR for addressing issue https://github.com/NVIDIA/TensorRT-Incubator/issues/302. Supporting mixed types is important,  since tripy does not have modules like `Identity` or `AveragePool`. With this, users can make sequential nets or module lists that include lambda functions or other callables like `tp.avgpool`.

- [x] modified `_iterate_members_of_type` to have a more relaxed check on checking module/paramater types.
- [x] Added mixed container tests to `tests/frontend/module/test_module.py`
- [x] Added mixed container tests to `tests/frontend/module/test_sequential.py`
- [x] Added example mixed container usage to `tp.Module`
- [x] Added example mixed container usage to `tp.Sequential`


tripy/tests/frontend/module/test_module.py
```diff
@@ -304,3 +304,77 @@ def test_state_dict(self, complex_network):
...
+        assert any(isinstance(value, tp.Module) for value in mixed_container_network.mixed_dict.values())
+        assert any(
+            callable(value) and not isinstance(value, tp.Module)
+            for value in mixed_container_network.mixed_dict.values()
+        )
+
+    def test_named_children(self, mixed_container_network):
+        children = list(mixed_container_network.named_children())
+        for _, child in children:
+            assert isinstance(child, tp.Module)
```
pranavm-nvidia, line 329: This isn't testing that we see all the children we expect to see. Maybe we can also check the count?

tripy/tests/frontend/module/test_module.py
```diff
@@ -304,3 +304,77 @@ def test_state_dict(self, complex_network):
...
+
+    def test_forward_pass(self, mixed_container_network):
+        output = mixed_container_network()
+        assert output is not None  # Ensure output is produced
+
+    def test_state_dict(self, mixed_container_network):
```
pranavm-nvidia, line 335: Should we rename to `test_load_state_dict`? I think it'd also be good to check that `state_dict()` includes parameters from modules in mixed containers.

tripy/tests/frontend/module/test_sequential.py
```diff
@@ -37,6 +37,14 @@ def dict_sequential_network():
     yield tp.Sequential({"layer1": tp.Linear(1, 3), "layer2": tp.Linear(3, 2)})
 
 
+@pytest.fixture
+def mixed_container_sequential_network():
+    yield tp.Sequential(
+        tp.Conv(in_channels=2, out_channels=2, kernel_dims=(1, 1), stride=(1, 1)),
+        lambda x: tp.avgpool(x, kernel_dims=(2, 2), stride=(1, 1)),
```
pranavm-nvidia, line 44: Would be good to have a module after the lambda to make sure indices are skipped correctly.

tripy/tests/frontend/module/test_sequential.py
```diff
@@ -153,6 +161,56 @@ def test_str_representation(self, dict_sequential_network):
...
+
+        assert np.array_equal(
+            cp.from_dlpack(mixed_container_sequential_network[0].weight), cp.from_dlpack(new_state_dict["0.weight"])
+        )
+        assert np.array_equal(
+            cp.from_dlpack(mixed_container_sequential_network[0].bias), cp.from_dlpack(new_state_dict["0.bias"])
+        )
+
+    def test_str_representation(self, mixed_container_sequential_network):
+        # Check if the string representation includes both Conv and lambda avgpool
```
pranavm-nvidia, line 199: Looks like the avgpool is being dropped here?

tripy/tests/frontend/module/test_module.py
```diff
@@ -304,3 +304,77 @@ def test_state_dict(self, complex_network):
...
+        assert str(mixed_container_network) == dedent(
+            """\
+            MixedContainerNetwork(
+              mixed_list.0=
+                DummyOp(
+                  nested=
+                    DummyNestedOp(
+                     param=[2],
+                    ),
+                ),
```
pranavm-nvidia, line 363: Would be good if we were able to include non-modules as well here

tripy/tripy/frontend/module/module.py
```diff
@@ -54,7 +54,8 @@ class Module:
     Base class used to define neural network modules.
     You can nest modules by assigning them as attributes of other modules.
 
-    Child modules or :class:`tripy.Parameter` s may be contained in Python ``list``\s or ``dict``\s.
+    Child modules, :class:`tripy.Parameter` s, or other callables/lambda functions may be contained
+    in Python ``list``\s or ``dict``\s.
     If using ``dict``\s, the keys must be strings.
```
pranavm-nvidia, line 59: I believe the backslashes here were intended to escape a space, not the `s`. 
```suggestion
    in Python ``list``\ s or ``dict``\ s.
    
    If using ``dict``\ s, the keys must be strings.
```

tripy/tripy/frontend/module/module.py
```diff
@@ -67,6 +68,15 @@ class Module:
             "layernorm": tp.LayerNorm(2),
         }
 
+    This is another valid example with a wrapped :class:`tripy.avgpool` lambda function
+    ::
+
+        self.conv = tp.Conv(in_channels=2, out_channels=2, kernel_dims=(1,1), stride=(1,1))
```
pranavm-nvidia, line 74: Can we either inline this or not make it a member? I think that would mess up `named_children()`

tripy/tripy/frontend/module/module.py
```diff
@@ -313,12 +323,14 @@ def _iterate_members_of_type(self, typ: T) -> Iterator[Tuple[str, T]]:
...
             if isinstance(value, typ):
                 yield name, value
-            elif isinstance(value, List) and _is_homogeneous_container(value, typ):
+            elif isinstance(value, List):
                 for i, obj in enumerate(value):
-                    yield f"{name}.{i}", obj
-            elif isinstance(value, Dict) and _is_homogeneous_container(value.values(), typ):
+                    if isinstance(obj, typ):
+                        yield f"{name}.{i}", obj
+            elif isinstance(value, Dict):
```
pranavm-nvidia, line 330: Is `_is_homogeneous_container` used anywhere? If not we can delete it

tripy/tests/frontend/module/test_sequential.py
farazkh80: yeah so that's a caveat. Because it is not a module, and it won't appear in `named_children`. If we wanted to show these callables, then we need to modify `__str__` for `Module` and also introduce some logic to keep track of callables. 

Option 2: name children can also include non modules but only use Module named_children in state dict?

tripy/tests/frontend/module/test_module.py
farazkh80: changed the name. Wdym by checking if parameters from modules are included? does this test not do that implicitly

```python
        state_dict = {
            "param": tensor,
            "mixed_list.0.nested.param": tensor,
            "mixed_list.2.nested.param": tensor,
            "mixed_dict.dummy.nested.param": tensor,
        }

        missing_keys, unexpected_keys = mixed_container_network.load_state_dict(state_dict, strict=True)

```

tripy/tests/frontend/module/test_module.py
```diff
@@ -304,3 +304,77 @@ def test_state_dict(self, complex_network):
...
+        )
+
+    def test_named_children(self, mixed_container_network):
+        children = list(mixed_container_network.named_children())
+        for _, child in children:
+            assert isinstance(child, tp.Module)
+
+    def test_forward_pass(self, mixed_container_network):
+        output = mixed_container_network()
+        assert output is not None  # Ensure output is produced
```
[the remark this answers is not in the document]
farazkh80: removed

tripy/tests/frontend/module/test_module.py
farazkh80: added count check

tripy/tripy/frontend/module/module.py
farazkh80: deleted

tripy/tests/frontend/module/test_module.py
pranavm-nvidia: No, this is only checking the setter (`load_state_dict`) but not the getter (`state_dict()`)

tripy/tripy/frontend/module/module.py
farazkh80: Oh good point, may bad there. Fixed now

tripy/tests/frontend/module/test_module.py
farazkh80: ok will fix the getter

tripy/tests/frontend/module/test_sequential.py
farazkh80: After discussing with @pranavm-nvidia we think modifying named_children is not a good idea. Since, state_dict won't include non-Module information anyways, it is fine for string representation not include it as well.

Hence, not changing anything.

tripy/tests/frontend/module/test_module.py
farazkh80: replied here https://github.com/NVIDIA/TensorRT-Incubator/pull/365#discussion_r1842895068

tripy/tests/frontend/module/test_sequential.py
farazkh80: good catch, added now!

farazkh80:
Thanks for the review @pranavm-nvidia , addressed all PR changes. Should be good for a second pass

tripy/tests/frontend/module/test_module.py
```diff
@@ -304,3 +304,83 @@ def test_state_dict(self, complex_network):
...
+
+        assert not missing_keys
+        assert not unexpected_keys
+        assert mixed_container_network.param is state_dict["param"]
+        assert mixed_container_network.mixed_list[0].nested.param is state_dict["mixed_list.0.nested.param"]
+        assert mixed_container_network.mixed_list[2].nested.param is state_dict["mixed_list.2.nested.param"]
+        assert mixed_container_network.mixed_dict["dummy"].nested.param is state_dict["mixed_dict.dummy.nested.param"]
+
+    def test_print_structure(self, mixed_container_network):
+        assert str(mixed_container_network) == dedent(
```
pranavm-nvidia, line 360: FYI this check will now fail due to my changes to `Module.__str__`. You'll need to update the expected string

farazkh80: fixed this! Thanks

PR #373, opened 2024-11-14
[mlir-tensorrt][CI] Fix CPM cache path

shelkesagar29:
This PR fixes CPM cache path in mlir-tensorrt CI.

### DIFF
--- a/.github/workflows/mlir-tensorrt-ci.yml
+++ b/.github/workflows/mlir-tensorrt-ci.yml
@@ -6,14 +6,18 @@ on:
       - main
     types: [synchronize, opened, reopened, ready_for_review]
     paths: ["mlir-tensorrt/**"]
+  push:
+    branches:
+      - main
+    paths: ["mlir-tensorrt/**"]
 
 env:
   DEFAULT_IMAGE: ghcr.io/nvidia/tensorrt-incubator/mlir-tensorrt:cuda12.5-ubuntu-llvm17
   REGISTRY: ghcr.io
 
 jobs:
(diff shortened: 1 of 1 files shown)


yizhuoz004 (approved):

christopherbate:
What about the CCACHE path?

christopherbate:
> > What about the CCACHE path?
> 
> CCACHE path seems to be correct. CPM cache path was wrong.
> 
> We mount ccache as `${{ github.workspace }}/ccache:/ccache` cache from GitHub workspace gets mounted into the container at `/ccache`. (project repo gets mounted at `/mlir-tensorrt`)
> 
> In run command, we do below BEFORE moving to `/mlir-tensorrt` folder. That means `CCACHE_DIR` env variable correctly points to ccache folder.
> 
> ```
> export CCACHE_BASEDIR="$PWD"
> export CCACHE_DIR="$PWD/ccache"
> export CCACHE_COMPILERCHECK=content
> export CCACHE_MAXSIZE=10G
> ```
> 
> Sometime its cache is loaded and sometimes its not. I am checking why this might be happening

We should be eliminating any room for error. If you know the absolute path of the cache directory, then use it. There's no point prefixing by `$PWD`. 

Also, if `$PWD` at this point is indeed "/", we should not besetting `CCACHE_BASEDIR=/`, see docs ().

Let's do this:

1. Don't set `CCACHE_BASEDIR` or `CCACHE_COMPILERCHECK`.
2. Change  directory to drop PWD prefix:`export CCACHE_DIR=/ccache`

Issue #360, opened 2024-11-12
Improve efficiency of getting shapes of constants


pranavm-nvidia:
In cases where the shape of a tensor is known in the frontend (e.g. constants), we should not generate code to retrieve the shape dynamically. Instead, we can just return the shape of the trace tensor. 

One way to do this would be to update the `shape` property of the frontend tensor to check if the shape of the trace tensor is known (check that all dimensions are non-negative integers). If it is, we can just return that shape directly. 

A case where this optimization could be very important is in `tp.Parameter`s. When we overwrite the parameters of a module, we perform shape/dtype checks to ensure that the new parameter is compatible. We should add a test to make sure this is sufficiently fast (we can use the `benchmark` fixture and add a test under `test_perf.py`).

slyubomirsky:
Implemented.

PR #379, opened 2024-11-15
[Tripy] Return the shape immediately if it is statically known instead of producing a trace operator.

slyubomirsky:
Addresses issue #360.

### DIFF
--- a/tripy/tests/performance/test_perf.py
+++ b/tripy/tests/performance/test_perf.py
@@ -135,9 +135,6 @@ def test_tripy_overhead():
         def measure_thunk():
             return compiled_one_io(*inputs)
 
-        for _ in range(warm_up_runs):
-            compiled_one_io(*inputs)
-
         return run_timed_trials(measure_thunk, warm_up_runs=warm_up_runs, iterations=iterations)
 
     assert measure_overhead(1) < 60.0
--- a/tripy/tests/performance/test_perf.py
+++ b/tripy/tests/performance/test_perf.py
@@ -151,13 +151,11 @@ def test_tripy_overhead():
     assert all(abs(delta - average_delta) < 10 for delta in deltas)
 
 
-def test_tripy_param_update():
+def test_tripy_param_update(benchmark):
     m = tp.Module()
     m.param = tp.Parameter([1, 2, 3, 4])
 
     def measure_thunk():
         m.param = tp.Parameter([5, 6, 7, 8])
 
-    # Since all the shapes are statically known, no compilation should need to happen
-    # and this should impose practically no overhead.
-    assert run_timed_trials(measure_thunk) < 60.0
+    benchmark(measure_thunk)
--- a/tripy/tests/performance/test_perf.py
+++ b/tripy/tests/performance/test_perf.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 import time
 from textwrap import dedent
+from typing import Callable
 
 import pytest
 import torch
@@ -23,6 +24,22 @@ from tests.performance.cases import *
 from tests.performance.conftest import PERF_CASES
 
 import tripy as tp
+
+
+def run_timed_trials(thunk: Callable[[], None], warm_up_runs=10, iterations=1000):
+    """
+    Returns the average time measured for calls to the thunk (the function intended to be timed)
+    in microseconds. First performs the specified number of untimed warm-ups.
+    """
+
+    for _ in range(warm_up_runs):
+        thunk()
+
+    start = time.perf_counter_ns()
+    for _ in range(iterations):
+        thunk()
+    end = time.perf_counter_ns()
+    return (end - start) / (iterations * 1000.0)
 
 
 @pytest.mark.parametrize("perf_case", PERF_CASES)
@@ -115,15 +132,13 @@ def test_tripy_overhead():
         for input in inputs:
             input.eval()
 
+        def measure_thunk():
+            return compiled_one_io(*inputs)
+
         for _ in range(warm_up_runs):
             compiled_one_io(*inputs)
 
-        start = time.perf_counter_ns()
-        for _ in range(iterations):
-            compiled_one_io(*inputs)
-        end = time.perf_counter_ns()
-
-        return (end - start) / (iterations * 1000.0)
+        return run_timed_trials(measure_thunk, warm_up_runs=warm_up_runs, iterations=iterations)
 
     assert measure_overhead(1) < 60.0
 
@@ -137,3 +152,15 @@ def test_tripy_overhead():
     # Ensure all deltas are within a few microseconds of each other
     average_delta = sum(deltas) / float(len(deltas))
     assert all(abs(delta - average_delta) < 10 for delta in deltas)
+
+
+def test_tripy_param_update():
+    m = tp.Module()
+    m.param = tp.Parameter([1, 2, 3, 4])
+
+    def measure_thunk():
+        m.param = tp.Parameter([5, 6, 7, 8])
+
+    # Since all the shapes are statically known, no compilation should need to happen
+    # and this should impose practically no overhead.
+    assert run_timed_trials(measure_thunk) < 60.0
--- a/tripy/tripy/backend/mlir/utils.py
+++ b/tripy/tripy/backend/mlir/utils.py
@@ -85,7 +85,6 @@ def get_mlir_scalar_attr(mlir_dtype, value):
     # MLIR represents float dtypes as FloatAttr
     # and non-float dtypes as IntegerAttr
     attr_func = ir.IntegerAttr.get if isinstance(mlir_dtype, ir.IntegerType) else ir.FloatAttr.get
-    value = int(value) if isinstance(mlir_dtype, ir.IntegerType) else float(value)
     return attr_func(mlir_dtype, value)
 
 
--- a/tripy/tripy/backend/mlir/utils.py
+++ b/tripy/tripy/backend/mlir/utils.py
@@ -85,6 +85,7 @@ def get_mlir_scalar_attr(mlir_dtype, value):
     # MLIR represents float dtypes as FloatAttr
     # and non-float dtypes as IntegerAttr
     attr_func = ir.IntegerAttr.get if isinstance(mlir_dtype, ir.IntegerType) else ir.FloatAttr.get
+    value = int(value) if isinstance(mlir_dtype, ir.IntegerType) else float(value)
     return attr_func(mlir_dtype, value)
 
 
--- a/tripy/tripy/frontend/trace/ops/reduce.py
+++ b/tripy/tripy/frontend/trace/ops/reduce.py
@@ -296,22 +296,24 @@ def prod(
 
 
 def mean_impl(tensor: "tripy.Tensor", dim: Union[int, Sequence] = None, keepdim: bool = False, apply_to_divisor=None):
-    from tripy.frontend.dimension_size import DimensionSize
     from tripy.frontend.tensor import Tensor
     from tripy.frontend.trace.ops.cast import cast
 
     sum_val = sum(tensor, dim=dim, keepdim=keepdim)
 
     # compute number of elements in the array and divide by number of elements in dims
-
-    # Need to convert shape to DimensionSize if it isn't already for the MLIR lowering to work
-    shape = [d if isinstance(d, Tensor) else DimensionSize(d) for d in tensor.shape]
+    shape = tensor.shape
     num_elements = math.prod(shape if dim is None else [shape[d] for d in make_list(dim)])
 
     if apply_to_divisor:
         num_elements = apply_to_divisor(num_elements)
 
-    return sum_val / (cast(num_elements, sum_val.dtype))
+    num_elements = (
+        cast(num_elements, sum_val.dtype)
+        if isinstance(num_elements, Tensor)
+        else Tensor(num_elements, dtype=sum_val.dtype)
+    )
+    return sum_val / num_elements
 
 
 @export.public_api(document_under="operations/functions")
--- a/tripy/tripy/frontend/trace/ops/reduce.py
+++ b/tripy/tripy/frontend/trace/ops/reduce.py
@@ -296,12 +296,16 @@ def prod(
 
 
 def mean_impl(tensor: "tripy.Tensor", dim: Union[int, Sequence] = None, keepdim: bool = False, apply_to_divisor=None):
+    from tripy.frontend.dimension_size import DimensionSize
+    from tripy.frontend.tensor import Tensor
     from tripy.frontend.trace.ops.cast import cast
 
     sum_val = sum(tensor, dim=dim, keepdim=keepdim)
 
     # compute number of elements in the array and divide by number of elements in dims
-    shape = tensor.shape
+
+    # Need to convert shape to DimensionSize if it isn't already for the MLIR lowering to work
+    shape = [d if isinstance(d, Tensor) else DimensionSize(d) for d in tensor.shape]
     num_elements = math.prod(shape if dim is None else [shape[d] for d in make_list(dim)])
 
     if apply_to_divisor:
--- a/tripy/tripy/frontend/trace/ops/reshape.py
+++ b/tripy/tripy/frontend/trace/ops/reshape.py
@@ -49,8 +49,7 @@ def infer_dimensions(input: "tripy.Tensor", shape: ShapeLike) -> ShapeLike:
         raise_error(f"The new shape can have at most one inferred dimension (denoted by -1)", [f"Got shape: {shape}."])
 
     if num_unknown_dims == 1:
-        input_shape = input.shape
-        input_volume = math.prod(input_shape)
+        input_volume = math.prod(input.shape)
         known_dims_volume = math.prod(dim for dim in shape if not op_utils.is_minus_one(dim))
         inferred_dim = input_volume / known_dims_volume
 
--- a/tripy/tripy/frontend/trace/ops/reshape.py
+++ b/tripy/tripy/frontend/trace/ops/reshape.py
@@ -43,16 +43,13 @@ class Reshape(BaseTraceOp):
 
 
 def infer_dimensions(input: "tripy.Tensor", shape: ShapeLike) -> ShapeLike:
-    from tripy.frontend.dimension_size import DimensionSize
-    from tripy.frontend.tensor import Tensor
 
     num_unknown_dims = len([dim for dim in shape if op_utils.is_minus_one(dim)])
     if num_unknown_dims > 1:
         raise_error(f"The new shape can have at most one inferred dimension (denoted by -1)", [f"Got shape: {shape}."])
 
     if num_unknown_dims == 1:
-        # need to convert the shape to DimensionSize if it isn't already
-        input_shape = [DimensionSize(d) if not isinstance(d, Tensor) else d for d in input.shape]
+        input_shape = input.shape
         input_volume = math.prod(input_shape)
         known_dims_volume = math.prod(dim for dim in shape if not op_utils.is_minus_one(dim))
         inferred_dim = input_volume / known_dims_volume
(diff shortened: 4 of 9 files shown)


slyubomirsky:
When informally benchmarking, I did not find a large difference between the current implementation and this change, however. Perhaps the issue is how the benchmark was measured. Informal benchmark:

The function to test:
```python
def func(t1, t2):
    return tp.Tensor(list(t1.shape) != list(t2.shape), dtype=tp.bool)
```

I wrote it this way because the `_is_compatible` helper in `Parameter` returns a Python bool, but `tp.compile` can only compile functions returning a Tripy tensor.

Comparison:
```
c = tp.compile(func, args=[tp.InputInfo(shape=[1, 3, 3], dtype=tp.int32), tp.InputInfo(shape=[1, 3, 3], dtype=tp.int32)])
t1 = tp.Tensor([[[1, 2, 3], [4, 5, 6], [7, 8, 9]]])
t2 = tp.Tensor([[[1, 2, 3], [4, 5, 6], [7, 8, 9]]])

def time_run():
    start = time.perf_counter_ns()
    for _ in range(1000):
        c(t1, t2)
    end = time.perf_counter_ns()
    return (end - start) / 1.0e6
```
After warming up, both complete in about 25 microseconds. Should we do this comparison without compiling instead? I could also vary the input values.

tripy/tripy/frontend/trace/ops/shape.py
```diff
@@ -63,4 +63,12 @@ def shape(self: "tripy.Tensor") -> List["tripy.DimensionSize"]:
...
 
-    return [GetDimensionSize.build([self], dim=index, always_cast_to_dimension_size=True) for index in range(self.rank)]
+    from tripy.frontend.dimension_size import DimensionSize
+
+    # If the shape is statically known, we do not need to insert any operator calls and we can memoize.
+    if self.shape_memo is None and all(dim >= 0 for dim in self.trace_tensor.shape):
+        self.shape_memo = [DimensionSize(dim) for dim in self.trace_tensor.shape]
+
+    return self.shape_memo or [
+        GetDimensionSize.build([self], dim=index, always_cast_to_dimension_size=True) for index in range(self.rank)
```
slyubomirsky, line 73: Memoizing the `GetDimensionSize` ops led to an MLIR error upon repeated calls, and I'm not entirely sure why.

The case that fails (doc for repeat):
```python
inp = tp.reshape(tp.arange(4, dtype=tp.int32), (2, 2))
out0 = tp.repeat(inp, 2, dim=0)
out1 = tp.repeat(inp, 2, dim=1)

np_inp = np.from_dlpack(tp.copy(inp, device=tp.device("cpu"))) # doc: omit
ref_out0 = np.repeat(np_inp, 2, 0)
assert np.array_equal(ref_out0, np.from_dlpack(tp.copy(out0, device=tp.device("cpu")))) # <- succeeds

ref_out1 = np.repeat(np_inp, 2, 1)
assert np.array_equal(ref_out1, np.from_dlpack(tp.copy(out1, device=tp.device("cpu")))) # <- fails
```

Error:
```
    (t347)error: number of output elements (8) doesn't match expected number of elements (4)

    This error occured while trying to compile the following FlatIR expression:
          |
          | t347: [rank=(3), shape=((-1, -1, -1)), dtype=(int32), loc=(gpu:0)] = DynamicReshapeOp(t305, t346)
          | 

    Note: This originated from the following expression:

    --> /tripy/docs/../tripy/frontend/ops/unsqueeze.py:56 in unsqueeze()
          |
       56 |     return reshape(input, result_shape)
          |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

    --> /tripy/docs/../tripy/frontend/ops/repeat.py:92 in repeat()
          |
       92 |     out = unsqueeze(input, dim + 1)
          |           ^^^^^^^^^^^^^^^^^^^^^^^^^ --- required from here
```

The unsqueeze gets the shape of the input and computes a new shape as an argument to `reshape`. I will see if I can figure out how that results in an error when lowering to MLIR.

tripy/tripy/frontend/trace/ops/shape.py
```diff
@@ -63,4 +63,12 @@ def shape(self: "tripy.Tensor") -> List["tripy.DimensionSize"]:
         assert shape == [8, 2]
     """
 
-    return [GetDimensionSize.build([self], dim=index, always_cast_to_dimension_size=True) for index in range(self.rank)]
+    from tripy.frontend.dimension_size import DimensionSize
+
+    # If the shape is statically known, we do not need to insert any operator calls and we can memoize.
+    if self.shape_memo is None and all(dim >= 0 for dim in self.trace_tensor.shape):
+        self.shape_memo = [DimensionSize(dim) for dim in self.trace_tensor.shape]
```
pranavm-nvidia, line 70: I'm not sure there's actually a benefit to storing `shape_memo`. Can't we simplify this to:
```py
if all(dim >= 0 for dim in self.trace_tensor.shape):
    return list(self.trace_tensor.shape)
```
?
I don't think we even need to return `DimensionSize` in this case.

pranavm-nvidia:
@slyubomirsky most of the impact will be on the compile time I think.

slyubomirsky:
Ah, okay. I'll profile compile time then.

tripy/tripy/frontend/trace/ops/shape.py
slyubomirsky: It's the signature of the shape operator.

pranavm-nvidia: Sure, but we could always change the signature. Ideally, `DimensionSize` and `int` behave basically the same from the user's perspective. I'm assuming you have `shape_memo` here because it's expensive to construct `DimensionSize`s?

slyubomirsky: Oh that's right, most of the interfaces will convert `ShapeLike`s anyway. I didn't want to create new `DimensionSize`s constantly, yes. I'll keep it as a tuple of ints if it won't break anything.

slyubomirsky: One place where returning straight ints does break things: In `slice`, some of the preprocessing assumes that the elements of a shape will be tensors. That will just take some wrapping in those cases.

slyubomirsky: We can compare the changes in commit [8e3a4f4b](https://github.com/NVIDIA/TensorRT-Incubator/pull/379/commits/8e3a4f4b7ac71505501ba8964f280bef960975fc). It might be less of a headache to do the conversion in the shape operator than to figure out whether to convert afterwards.

slyubomirsky:
FWIW, some cursory benchmarking also did not reveal much change in the compile time either (a difference about 100 microseconds over 1000 trials), though it might be a matter of the function that is compiled.

The function whose compilation I timed (the return value needed to be a tensor for it to be valid to submit to `tp.compile`):

```
def func():
      return tp.concatenate(tp.Tensor([1, 2, 3, 4].shape, dim=0)
```

If there's a specific benchmark we have in mind, that might better to measure.

tripy/tripy/frontend/trace/ops/shape.py
pranavm-nvidia: That's quite strange - why aren't we able to use integers in these places? That seems like an MLIR-TRT bug.

pranavm-nvidia:
@slyubomirsky I think a benchmark where it could make a difference is setting parameters on modules without involving any Tripy operations, e.g.:
```py
module.param = tp.Parameter(torch.tensor([1, 2, 3]))
```
Previously, this would have triggered a compilation in order to compare the shape of the new parameter with that of the old one. With your changes, the compilation should be entirely bypassed.

tripy/tripy/frontend/trace/ops/shape.py
slyubomirsky: Okay, looks like these can be corrected by handling scalars in a couple of other places. Probably less complicated to do it that way than to wrap `DimensionSize` manually, so I'll change it over.

pranavm-nvidia: Are we returning floats instead of ints for the shape? Seems like it's turning into a float at some point for whatever reason.

slyubomirsky:
Comparing the parameter update time directly does yield the expected difference 🙂 100x faster due to not compiling.

tripy/tests/performance/test_perf.py
```diff
@@ -115,15 +132,13 @@ def func({arg_str}):
...
 
-        for _ in range(warm_up_runs):
-            compiled_one_io(*inputs)
+        def measure_thunk():
+            return compiled_one_io(*inputs)
 
-        start = time.perf_counter_ns()
-        for _ in range(iterations):
+        for _ in range(warm_up_runs):
             compiled_one_io(*inputs)
```
pranavm-nvidia, line 139: We can drop the warm-up loop too right? It's already being done in the helper.

tripy/tests/performance/test_perf.py
```diff
@@ -137,3 +152,15 @@ def func({arg_str}):
...
+def test_tripy_param_update():
+    m = tp.Module()
+    m.param = tp.Parameter([1, 2, 3, 4])
+
+    def measure_thunk():
+        m.param = tp.Parameter([5, 6, 7, 8])
+
+    # Since all the shapes are statically known, no compilation should need to happen
+    # and this should impose practically no overhead.
+    assert run_timed_trials(measure_thunk) < 60.0
```
pranavm-nvidia, line 166: I think we want this to be a regression test - we can use the `benchmark` fixture to achieve that.

tripy/tripy/backend/mlir/utils.py
```diff
@@ -85,6 +85,7 @@ def get_mlir_scalar_attr(mlir_dtype, value):
     # MLIR represents float dtypes as FloatAttr
     # and non-float dtypes as IntegerAttr
     attr_func = ir.IntegerAttr.get if isinstance(mlir_dtype, ir.IntegerType) else ir.FloatAttr.get
+    value = int(value) if isinstance(mlir_dtype, ir.IntegerType) else float(value)
```
pranavm-nvidia, line 88: Do you know why we were getting the wrong type here in the first place?

tripy/tripy/frontend/trace/ops/reshape.py
```diff
@@ -43,12 +43,14 @@ def to_flat_ir(self, inputs, outputs):
...
 def infer_dimensions(input: "tripy.Tensor", shape: ShapeLike) -> ShapeLike:
+
     num_unknown_dims = len([dim for dim in shape if op_utils.is_minus_one(dim)])
     if num_unknown_dims > 1:
         raise_error(f"The new shape can have at most one inferred dimension (denoted by -1)", [f"Got shape: {shape}."])
 
     if num_unknown_dims == 1:
-        input_volume = math.prod(input.shape)
+        input_shape = input.shape
+        input_volume = math.prod(input_shape)
```
pranavm-nvidia, line 53: Can probably revert this file now?

tripy/tripy/frontend/trace/ops/slice.py
```diff
@@ -187,20 +188,23 @@ def __getitem__(
     flip_dims = []
     for i, idx in enumerate(index):
 
+        # t.shape could potentially return ints, so we would want to convert in those cases
+        shape_dim = DimensionSize(t_shape[i]) if not isinstance(t_shape[i], Tensor) else t_shape[i]
```
pranavm-nvidia, line 192: Should we do this conversion only at the point where it's needed, i.e. in `where`?

slyubomirsky: I'll do that. Might be simpler that way.

tripy/tripy/frontend/trace/ops/reduce.py
```diff
@@ -307,7 +308,12 @@ def mean_impl(tensor: "tripy.Tensor", dim: Union[int, Sequence] = None, keepdim:
     if apply_to_divisor:
         num_elements = apply_to_divisor(num_elements)
 
-    return sum_val / (cast(num_elements, sum_val.dtype))
+    num_elements = (
+        cast(num_elements, sum_val.dtype)
+        if isinstance(num_elements, Tensor)
+        else Tensor(num_elements, dtype=sum_val.dtype)
+    )
+    return sum_val / num_elements
```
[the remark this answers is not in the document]
slyubomirsky: It is the return type for `mean`. We could do the conversion there if we wish.

tripy/tripy/backend/mlir/utils.py
slyubomirsky: It came from here in `reshape.py`:

```python
def infer_dimensions(input: "tripy.Tensor", shape: ShapeLike) -> ShapeLike:

    num_unknown_dims = len([dim for dim in shape if op_utils.is_minus_one(dim)])
    if num_unknown_dims > 1:
        raise_error(f"The new shape can have at most one inferred dimension (denoted by -1)", [f"Got shape: {shape}."])

    if num_unknown_dims == 1:
        input_volume = math.prod(input.shape)
        known_dims_volume = math.prod(dim for dim in shape if not op_utils.is_minus_one(dim))
        inferred_dim = input_volume / known_dims_volume # <- output is floating point

        shape = [inferred_dim if op_utils.is_minus_one(dim) else dim for dim in shape]

    return {"shape": shape}
```

We don't get a floating point value on that line if we have `DimensionSize` since it would be doing `int32` division, but we do for a Python number.

tripy/tripy/frontend/trace/ops/reshape.py
slyubomirsky: Done.

tripy/tests/performance/test_perf.py
slyubomirsky: Very handy, I didn't realize it was built into pytest.

tripy/tests/performance/test_perf.py
slyubomirsky: Done.

tripy/tripy/frontend/trace/ops/reduce.py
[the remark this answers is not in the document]
slyubomirsky: If you mean specifically `num_elements`, there was already a call to `cast` there but `cast` doesn't work for a Python number as an argument.

tripy/tripy/backend/mlir/utils.py
pranavm-nvidia: I see. I wonder if we should do a floor division to more accurately convey the intent?

tripy/tripy/frontend/trace/ops/slice.py
```diff
@@ -197,10 +198,14 @@ def convert_to_positive_idx(index: Union[int, Tensor]) -> Union[int, Tensor]:
...
+                return 0
+
+            # the dimension needs to be a tensor for the condition to where
+            shape_dim = t_shape[i] if isinstance(t_shape[i], Tensor) else DimensionSize(t_shape[i])
             if isinstance(bound, int):
-                return 0 if bound < 0 else where(bound > t_shape[i], t_shape[i], Tensor([bound]))
-            else:
-                return where(bound < 0, Tensor([0]), where(bound > t_shape[i], t_shape[i], bound))
+                return where(bound > shape_dim, shape_dim, Tensor([bound]))
+            return where(bound < 0, Tensor([0]), where(bound > shape_dim, shape_dim, bound))
```
pranavm-nvidia, line 208: Not related to your change, but why did we use `where` instead of `min`/`max`?

pranavm-nvidia (approved):

tripy/tripy/frontend/trace/ops/slice.py
slyubomirsky: I don't think there's a reason; I'll replace them.

tripy/tripy/backend/mlir/utils.py
slyubomirsky: It seems floor division for tensors is just division and doesn't do any casting. I will see if it's easy to change it to also include a cast to an int type.

slyubomirsky: Replaced it here and removed the cast in utils. I can put it back if we prefer but it might be handy to have something fail if we get the type wrong.

tripy/tripy/frontend/trace/ops/slice.py
```diff
@@ -164,7 +164,9 @@ def __getitem__(
         assert np.array_equal(cp.from_dlpack(output).get(), np.arange(10)[8:2:-1])
 
     """
+    from tripy.frontend.dimension_size import DimensionSize
     from tripy.frontend.tensor import Tensor
+    from tripy.frontend.trace.ops.binary_elementwise import maximum, minimum as tensor_min
```
pranavm-nvidia, line 169: Why the asymmetry here? Can we just import it as `minimum`?

tripy/tripy/frontend/trace/ops/shape.py
```diff
@@ -63,4 +63,8 @@ def shape(self: "tripy.Tensor") -> List["tripy.DimensionSize"]:
         assert shape == [8, 2]
     """
 
+    # If the shape is statically known, we do not need to insert any operator calls.
+    if all(dim >= 0 for dim in self.trace_tensor.shape):
```
pranavm-nvidia, line 67: Oh one more thing here is that we don't want to do this if we're tracing for compilation. Can you add another condition: `and not self.trace_tensor.is_compile_tracer`?

pranavm-nvidia: The reason is that we may want to use dynamic shapes at runtime that are different than the shape of the tensor that's being passed in for tracing. In that case we have to insert operations to get the shape at runtime.

tripy/tripy/frontend/trace/ops/slice.py
slyubomirsky: The ordinary Python `min` is used elsewhere and I didn't want to introduce confusion.

pranavm-nvidia: But it's `minimum` and not `min` right?

tripy/tripy/frontend/trace/ops/shape.py
pranavm-nvidia: Why is `compile_tracer` being set for the parameter update? That should only happen in the `compile` call.

tripy/tripy/frontend/trace/ops/slice.py
slyubomirsky: Yeah, just seemed like potential for confusion to me. I'll get rid of it if you don't think it's necessary.

tripy/tripy/frontend/trace/ops/shape.py
slyubomirsky: My mistake, I was assuming that it was tracing based on having a difference between profiling results but it seems to have just been variability on my system. There was no difference when I checked again and I put in an assert to verify that the compile tracer wasn't being used for that case.

pranavm-nvidia: Could we also add a test case for this? We could have some code like:
```py
compiled_ones = tp.compile(tp.ones_like, args=[...]) # Use an input with a dynamic shape here

a = tp.Tensor(...)
out = compiled_ones(a.shape) 
assert out.shape == a.shape # Test with various shapes of `a` and make sure it works
```
and then compile it with dynamic shapes. If we accidentally freeze the shapes while compiling, this should catch it.

slyubomirsky: This particular test case works without the additional condition as well. Ones I try where the shape is used in the compiled function still all pass without the extra condition, since the dims for the trace tensor are all -1 regardless of the input info.

Per compile.py, it doesn't seem that the shape bounds specified in the input info will make it to the individual ops:
```
    def process_arg(name, arg):
        if isinstance(arg, InputInfo):
            # Make new tensors for tracing.
            from tripy.common.datatype import floating, integer
            from tripy.frontend.trace.ops.fill import full

            init_value = 1 if issubclass(arg.dtype, integer) else 1.0 if issubclass(arg.dtype, floating) else True
            tensor = full(shape=arg.shape_bounds.opt, value=init_value, dtype=arg.dtype) # arg.shape_bounds.opt comes from input info
            tensor.name = name
            tensor.trace_tensor.is_compile_tracer = True
            # however, tensor.trace_tensor.shape is all -1

            trace_input_map[name] = tensor
            shapes.append(arg.shape_bounds)
            input_names.add(name)

            return tensor
        return arg
```

pranavm-nvidia: Ah that must be because we use `full` to generate the input tensors for tracing. I think it's still a good test to have to ensure we don't make changes to how we generate input tensors that will break dynamic shapes.

pranavm-nvidia (approved):

Issue #206, opened 2024-09-16
Slicing a shape with a negative index sometimes does not work


pranavm-nvidia:
If we use the output of slicing a shape with a negative index in certain operations, we get a failure:
```py
>>> a = tp.Tensor([1, 2, 3])
>>> a.shape
shape([3], dtype=int32, loc=gpu:0, shape=[1])

>>> tp.ones([a.shape[0]])
tensor([1.0000, 1.0000, 1.0000], dtype=float32, loc=gpu:0, shape=[3])

>>> tp.ones([a.shape[-1]])

MTRTException: InternalError: failed to run compilation on module with symbol name: outs_t439_19

Additional context:
Traceback (most recent call last):
  File "/tripy/tripy/backend/mlir/compiler.py", line 102, in compile
    executable = compiler.compiler_stablehlo_to_executable(
mlir_tensorrt.runtime._mlir_libs._api.MTRTException: InternalError: failed to run compilation on module with symbol name: outs_t439_19
.
    error: StableHLO dynamic shape canonicalization failed to converge within 4 iterations
```

slyubomirsky:
MLIR dump from this example:

```
module @outs_t39_2 {
  func.func @main() -> tensor<?xf32> {
    %c = stablehlo.constant dense<[1, 2, 3]> : tensor<3xi32>
    %c_0 = stablehlo.constant dense<3> : tensor<i32>
    %c_1 = stablehlo.constant dense<1> : tensor<1xi32>
    %c_2 = stablehlo.constant dense<3> : tensor<1xi32>
    %c_3 = stablehlo.constant dense<1> : tensor<i32>
    %c_4 = stablehlo.constant dense<1> : tensor<1xi32>
    %c_5 = stablehlo.constant dense<0> : tensor<i32>
    %c_6 = stablehlo.constant dense<1> : tensor<i32>
    %c_7 = stablehlo.constant dense<0> : tensor<1xi32>
    %c_8 = stablehlo.constant dense<1> : tensor<1xi32>
    %0 = stablehlo.compare  LE, %c_7, %c_8 : (tensor<1xi32>, tensor<1xi32>) -> tensor<1xi1>
    %1 = stablehlo.select %0, %c_7, %c_8 : tensor<1xi1>, tensor<1xi32>
    %c_9 = stablehlo.constant dense<1> : tensor<1xi32>
[...]

PR #385, opened 2024-11-16
[compiler/StableHloExt] Bump max iterations in `stablehlo-ext-refine-shapes`

christopherbate:
Increases the max number of iterations that the initial dynamic shape refinement
pipeline will iterate in order to better reported use cases.


christopherbate:
Resolves #206

PR #387, opened 2024-11-19
Add compile fixture to test integration ops with compile mode

parthchadha:


tripy/tests/integration/conftest.py
```diff
@@ -0,0 +1,86 @@
...
+
+import pytest
+import inspect
+
+import tripy as tp
+
+
+@pytest.fixture(params=["compile", "eager"])
+def execution_mode(request):
+    return request.param
```
pranavm-nvidia, line 26: Why do we need this fixture instead of directly parametrizing `compile_fixture`?

tripy/tests/integration/conftest.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+@pytest.fixture
+def compile_fixture(execution_mode):
+    def wrapper(func, *args, exclude_arg_names=None, **kwargs):
+        def get_shape(x: tp.Tensor):
+            x.eval()
+            return tp.InputInfo(x.trace_tensor.shape, dtype=x.dtype)
```
pranavm-nvidia, line 34: I don't think we need to evaluate the tensor:
```suggestion
        def get_shape(x: tp.Tensor):
            return tp.InputInfo(list(map(int, x.shape)), dtype=x.dtype)
```

pranavm-nvidia: Actually, it should be possible to use `InputInfo` with `DimensionSize`s. We could probably just replace the `isinstance(elem, numbers.Number)` checks to also look for `DimensionSize`s. I can file an issue to do that separately if you want.

tripy/tests/integration/conftest.py
```diff
@@ -0,0 +1,86 @@
...
+
+        exclude_arg_names = set() if exclude_arg_names is None else set(exclude_arg_names)
+
+        if execution_mode == "compile":
```
pranavm-nvidia, line 38: nit: we can do an early exit here and unindent the rest of the code:
```suggestion
        if execution_mode == "eager":
            return func(*args, **kwrags) 
        
        assert execution_mode == "compile"
        ...
```

tripy/tests/integration/conftest.py
```diff
@@ -0,0 +1,86 @@
...
+            # For args, we need to get their parameter names from function signature
+            sig = inspect.signature(func)
+            param_names = list(sig.parameters.keys())
```
pranavm-nvidia, line 41: We have a `get_positional_arg_names` helper that does this

tripy/tests/integration/conftest.py
```diff
@@ -0,0 +1,86 @@
...
+
+            compile_args = []
+            for i, arg in enumerate(args):
+                if i < len(param_names) and param_names[i] in exclude_arg_names:
+                    compile_args.append(arg)
+                elif isinstance(arg, tp.Tensor) and not isinstance(arg, tp.DimensionSize):
```
pranavm-nvidia, line 47: Why is `DimensionSize` treated specially? Can it not be an input to a compiled function?

tripy/tests/integration/conftest.py
```diff
@@ -0,0 +1,86 @@
...
+import tripy as tp
+
+
+@pytest.fixture(params=["compile", "eager"])
+def execution_mode(request):
+    return request.param
+
+
+@pytest.fixture
+def compile_fixture(execution_mode):
```
pranavm-nvidia, line 30: Maybe we can rename to `eager_or_compiled` or similar to convey the intent

tripy/tests/integration/test_iota.py
```diff
@@ -54,11 +54,11 @@ def _compute_ref_iota(self, dtype, shape, dim):
...
         ],
     )
-    def test_iota(self, dtype, shape, dim):
+    def test_iota(self, dtype, shape, dim, compile_fixture):
         if dim:
-            output = tp.iota(shape, dim, dtype[1])
+            output = compile_fixture(tp.iota, shape, dim, dtype[1])
         else:
-            output = tp.iota(shape, dtype=dtype[1])
+            output = compile_fixture(tp.iota, shape, dtype=dtype[1])
```
pranavm-nvidia, line 61: This is weird. `dim=None` should work, so we only need one branch here:
```py
output = compile_fixture(tp.iota, shape, dim, dtype[1])
```

tripy/tests/integration/test_dequantize.py
```diff
@@ -29,28 +29,28 @@ class TestDequantize:
...
         "dtype", [tp.float32, tp.float16, pytest.param(tp.bfloat16, marks=skip_if_older_than_sm80)]
     )
-    def test_dequantize_int8_per_tensor(self, dtype):
+    def test_dequantize_int8_per_tensor(self, dtype, compile_fixture):
         data = [4, 8]
         input_tp = tp.Tensor(data, dtype=tp.int8)
         scale = torch.tensor(0.5, dtype=TORCH_DTYPES[dtype])
         scale_tp = tp.Tensor(scale, dtype=dtype)
-        dequantized = tp.dequantize(input_tp, scale_tp, dtype)
+        dequantized = compile_fixture(tp.dequantize, input_tp, scale_tp, dtype, exclude_arg_names=["scale"])
```
pranavm-nvidia, line 37: I think we can drop `exclude_arg_names` by using the same approach you used in `cast`:
```py
def func(input):
    return tp.dequantize(input, scale_tp, dtype)

dequantized = compile_fixture(func, input_tp)
```
That will simplify the fixture implementation considerably IMO.

tripy/tests/integration/conftest.py
pranavm-nvidia: If we fix the tests, then we can actually drop this logic. See https://github.com/NVIDIA/TensorRT-Incubator/pull/387#discussion_r1848826467

tripy/tests/integration/conftest.py
parthchadha: This is to prevent conversion to Tensor since we just want to feed DimensionSize directly as input.

pranavm-nvidia: Ah but we would still need to use an `InputInfo` while compiling right?

tripy/tests/integration/conftest.py
```diff
@@ -0,0 +1,86 @@
...
+    def wrapper(func, *args, exclude_arg_names=None, **kwargs):
+        def get_shape(x: tp.Tensor):
```
pranavm-nvidia, line 32: `get_input_info` would be a better name

tripy/tests/integration/conftest.py
parthchadha: But we don't really have value range implemented yet so `DimensionSize` can be constant folded into the graph (therefore passed as an input directly)

tripy/tests/integration/test_iota.py
pranavm-nvidia: @parthchadha my mistake, I was thinking of other ops. `dim=None` doesn't make sense for iota since it can only operate along one dimension. Maybe we should just update the test to use `dim=0` instead of `None`?

parthchadha:
The only failing test will be resovled after container update.

pranavm-nvidia (approved):

Issue #338, opened 2024-11-05
Enable function registry for methods


pranavm-nvidia:
When we decorate a class with a function registry (e.g. `export.public_api()`) we should apply it to all methods of that class. This allows us to support features like type checking and overloading for methods. 

We can implement this by updating the function registry decorator to recursively call itself on each method of the class. The key we call it with will need to include the class name as well as the method name, or we will get collisions quickly. We may need to update `get_callee_func_name` to strip out the class name so we can add column information. To test this, we can check the stack information of a tensor produced by a Module's `__call__` method to ensure it contains accurate column ranges.

slyubomirsky:
Regarding the `__call__` on a method example, I'm not sure there is a great way to register all the methods on a class without ignoring magic methods without unleashing chaos. Namely, if we want to process all the magic methods in something like `export.public_api`, many of them will cause errors due to not having type annotations written out. For magic methods, we might want to use a whitelist for those that make sense.

pranavm-nvidia:
@slyubomirsky I assumed there might be some way to determine which methods are implemented/overridden by a child class. Another way would be to only apply this to methods with docstrings (i.e. public facing methods).

slyubomirsky:
I'll see if there's a way to check for overrides specifically with `inspect`. Docstrings are a good idea too, perhaps that will be the way to go.

slyubomirsky:
I think checking for a docstring might be safer because of another edge case that comes up: methods added by decorators. For example, `@dataclass` adds implementations of `__eq__`, etc., and it is not done by inheritance. We could probably detect that case but others like it might come up.

slyubomirsky:
Implemented.

PR #388, opened 2024-11-19
[Tripy] Change function registry to also register methods of registered classes

slyubomirsky:
Addresses #338. The function registry decorator will now recursively apply the decorator to methods of a decorated class.

### DIFF
--- a/tripy/tests/performance/test_perf.py
+++ b/tripy/tests/performance/test_perf.py
@@ -155,7 +155,10 @@ def test_tripy_param_update(benchmark):
     m = tp.Module()
     m.param = tp.Parameter([1, 2, 3, 4])
 
+    # Leave the instantiation outside of the measured section to avoid overhead from registry calls.
+    new_param = tp.Parameter([5, 6, 7, 8])
+
     def measure_thunk():
-        m.param = tp.Parameter([5, 6, 7, 8])
+        m.param = new_param
 
     benchmark(measure_thunk)


tripy/tests/frontend/test_utils.py
```diff
@@ -59,7 +59,7 @@ def func(a: tp.types.TensorLike):
         a = func(1.0)
 
         assert isinstance(a, tp.Tensor)
-        assert a.stack_info[2].column_range == (17, 20)
+        assert a.stack_info[4].column_range == (17, 20)
```
slyubomirsky, line 62: It seems that the the slice method is registered twice, once due to being in the tensor method registry and a second time probably because of the `export.public_api` decorator's being applied to all methods of `Tensor`. Probably not a problem, but amusing.

pranavm-nvidia: Is there a way to avoid that? It will slow down the code unnecessarily since we'll end up doing type checking twice. We could either have the function registry skip methods that are already part of a function registry, or we could change the `TENSOR_METHOD_REGISTRY` to be just a normal `dict` instead of a `FunctionRegistry`. 

The latter seems cleaner from a design perspective although we will probably then lose the ability to do function overloading for methods (which might be fine since it seems unlikely that we'd want to).

tripy/tripy/common/device.py
```diff
@@ -36,7 +36,7 @@ class device:
     kind: str
     index: int
 
-    def __init__(self, device: str) -> None:
+    def __init__(self, device: Union[str, Tuple[str, int]]) -> None:
```
pranavm-nvidia, line 39: Is there a way we could bypass the type checks? Some of the constructors have backdoors like this which are not intended for the end user. It would be good if we could keep the type annotation the same but disable type checking when we use the internal paths. `Tensor` and `DimensionSize` have similar backdoors where the data can be `None`.

pranavm-nvidia: If it's too difficult to do that, we could maybe look at an alternate approach like using `__new__` in such cases.

tripy/tripy/frontend/trace/ops/storage.py
```diff
@@ -79,7 +79,7 @@ def str_skip_fields(self) -> Set[str]:
             return {"data"}
         return set()
 
-    def __eq__(self, other) -> bool:
+    def __eq__(self, other: "tripy.Storage") -> bool:
```
pranavm-nvidia, line 82: Is this being added to the function registry or are you just adding the type annotation for readability? (It shouldn't be part of the function registry)

tripy/tripy/function_registry.py
```diff
@@ -371,6 +371,21 @@ def impl(func):
             # based on argument types.
             if isinstance(func, property):
                 self[key] = func
+            # For classes, we apply the wrapper to all methods.
+            elif inspect.isclass(func):
+                for member in dir(func):
+                    # Ignore properties and functions not defined in the class (we will use the presence of a docstring as a proxy for that).
+                    # It does not suffice to check just that the method is inherited because some decorators like @dataclass add methods
+                    # that are not documented or annotated and do not use inheritance to do so.
+                    # We need to check hasattr first because dir can list some properties even if they're not defined (e.g.,  __abstractmethods__).
```
pranavm-nvidia, line 380: Would `inspect.getmembers(func, predicate=lambda obj: inspect.isfunction(obj) and obj.__doc__)` work better here?

tripy/tripy/frontend/trace/ops/storage.py
slyubomirsky: It was in the registry due to export.public_api and it gave an error without it.

tripy/tripy/common/device.py
slyubomirsky: The question would be how to encode those special cases. We could add an argument to the registry for exceptions, maybe.

tripy/tripy/frontend/trace/ops/storage.py
pranavm-nvidia: Why is this being exported? `Storage` is just a trace op, not a user-facing API.

tripy/tripy/function_registry.py
slyubomirsky: Oh that's cool. Yeah, it works and is more compact.

tripy/tripy/common/device.py
pranavm-nvidia: But we do want type checking to be enabled when a user calls the method. Maybe we can introduce a static method that calls `__new__` and bypasses `__init__` entirely.

tripy/tripy/utils/stack_info.py
```diff
@@ -116,7 +116,11 @@ def get_stack_info(include_code_index: int = None) -> StackInfo:
             column_range=None,
         )
         if source_info.module == tripy.function_registry.__name__ and source_info.function == "wrapper":
-            source_info._dispatch_target = frame.f_locals["key"]
+            # If it comes from a decorated function, the format will be Class.method and we should strip out the class name.
+            target_name = frame.f_locals["key"]
+            if "." in frame.f_locals["key"]:
+                target_name = target_name.split(".")[-1]
+            source_info._dispatch_target = target_name
```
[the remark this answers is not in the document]
slyubomirsky: I tried changing it only there but `dispatch target` is what is used to match stack frames for generating error messages. That could be solved by stripping out the classname where that happens (in `_add_column_info`) too, so I'll try it there as well.

tripy/tests/frontend/test_utils.py
slyubomirsky: The second approach would probably be reasonable since the tensor methods we would register would also be in the public function registry. We would also still be able have overloading by registering them under the same key in the public function registry.

I'll see if there would be a simple way to avoid registering in two registries (this would build in an assumption that we would want functions to be members of only one registry at a time).

tripy/tripy/frontend/trace/ops/storage.py
slyubomirsky: Looks like I was mistaken. I was getting errors from the registry and saw this listed in the stack trace when I was debugging, but it's indeed not exported and it did not cause errors to remove it.

tripy/tests/frontend/test_utils.py
slyubomirsky: I went with the approach of turning the tensor method registry into an ordinary dict because I didn't want to introduce convoluted logic for trying to prioritize which registry should be the one that checks overloads. This causes few problems.

Incidentally, the reason for changing the stack depth here was not because of the second registry; it was because `__init__` previously was not registered but now is. However, I checked with other registered methods and confirmed that there were two registry checks without the changes in `[ceefce7](https://github.com/NVIDIA/TensorRT-Incubator/pull/388/commits/ceefce7409bc7efc2ed03e517086976a5e2cc4b4)`.

tripy/tripy/common/device.py
slyubomirsky: Went with the static method approach, ends up being very simple.

tripy/tripy/utils/stack_info.py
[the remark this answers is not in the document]
slyubomirsky: Changing it in two places sufficed.

slyubomirsky:
The test failure on Tripy overhead might be legitimate, since this would add a call into the registry during the `__init__` for a tensor. If there's no objection, I can increase the threshold for it, or we could make an exception for the tensor `__init__` if we really don't want any more overhead.

tripy/tripy/frontend/ops/registry.py
```diff
@@ -15,23 +15,30 @@
...
+    """
+    Decorator to add the method to the tensor method registry with the name specified.
+    This does not use the FunctionRegistry decorator because every tensor method would also be
+    registered in the public function registry and we would prefer to avoid having overhead
+    from having to dispatch overloads and check types twice.
+    """
 
-        return super().__call__(key)
+    # We make a special exception for "shape" since we actually do want that to be a property
+    allowed_methods = ["numpy", "cupy", "shape"]
```
pranavm-nvidia, line 35:
```suggestion
    allowed_methods = ["shape"]
```

pranavm-nvidia (approved):

pranavm-nvidia:
> The test failure on Tripy overhead might be legitimate, since this would add a call into the registry during the `__init__` for a tensor. If there's no objection, I can increase the threshold for it, or we could make an exception for the tensor `__init__` if we really don't want any more overhead.

@slyubomirsky Can we add something like `create_directly` for the Tensor class too? That could let us bypass the type checking in the critical path (i.e. `executable.py`).

tripy/tripy/frontend/dimension_size.py
```diff
@@ -35,9 +35,17 @@ def __init__(self, data: int, name: Optional[str] = None) -> None:
             name: An optional name.
         """
         # NOTE: We do not use isinstance here because bool is a subclass of int.
-        assert data is None or type(data) is int, "DimensionSize can only be created from integers"
+        assert type(data) is int, "DimensionSize can only be created from integers"
```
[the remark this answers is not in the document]
slyubomirsky: Good point.

slyubomirsky:
There's still an extra function call that can cause the overhead perf test to sometimes be above the threshold, so I'll raise it slightly. I can take out one of the function calls (not a registry call) if that adjustment is more than we'd want.

slyubomirsky:
That is curious. I am not getting a performance difference locally. I will see if it's consistent.

slyubomirsky:
There are about 200% more registry calls in that test, so that could explain the regression.

Number of registry calls normally:
```
{
    'arange': 30,
    'iota': 16,
    'concatenate': 62,
    'cast': 46,
    'full': 31,
    '__mul__': 15,
    '__rmul__': 15,
    '__add__': 15,
    '__radd__': 15,
    'reshape': 15,
    '__eq__': 30,
    'default_stream': 32,
    'compile': 1,
    'transpose': 15,
    'permute': 15,
    '__matmul__': 15
}
```

With these changes:
```
{
    'device.__init__': 2,
    'Linear.__init__': 15,
    'Linear.named_parameters': 105,
    'Module.named_parameters': 107,
    'arange': 30,
    'iota': 16,
    'concatenate': 62,
    'cast': 46,
    'full': 31,
    'Tensor.__mul__': 15,
    'Tensor.__add__': 15,
    'reshape': 15,
    'Parameter.__init__': 30,
    'DimensionSize.__eq__': 30,
    'Tensor.__eq__': 30,
    'default_stream': 32,
    'Stream.__init__': 1,
    'Stream.synchronize': 164,
    'InputInfo.__init__': 1,
    'Module.state_dict': 16,
    'Module.named_children': 16,
    'Linear.state_dict': 15,
    'Linear.named_children': 15,
    'compile': 1,
    'Linear.__call__': 15,
    'transpose': 15,
    'permute': 15,
    'Tensor.__matmul__': 15,
    'Executable.__call__': 133
}
```

Since we probably would not expect to overload most of these registered classes, perhaps it would make sense not to have all the registry calls for some of these classes?

pranavm-nvidia:
@slyubomirsky The registry provides other value besides overloading, like type checking, so I think it should be fine to keep those. Only the `Stream.synchronize()` and `Executable.__call__` calls should be on the hot path of the ones you listed. For those it would make sense to omit them from the registry or at least bypass the type checking. Everything else is only relevant at compile time and not for inference.

slyubomirsky:
I added those exceptions and it seems to have worked out here. Very interesting to see how these lookups can add up!

tripy/tripy/frontend/tensor.py
```diff
@@ -98,7 +98,34 @@ def __init__(
...
         """
-
+        # We use None internally but users should not be permitted to do it
+        assert data is not None, "Data argument to Tensor must not be None"
+        Tensor.raw_init(self, data, dtype, device, name, fetch_stack_info)
+
+    # Left undocumented because this should only be used internally.
+    # Produces a new instance of a Tensor but avoids calling into the function registry, unlike the normal constructor.
+    @staticmethod
+    def create_directly(
```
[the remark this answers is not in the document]
slyubomirsky: No, as `__init__` only mutates the local state but does not return a value.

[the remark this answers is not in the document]
pranavm-nvidia: Can we do the reverse then? `create_directly` can call `Tensor.__init__` so we don't have `raw_init`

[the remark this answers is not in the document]
slyubomirsky: We could do this, but the registered `__init__` would do type checks so we wouldn't be able to pass `None` as the `data` parameter. If we wanted to set `data` to `None`, we would have to allocate something for `data` and then set the field to `None` separately (throwing away what we allocated).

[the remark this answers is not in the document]
pranavm-nvidia: Ah I see the issue. Could you add a comment explaining that? Makes sense to keep `raw_init`.

tripy/tests/performance/test_perf.py
```diff
@@ -125,7 +125,7 @@ def func({arg_str}):
 
         return (end - start) / (iterations * 1000.0)
 
-    assert measure_overhead(1) < 60.0
+    assert measure_overhead(1) < 65.0
```
pranavm-nvidia, line 128: Is this still needed? Ideally the inference overhead is unchanged

tripy/tripy/frontend/module/parameter.py
```diff
@@ -49,7 +49,7 @@ def __init__(self, tensor: Any) -> None:
         t = tensor
         # for convenience, this will convert other dlpack-supporting representations too
         if not isinstance(t, Tensor):
-            t = Tensor(t)
+            t = Tensor.create_directly(t)
```
pranavm-nvidia, line 52: Just curious, do we not want type checking here?

tripy/tripy/frontend/trace/ops/shape.py
```diff
@@ -20,7 +20,7 @@
 
 from tripy import constraints
 from tripy.common.datatype import DATA_TYPES
-from tripy.frontend.ops.registry import TENSOR_METHOD_REGISTRY
+import tripy.frontend.ops.registry as registry
```
pranavm-nvidia, line 23: nit: we do:
```py
from tripy.frontend.ops.registry import register_tensor_method
```
everywhere else

tripy/tripy/frontend/utils.py
```diff
@@ -35,7 +35,7 @@ def tensor_from_shape_like(arg: ShapeLike) -> "tripy.Tensor":
     from tripy.frontend.trace.ops.reshape import Reshape
 
     if not arg:
-        return Tensor([], dtype=int32)
+        return Tensor.create_directly([], dtype=int32)
```
pranavm-nvidia, line 38: Are these just for performance?

tripy/tripy/function_registry.py
```diff
@@ -356,23 +356,54 @@ def raise_overload_error(msg, candidate_overloads, mismatch_reasons=None, extra_
 
         raise_overload_error("Could not find an implementation", self.overloads[key], mismatch_reasons)
 
-    def __call__(self, key: Any):
+    def __call__(self, key: Any, bypass_dispatch: Optional[Union[bool, Sequence[str]]] = None):
         """
         Registers a function with this function registry.
```
pranavm-nvidia, line 361: Can we mention that this works for *documented* methods of classes as well?

tripy/tripy/function_registry.py
```diff
@@ -356,23 +356,54 @@ def raise_overload_error(msg, candidate_overloads, mismatch_reasons=None, extra_
...
+                    bypass_func = bypass_dispatch if isinstance(bypass_dispatch, bool) else name in bypass_dispatch
+                    setattr(func, name, self(f"{key}.{name}", bypass_dispatch=bypass_func)(member))
+                self[key] = func
             else:
+                if bypass_dispatch and key in self:
+                    from tripy.common.exception import raise_error
+
+                    raise_error(
+                        f"Attempting to add key {key} into a function registry with dispatch disabled, but there is already an overload present."
+                    )
```
pranavm-nvidia, line 398: This should be an `assert` since it's not a user-facing error.

tripy/tripy/function_registry.py
```diff
@@ -385,6 +416,10 @@ def wrapper(*args, **kwargs):
 
                     self[key] = wrapper
                 else:
+                    assert not hasattr(
```
pranavm-nvidia, line 419: Could you also add a test where we bypass dispatch for a function and then register an overload?

tripy/tests/performance/test_perf.py
slyubomirsky: I'll double-check.

tripy/tripy/frontend/module/parameter.py
slyubomirsky: For internal uses, we probably don't want more registry calls than we have to but there's no inherent reason not to use ordinary constructor calls.

tripy/tripy/frontend/utils.py
slyubomirsky: Yeah, I figured most internal uses of the constructor shouldn't need to go through the registry.

tripy/tripy/frontend/module/parameter.py
pranavm-nvidia: Yup, makes sense.

tripy/tripy/frontend/utils.py
pranavm-nvidia: Yup, makes sense.

tripy/tests/performance/test_perf.py
slyubomirsky: Works for me locally with the original value, so I hope it will hold up that way in the CI.

tripy/tripy/function_registry.py
slyubomirsky: Added a line to the docstring, good catch.

tripy/tripy/function_registry.py
slyubomirsky: Good point.

tripy/tripy/function_registry.py
slyubomirsky: Added.

tripy/tripy/frontend/trace/ops/slice.py
```diff
@@ -271,19 +272,26 @@ def slice_helper(tensor, *slice_params: TensorLike):
     frame_index = -1
     assert slice_params
 
+    # Internal WAR: the constraints decorator is applied before the function registry decorator, so in the constraints tests,
+    # we will not find the Tensor.__getitem__ decorator. We can use a fallback in that case.
```
slyubomirsky, line 276: This was a bit of a wacky interaction with #333. FWIW, the previous approach of "find the call to `__getitem__` and skip one more stack entry" would avoid the need for a WAR here. The fundamental issue is the order in which decorators are applied.

tripy/tripy/frontend/module/parameter.py
slyubomirsky: That is true, we can call into the registry for this one.

slyubomirsky:
I think the regression in the benchmark is due to calling into the registry during the constructor for `tp.Parameter`. I will verify that and see if building the parameter outside the timed section gets rid of it or if we'll need to add other exceptions.

pranavm-nvidia (approved):

PR #391, opened 2024-11-19
[mlir-tensorrt] Add TensorRT 8.6 support

shelkesagar29:
This PR makes the following changes,
    - Make TensorRT 10.5 as default version
    - Add TensorRT 8.6 download support
    - Add TensorRT 8.6 to CI
    - TensorRT 9 checks from CI are removed to deal with
    device space error.
    - Fix tests to support above changes


shelkesagar29:
@christopherbate TRT 9 is remove from CI.
Running TRT10, TRT10 ASAN, TRT9 and TRT8 results in device running out of space.
TRT10 covers TRT9 so we should be good

.github/workflows/mlir-tensorrt-ci.yml
```diff
@@ -235,7 +235,7 @@ jobs:
               -DCMAKE_BUILD_TYPE=RelWithDebInfo \
               -DMLIR_TRT_PACKAGE_CACHE_DIR=/.cache.cpm \
               -DMLIR_TRT_ENABLE_ASSERTIONS=ON \
-              -DMLIR_TRT_DOWNLOAD_TENSORRT_VERSION=9.2.0.5 \
+              -DMLIR_TRT_DOWNLOAD_TENSORRT_VERSION=8.6.1.6 \
```
christopherbate, line 238: Looks like this won't work

mlir-tensorrt/build_tools/cmake/Dependencies.cmake
```diff
@@ -100,6 +104,28 @@ function(download_tensorrt)
 
   set(TRT_VERSION "${ARG_VERSION}")
 
+  # Handle TensorRT 8 versions. These are publicly accessible download links.
```
christopherbate, line 107: This is for TRT9

christopherbate (changes requested):
I think we also discussed splitting out the existing test cases that fail with TRT 10 so that we can test all common cases using "tensorrt-major-version=8" as well.

mlir-tensorrt/build_tools/cmake/Dependencies.cmake
christopherbate: I was wrong, thanks for figuring this out

.github/workflows/mlir-tensorrt-ci.yml
shelkesagar29: `cmake --preset ninja-llvm -DMLIR_TRT_DOWNLOAD_TENSORRT_VERSION=8.6.1.6`
does download TRT 8.6.1.6.

mlir-tensorrt/build_tools/cmake/Dependencies.cmake
shelkesagar29: This also works for TRT 8 versions with slight modifications.
Right now I am allowing only TRT 8.6.1.6

PR #396, opened 2024-11-20
Exposes base `OptionsContext` class in the API

pranavm-nvidia:
In order to support more than just the StableHloToExecutable pipeline, we need to be able to create different option types from the API. This commit exposes the base `OptionsContext` class in the Python API and includes a mechanism for child classes to register themselves with the client, allowing them to be created through a common API.

### DIFF
--- a/mlir-tensorrt/compiler/include/mlir-tensorrt-c/Compiler/Compiler.h
+++ b/mlir-tensorrt/compiler/include/mlir-tensorrt-c/Compiler/Compiler.h
@@ -52,6 +52,29 @@ static inline bool mtrtCompilerClientIsNull(MTRT_CompilerClient options) {
 }
 
 //===----------------------------------------------------------------------===//
+// MTRT_OptionsContext
+//===----------------------------------------------------------------------===//
+
+typedef struct MTRT_OptionsContext {
+  void *ptr;
+} MTRT_OptionsContext;
+
+MLIR_CAPI_EXPORTED MTRT_Status mtrtOptionsContextCreateFromArgs(
+    MTRT_CompilerClient client, MTRT_OptionsContext *options,
+    MlirStringRef optionsType, const MlirStringRef *argv, unsigned argc);
+
+MLIR_CAPI_EXPORTED void mtrtOptionsContextPrint(MTRT_OptionsContext options,
+                                                MlirStringCallback append,
+                                                void *userData);
+
+MLIR_CAPI_EXPORTED MTRT_Status
+mtrtOptionsContextDestroy(MTRT_OptionsContext options);
+
+static inline bool mtrtOptionsConextIsNull(MTRT_OptionsContext options) {
+  return !options.ptr;
+}
+
+//===----------------------------------------------------------------------===//
 // MTRT_StableHLOToExecutableOptions
 //===----------------------------------------------------------------------===//
 
--- a/mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsRegistry.h
+++ b/mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsRegistry.h
@@ -1,0 +1,87 @@
+//===- OptionsRegistry.h -----------------------------------------*- C++-*-===//
+//
+// SPDX-FileCopyrightText: Copyright 2024 NVIDIA CORPORATION & AFFILIATES.
+// All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//===----------------------------------------------------------------------===//
+///
+/// This is the top-level C++ interface for clients that wish to compile
+/// MLIR programs from a supported input (e.g. StableHLO) into a supported
+/// target (e.g. TensorRT engine). This file contains just the declarations for
+/// the CompilerClient object, see below for details.
+///
+//===----------------------------------------------------------------------===//
+#ifndef MLIR_TENSORRT_COMPILER_OPTIONS_REGISTRY
+#define MLIR_TENSORRT_COMPILER_OPTIONS_REGISTRY
+
+#include "mlir-tensorrt-dialect/Utils/Options.h"
+#include "mlir-tensorrt/Compiler/Client.h"
+#include "mlir-tensorrt/Dialect/Plan/IR/Plan.h"
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/StringRef.h"
+#include <functional>
+
+namespace mlirtrt::compiler {
+
+using OptionsConstructorFuncT =
+    std::function<StatusOr<std::unique_ptr<mlir::OptionsContext>>(
+        const CompilerClient &client, const llvm::ArrayRef<llvm::StringRef>)>;
+
+/// Registers an options creation function for a specific options type.
+void registerOption(const llvm::StringRef optionsType,
+                    OptionsConstructorFuncT func);
+
+/// Creates an options instance for the specified options type using a creation
+/// function that was previously registered.
+StatusOr<std::unique_ptr<mlir::OptionsContext>>
+createOptions(const CompilerClient &client, const llvm::StringRef optionsType,
+              const llvm::ArrayRef<llvm::StringRef> args);
+
+/// Helper to build callbacks that can create options.
+template <typename OptionsT, typename TaskT>
+StatusOr<std::unique_ptr<mlir::OptionsContext>>
+optionsCreateFromArgs(const CompilerClient &client,
+                      const llvm::ArrayRef<llvm::StringRef> args) {
+  // Load available extensions.
+  mlir::MLIRContext *context = client.getContext();
+  mlir::plan::PlanDialect *planDialect =
+      context->getLoadedDialect<mlir::plan::PlanDialect>();
+  compiler::TaskExtensionRegistry extensions =
+      planDialect->extensionConstructors.getExtensionRegistryForTask<TaskT>();
+
+  auto result = std::make_unique<OptionsT>(std::move(extensions));
+
+  std::string err;
+  if (failed(result->parse(args, err))) {
+    return getInternalErrorStatus(
+        "failed to parse options string \"{0:$[ ]}\" due to error {1}",
+        llvm::iterator_range(args), err);
+  }
+
+  // TODO: Figure out whether to add a method in the base class like
+  // "finalizeOptions" or a callback here, or something else if
+  // `inferDeviceOptionsFromHost` is unique to StableHLO.
+  //
+  // Populate device options from host information.
+  Status inferStatus = result->inferDeviceOptionsFromHost();
+  if (!inferStatus.isOk())
+    return inferStatus;
+
+  return std::unique_ptr<mlir::OptionsContext>(result.release());
+}
+} // namespace mlirtrt::compiler
+
+#endif
--- a/mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/StableHloToExecutable.h
+++ b/mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/StableHloToExecutable.h
@@ -189,6 +189,9 @@ public:
                                const StableHLOToExecutableOptions &options);
 };
 
+/// Register the task/options with the client's registry.
+void registerStableHloToExecutableTask();
+
 //===----------------------------------------------------------------------===//
 // Pipeline Registrations
 //===----------------------------------------------------------------------===//
--- a/mlir-tensorrt/compiler/lib/CAPI/Compiler/Compiler.cpp
+++ b/mlir-tensorrt/compiler/lib/CAPI/Compiler/Compiler.cpp
@@ -26,11 +26,14 @@
 #include "mlir-c/Support.h"
 #include "mlir-executor-c/Support/Status.h"
 #include "mlir-tensorrt-dialect/Target/TranslateToTensorRT.h"
+#include "mlir-tensorrt-dialect/Utils/Options.h"
 #include "mlir-tensorrt/Compiler/Extension.h"
+#include "mlir-tensorrt/Compiler/OptionsRegistry.h"
 #include "mlir-tensorrt/Compiler/StableHloToExecutable.h"
 #include "mlir-tensorrt/Compiler/TensorRTExtension/TensorRTExtension.h"
 #include "mlir-tensorrt/Dialect/Plan/IR/Plan.h"
 #include "mlir/CAPI/IR.h"
+#include "mlir/CAPI/Utils.h"
 #include "llvm/ADT/StringExtras.h"
 
 using namespace mlirtrt;
@@ -44,6 +47,7 @@ using namespace mlir;
 DEFINE_C_API_PTR_METHODS(MTRT_CompilerClient, CompilerClient)
 DEFINE_C_API_PTR_METHODS(MTRT_StableHLOToExecutableOptions,
                          StableHLOToExecutableOptions)
+DEFINE_C_API_PTR_METHODS(MTRT_OptionsContext, OptionsContext)
 #if defined(__GNUC__) || defined(__clang__)
 #pragma GCC diagnostic pop
 #endif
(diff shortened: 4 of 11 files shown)


mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/Client.h
```diff
@@ -145,8 +172,43 @@ class CompilerClient {
...
+    return getInternalErrorStatus(
+        "failed to parse options string {0} due to error: {1}", line, err);
+  }
+
+  // TODO: Figure out whether to add a method in the base class like
+  // "finalizeOptions" or a callback here, or something else if
+  // `inferDeviceOptionsFromHost` is unique to StableHLO.
+  //
+  // Populate device options from host information.
+  Status inferStatus = result->inferDeviceOptionsFromHost();
```
pranavm-nvidia, line 205: Not sure if `inferDeviceOptionsFromHost` is specific to the StableHLO options or not. If it is, I can add a method in the base class to allow child classes to do arbitrary custom things here.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/StableHloToExecutable.h
```diff
@@ -189,6 +189,13 @@ class StableHloToExecutableTask
                                const StableHLOToExecutableOptions &options);
 };
 
+// TODO: Figure out if we want to register in a different way.
+static const bool registeredStableHloToExecutable =
```
pranavm-nvidia, line 193: Lots of ways that we could implement the registration mechanism. I went with a straightforward approach here but we can do something better/cleaner if preferred.

mlir-tensorrt/python/bindings/Compiler/CompilerPyBind.cpp
```diff
@@ -240,6 +251,29 @@ PYBIND11_MODULE(_api, m) {
...
+             MTRT_Status s = mtrtOptionsContextCreateFromArgs(
+                 client, &options, optionsType.c_str(), refs.data(),
+                 refs.size());
+             THROW_IF_MTRT_ERROR(s);
+             return new PyOptionsContext(options);
+           }),
+           py::arg("client"), py::arg("options_type"), py::arg("args"))
+
+      .def("print", [](PyOptionsContext &self) {
+        // TODO: Should this be exposed via __str__ instead?
```
pranavm-nvidia, line 272: Might be cleaner to expose this as a `__str__` method, but this is primarily exposed for testing purposes, so not sure if it matters too much.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/Client.h
```diff
@@ -145,8 +172,43 @@ class CompilerClient {
   /// used to create the PM.
   llvm::DenseMap<PassManagerKey, std::unique_ptr<CompilationTaskBase>>
       cachedPassManagers;
+
+  static llvm::ManagedStatic<llvm::StringMap<OptionsConstructorFuncT>> registry;
```
christopherbate, line 176: This shouldn't be exposed in the header. It can be a global static variable in the C++ file.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/Client.h
```diff
@@ -131,6 +147,17 @@ class CompilerClient {
   static void setupPassManagerLogging(mlir::PassManager &pm,
                                       const DebugOptions &options);
 
+  StatusOr<std::unique_ptr<mlir::OptionsContext>>
+  createOptions(const std::string &optionsType,
```
christopherbate, line 151: - The LLVM/MLIR conventions that we follow encourage using `llvm::StringRef` (by value) in place of `const std::string &`.
- Similarly, please use `llvm::ArrayRef<T>` (by value) in place of `const std::vector<T>&`
- Please move implementation to header file.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/Client.h
```diff
@@ -95,6 +101,16 @@ class CompilationTask : public CompilationTaskBase {
 /// builder caches are being persisted to disk.
 class CompilerClient {
 public:
+  using OptionsConstructorFuncT =
+      std::function<StatusOr<std::unique_ptr<mlir::OptionsContext>>(
+          const CompilerClient &client, const std::vector<llvm::StringRef> &)>;
+
+  static bool registerOption(std::string optionsType,
```
christopherbate, line 108: - Please move implementation to C++ file.
- Please us `llvm::StringRef` in place of `std::string`.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/Client.h
```diff
@@ -95,6 +101,16 @@ class CompilationTask : public CompilationTaskBase {
 /// builder caches are being persisted to disk.
 class CompilerClient {
 public:
+  using OptionsConstructorFuncT =
+      std::function<StatusOr<std::unique_ptr<mlir::OptionsContext>>(
```
christopherbate, line 105: Please use `llvm::ArrayRef<T>` by value in place of `const std::vector<T>&`.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/Client.h
```diff
@@ -145,8 +172,43 @@ class CompilerClient {
...
+  mlir::plan::PlanDialect *planDialect =
+      context->getLoadedDialect<mlir::plan::PlanDialect>();
+  compiler::TaskExtensionRegistry extensions =
+      planDialect->extensionConstructors.getExtensionRegistryForTask<TaskT>();
+
+  auto result = std::make_unique<OptionsT>(std::move(extensions));
+
+  std::string err;
+  if (failed(result->parse(args, err))) {
+    std::string line = llvm::join(args, " ");
```
christopherbate, line 195: The two lines in the body can be simplified to:

```
return getInternalErrorStatus("failed to parse options string \"{0:$[ ]}\" due to error {1}", llvm::iterator_range(args), err);
```

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/Client.h
christopherbate: The "device" options (e.g. compute capability, max shared mem, etc) and the population of the device options  from local CUDA device should be moved to the common base class.

However, we shouldn't unconditionally invoke `inferDeviceOptionsFromHost()`. That should be a behavior that is moved into the base class constructor. Currently I think we are using a flag `--infer-device-options-from-host`, but we're not doing any validation to check whether the user passed both `--infer-device-options-from-host` as well as `--device-compute-capability=XY`. It may be better to remove `--infer-device-options-from-host` and instead have that be signaled by `--device-compute-capability=0` (which can be the default value).

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/Client.h
```diff
@@ -145,8 +172,43 @@ class CompilerClient {
...
+  // TODO: Figure out whether to add a method in the base class like
+  // "finalizeOptions" or a callback here, or something else if
+  // `inferDeviceOptionsFromHost` is unique to StableHLO.
+  //
+  // Populate device options from host information.
+  Status inferStatus = result->inferDeviceOptionsFromHost();
+  if (!inferStatus.isOk())
+    return inferStatus;
+
+  return std::unique_ptr<mlir::OptionsContext>(result.release());
```
christopherbate, line 209: I think you should be able to just `return result` or `return std::move(result)`?

mlir-tensorrt/compiler/lib/Compiler/Client.cpp
```diff
@@ -46,6 +46,8 @@ CompilationTaskBase::~CompilationTaskBase() {}
 // CompilerClient
 //===----------------------------------------------------------------------===//
 
+decltype(CompilerClient::registry) CompilerClient::registry = {};
```
christopherbate, line 49: static registry can just be defined here as a static global

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/StableHloToExecutable.h
christopherbate: this definitely shouldn't be in a header file, that will break some build configs.

We should just do it like this:

In the header file do

```
void registerStableHloToExecutableTask();
```

In CPP file:

```C++
void compiler::registerStableHloToExecutableTask() {
   CompilerClient::registerOption(
        "stable-hlo-to-executable",
        optionsCreateFromArgs<StableHLOToExecutableOptions,
                              StableHloToExecutableTask>);
}
```

Now you need to invoke `registerStableHloToExecutableTask` somewhere. The easiest thing to do is just grep for `registerStablehloClusteringPipelines` and call `registerStableHloToExecutableTask` there as well.

mlir-tensorrt/compiler/include/mlir-tensorrt-c/Compiler/Compiler.h
```diff
@@ -51,6 +51,28 @@ static inline bool mtrtCompilerClientIsNull(MTRT_CompilerClient options) {
...
+typedef struct MTRT_OptionsContext {
+  void *ptr;
+} MTRT_OptionsContext;
+
+MLIR_CAPI_EXPORTED MTRT_Status mtrtOptionsContextCreateFromArgs(
+    MTRT_CompilerClient client, MTRT_OptionsContext *options,
+    const char *optionsType, const MlirStringRef *argv, unsigned argc);
+
+MLIR_CAPI_EXPORTED MTRT_Status
+mtrtOptionsContextPrint(MTRT_OptionsContext options);
```
christopherbate, line 67: There's a simple uniform convention we use for print methods. Just look for the other C API methods that accept a `MlirStringCallback`.

Here's how to do it:

```
MLIR_CAPI_EXPORTED void mlirOptionsContextPrint(MTRT_OptionsContext options, MlirStringCallback append, void *userCtx);
```

In CAPI implementation:

```C++
void mlirOptionsContextPrint(MTRT_OptionsContext options, MlirStringCallback append, void *userCtx) { 
    mlir::detail::CallbackOstream stream(callback, userData);
    unwrap(options)->print(stream);
}
```

In PyBind:

```C++
.def("__repr__", [](PyOptionsContext &ctx)  {
               PyPrintAccumulator printAccum;
             printAccum.parts.append("Options(");
             mtrtOptionsContextPrint(self, printAccum.getCallback(),
                                 printAccum.getUserData());
             printAccum.parts.append(")");
             return printAccum.join();
}
```

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/Client.h
```diff
@@ -28,14 +28,20 @@
...
 #include "mlir/Pass/PassManager.h"
 #include "mlir/Support/TypeID.h"
 #include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/Hashing.h"
+#include "llvm/ADT/StringRef.h"
 #include "llvm/Support/ErrorHandling.h"
+#include "llvm/Support/ManagedStatic.h"
+#include <functional>
 #include <memory>
+#include <unordered_map>
```
christopherbate, line 44: Unordered_map can be dropped

mlir-tensorrt/compiler/lib/CAPI/Compiler/Compiler.cpp
```diff
@@ -99,6 +101,37 @@ MTRT_Status mtrtCompilerClientDestroy(MTRT_CompilerClient client) {
...
+  return mtrtStatusGetOk();
+}
+
+MLIR_CAPI_EXPORTED MTRT_Status
+mtrtOptionsContextPrint(MTRT_OptionsContext options) {
+  unwrap(options)->print(llvm::outs());
+  return mtrtStatusGetOk();
+}
+
+MLIR_CAPI_EXPORTED MTRT_Status
```
christopherbate, line 129: This can just return `void`.

mlir-tensorrt/python/bindings/Compiler/CompilerPyBind.cpp
christopherbate: Commented above on methodology the CAPI declaration

christopherbate (approved):
Overall it looks good. Noted in the comments a couple things that need to be tweaked before merging, but once that's done and it passes CI, then it should be good to merge.

One thing to note is that we currently have the CI building static libs. Since you're introducing static registration mechanisms, you definitely want to check the CMake build with `-DBUILD_SHARED_LIBS=ON` as well. Let me know if any questions.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/Client.h
pranavm-nvidia: It's a pointer to the derived type, so I get:
```
Client.h:197:10: error: no viable conversion from returned value of type 'typename _MakeUniq<StableHLOToExecutableOptions>::__single_object' (aka 'unique_ptr<mlirtrt::compiler::StableHLOToExecutableOptions>') to function return type 'StatusOr<std::unique_ptr<mlir::OptionsContext>>'
```

mlir-tensorrt/compiler/lib/CAPI/Compiler/Compiler.cpp
pranavm-nvidia: Sure, do you want me to change the other destroy functions as well?

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/Client.h
pranavm-nvidia: I get an error with this:
> static void llvm::format_provider<llvm::StringRef>::format(const T &, llvm::raw_ostream &, StringRef) [T = llvm::StringRef, Enable = void]: Assertion `false && "Style is not a valid integer"' failed.

I can still merge the two lines though.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/Client.h
```diff
@@ -85,6 +89,19 @@ class CompilationTask : public CompilationTaskBase {
...
+    std::function<StatusOr<std::unique_ptr<mlir::OptionsContext>>(
+        const CompilerClient &client, const llvm::ArrayRef<llvm::StringRef>)>;
+
+void registerOption(const llvm::StringRef optionsType,
+                    OptionsConstructorFuncT func);
+
+StatusOr<std::unique_ptr<mlir::OptionsContext>>
+createOptions(const CompilerClient &client, const llvm::StringRef optionsType,
+              const llvm::ArrayRef<llvm::StringRef> args);
+
```
pranavm-nvidia, line 104: Do we want to move all of this to a separate header? There's really nothing tying it to the client anymore.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsRegistry.h
```diff
@@ -0,0 +1,85 @@
+//===- OptionsRegistry.h -------------------------------------------------*- C++
+//-*-===//
```
christopherbate, line 2: Nit: please fix formatting

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsRegistry.h
```diff
@@ -0,0 +1,85 @@
...
+  auto result = std::make_unique<OptionsT>(std::move(extensions));
+
+  std::string err;
+  if (failed(result->parse(args, err))) {
+    return getInternalErrorStatus(
+        "failed to parse options string \"{0:$[ ]}\" due to error {1}",
+        llvm::iterator_range(args), err);
+  }
+
+  // TODO: Figure out whether to add a method in the base class like
```
christopherbate, line 72: OK I see what you are saying. The intended behavior is that in the constructor of the options object, it should check whether `shouldInferDeviceOptionsFromHost` is set. If so, then invoke this method. Otherwise, don't invoke this method.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsRegistry.h
```diff
@@ -0,0 +1,85 @@
...
+#include "llvm/ADT/StringRef.h"
+#include <functional>
+
+namespace mlirtrt::compiler {
+
+using OptionsConstructorFuncT =
+    std::function<StatusOr<std::unique_ptr<mlir::OptionsContext>>(
+        const CompilerClient &client, const llvm::ArrayRef<llvm::StringRef>)>;
+
+void registerOption(const llvm::StringRef optionsType,
```
christopherbate, line 44: Nit: please add documentation comment

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsRegistry.h
```diff
@@ -0,0 +1,85 @@
...
+/// This is the top-level C++ interface for clients that wish to compile
+/// MLIR programs from a supported input (e.g. StableHLO) into a supported
+/// target (e.g. TensorRT engine). This file contains just the declarations for
+/// the CompilerClient object, see below for details.
+///
+//===----------------------------------------------------------------------===//
+#ifndef MLIR_TENSORRT_COMPILER_OPTIONS_REGISTRY
+#define MLIR_TENSORRT_COMPILER_OPTIONS_REGISTRY
+
+#include "Client.h"
```
christopherbate, line 31: The include should include the path starting from `"mlir-tensorrt/....`

mlir-tensorrt/compiler/lib/CAPI/Compiler/Compiler.cpp
christopherbate: Try to keep your changes concentrated just to the scope of this PR

mlir-tensorrt/compiler/lib/Compiler/OptionsRegistry.cpp
```diff
@@ -0,0 +1,40 @@
+//===- OptionsRegistry.cpp
+//---------------------------------------------------------===//
```
christopherbate, line 2: Nit: please fix formatting

mlir-tensorrt/compiler/lib/Compiler/OptionsRegistry.cpp
```diff
@@ -0,0 +1,40 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "mlir-tensorrt/Compiler/OptionsRegistry.h"
+#include "llvm/Support/ManagedStatic.h"
+
+namespace mlirtrt::compiler {
+static llvm::ManagedStatic<llvm::StringMap<OptionsConstructorFuncT>> registry{};
```
christopherbate, line 24: The convention we use everywhere else (and in upstream MLIR) is to 

```
using namespace mlitrt::compiler;
```

and then implementations are given here as 

```
void compiler::registerOption(....) { 
   ...
}

```

christopherbate (approved):
LGTM except for some additional minor comments

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/Client.h
pranavm-nvidia: As discussed in the meeting, I can do this in a follow-up PR.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsRegistry.h
pranavm-nvidia: That makes sense. As we discussed in the meeting I can make this change in the next PR where I will implement the TRT options/task.

mlir-tensorrt/compiler/lib/CAPI/Compiler/Compiler.cpp
pranavm-nvidia: I can change this in a follow-up PR if desired.

Issue #378, opened 2024-11-15
Segfault in prompt_encoder


yizhuoz004:
[prompt_encoder.txt](https://github.com/user-attachments/files/17760086/prompt_encoder.txt)

shelkesagar29:
This is issue in upstream stablehlo pass.
It converts the following IR 
```
%3 = stablehlo.subtract %c_19, %c_37 : (tensor<1x1xi32>, tensor<1x1xi32>) -> tensor<?x?xi32>
  %4 = plan.with_shape %3(%c1_i32, %c1_i32) : (tensor<?x?xi32>, i32, i32) -> tensor<?x?xi32>
```
to
```
%57 = "stablehlo.constant"() <{value = dense<-1> : tensor<1x1xi32>}> : () -> tensor<1x1xi32>
  %58 = "plan.with_shape"(%57, %20, %20) : (tensor<1x1xi32>, i32, i32) -> tensor<?x?xi32>
```

However, if op is replaced with constant and consumer is non stablehlo op, `tensor.cast` should be used which is missing.

shelkesagar29:
PR https://github.com/NVIDIA/TensorRT-Incubator/pull/401
provides Stablehlo patch that resolves this issue.
I tested IR on `SHLO -> Executable` pipeline and it runs successfully.

PR #401, opened 2024-11-22
Fix stablehlo constant folding bugs through patch

shelkesagar29:
This PR provides a patch to fix two issues in StableHlo
upstream,
- If some op is constant folded and op users are certain
stablehlo ops (which don't allow different type for operand
and result) OR ops with different dialect, `tensor.cast`
should be inserted.
MLIR-TensorRT constant folding pass does this but upstream
stablehlo pass doesn't do this, causing error. Some
logic from MLIR-TRT is patched into upstream.
- Upstream `ReshapeOpCanon` pattern to simplify
`stablehlo.reshape` op failed if input to reshape
is `DenseResourceElementsAttr`. This is fixed.
- Many patterns where replacing op with a new op,
without giving consideration to casting. This is 
fixed.

Solves: https://github.com/NVIDIA/TensorRT-Incubator/issues/378

### DIFF
--- a/mlir-tensorrt/CMakeLists.txt
+++ b/mlir-tensorrt/CMakeLists.txt
@@ -183,7 +183,9 @@ if(MLIR_TRT_ENABLE_HLO AND NOT TARGET StablehloOps)
     VERSION 1.6.4
     GIT_TAG 1456dfa1e1a83aab0cc717714ba3695886f60302
     GIT_REPOSITORY "https://github.com/openxla/stablehlo.git"
-    PATCHES "${MLIR_TENSORRT_ROOT_DIR}/build_tools/stablehlo.patch"
+    PATCHES 
+      "${MLIR_TENSORRT_ROOT_DIR}/build_tools/stablehlo.patch"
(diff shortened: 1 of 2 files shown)


shelkesagar29:
@christopherbate helper functions are implemented in header itself to unblock Tripy for now.
As soon as you merge PR to update stablehlo, I will need to update this patch as well.
I will then move definitions  to `Utils.cpp` maybe.

christopherbate:
Can you verify that `ninja -C build check-stablehlo-tests` passes? Can that be added to the CI command as well?

If the StableHLO tests passes then we should be good to merge, although ideally we should be adding additional tests to verify the cases this solves (either in MLIR-TRT tests or in stablehlo via the patch)

christopherbate (approved):

shelkesagar29:
> Can you verify that `ninja -C build check-stablehlo-tests` passes? Can that be added to the CI command as well?
> 
> If the StableHLO tests passes then we should be good to merge, although ideally we should be adding additional tests to verify the cases this solves (either in MLIR-TRT tests or in stablehlo via the patch)

Yes, stablehlo tests do pass with changes. 
I will add tests for next patch that we will create for newer stablehlo version you will be updating.

PR #404, opened 2024-11-22
[compiler][conversion] Fix an issue in `stablehlo.slice` conversion

shelkesagar29:
This PR fixes an issue in `stablehlo.slice` conversion which popped when result type is dynamic.
In the converter, we use `tensorrt.slice` op builder that infers result based on `static_size` (output shape is same as size on TensorRT side) and `size` was taken as shape of stablehlo op output. This caused two issues,
- `truncateI64ToI32` failed for dynamic output because dynamic dim can't be truncated to i32 without loss.
- A new builder needed which could set result to dynamic shape even when size is static.

With this change,
- `size` is computed as `ceil(limit-start)/stride` (these are stablehlo slice op attributes) and NOT as shape of output.
- A new builder is added.

MLIR test is added.

### DIFF
--- a/mlir-tensorrt/compiler/lib/Conversion/StablehloToTensorRT/StablehloToTensorRT.cpp
+++ b/mlir-tensorrt/compiler/lib/Conversion/StablehloToTensorRT/StablehloToTensorRT.cpp
@@ -1931,13 +1931,13 @@ struct ConstantConverter : public ConvertHloOpToTensorRTPattern<HloOpType> {
 
 namespace {
 /// Convert `stablehlo.slice` op to `tensorrt.slice` operation.
-template <typename HloOpType>
-struct HloSliceConverter : public ConvertHloOpToTensorRTPattern<HloOpType> {
-  using ConvertHloOpToTensorRTPattern<HloOpType>::ConvertHloOpToTensorRTPattern;
-  LogicalResult matchAndRewrite(
-      HloOpType op,
-      typename ConvertHloOpToTensorRTPattern<HloOpType>::OpAdaptor adaptor,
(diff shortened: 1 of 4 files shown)


mlir-tensorrt/compiler/lib/Conversion/StablehloToTensorRT/StablehloToTensorRT.cpp
```diff
@@ -1947,19 +1947,26 @@ struct HloSliceConverter : public ConvertHloOpToTensorRTPattern<HloOpType> {
...
-    FailureOr<SmallVector<int32_t>> i32Shape =
-        truncateI64ToI32(loc, op.getType().getShape());
-    if (failed(i32Shape))
-      return rewriter.notifyMatchFailure(op,
-                                         "could not convert i64 shape to i32");
+
+    // Result shape is `ceil((limit_indices - start_indices)/stride)`
+    SmallVector<int32_t> i32Shape(limitIndices->size());
+    for (size_t i = 0; i < limitIndices->size(); i++) {
+      i32Shape[i] = std::ceil((((*limitIndices)[i] - (*startIndices)[i])) /
```
christopherbate, line 1964: I think you mean to use `llvm::divideCeil` here?

christopherbate (approved):

mlir-tensorrt/compiler/lib/Conversion/StablehloToTensorRT/StablehloToTensorRT.cpp
shelkesagar29: This is better

PR #414, opened 2024-11-26
Upgrade LLVM-Project, StableHLO dependencies 

christopherbate:
This is a combination of the following commits (from older to most recent):

## [compiler/Dialect/Plan] Improve efficiency of TensorRT clustering

Runs TensorRT analysis conversion on the container op and uses that result
in the clustering analysis instead of checking conversion op-by-op, which
will no longer work well in after an upcoming LLVM upgrade.

This change also removes the "tensorrt major version" parameter for the
`ClusterKindAttrInterface::getClusteringOpts` method arguments since that
information can be passed directly as a parameter of the `TensorRTClusterKindAttr``

## Upgrade all dependencies

This upgrades dependencies to the following base commits:

```
LLVM_COMMIT_UPSTREAM = "6c64c8a6f3f77c30745c751d4163ff6bf2fc323b"
TORCH_MLIR_COMMIT = "30c519369ed7eabad0282d0f874500a9b41fcbbd"
STABLEHLO_COMMIT = "6e403b1aa6a71f5eaa09cc720e4ad42f692745e6"
```

## Update how cluster kinds and their options are populated

This change updates the 'plan-stablehlo-clustering' pass so that it no longer
populates default "cluster kinds" if the 'plan.cluster_kinds' attribute is
missing on the module. Instead, it is up to the top-level compilation entrypoints
or the frontend to populate these key metadata attributes.

This allows for significant simplification in how options are passed to
various passes and pipelines. For example, the 'stablehlo-clustering' and
'plan-segmentation-pipeline' both no longer need to accept pass options
specifically for the TensorRT clustering options. These options are dictated
purely by the TensorRT clustering attribute attached to the module. Only the
top level compilation entrypoint requires such options, which it can use to
populate the clustering backend attributes parameters.

GitOrigin-RevId: 66aa8a1e07b4c94195ed27bcd96f7fd2d00c6f3e


shelkesagar29:
We need to update commit message with body same as description. Then it will pass CI.
Title we need to keep `Move internal changes` for CI automation purpose

shelkesagar29 (approved):
Please change commit message title to `Move internal changes` to allow CI automation.
Commit message body should tell about changes which it's doing here.

PR #417, opened 2024-11-28
Removes Parameter API

pranavm-nvidia:
Removes the Parameter API which was used to denote constants in computation graphs. We can instead use regular `Tensor`s since anything not specified as an input to the compile call is treated as a constant.

`DefaultParameter` is preserved as an internal implementation detail for efficiently defining expected parameter shapes/dtypes without materializing data.

### DIFF
--- a/tripy/tripy/frontend/module/module.py
+++ b/tripy/tripy/frontend/module/module.py
@@ -17,21 +17,47 @@
 
 import copy
 import operator
-from typing import Any, Dict, Iterator, List, Set, Tuple, TypeVar, Union
+from typing import Any, Dict, Iterator, List, Set, Tuple, Union
 
-from tripy import export
+from tripy import export, utils
 from tripy.common.exception import raise_error
-from tripy.frontend.module.parameter import Parameter
+from tripy.frontend.module.parameter import DefaultParameter
+from tripy.frontend.tensor import Tensor
+from tripy.function_registry import type_str_from_arg
 from tripy.logging import logger
-
-T = TypeVar("T")
 
 
 def _check_param_compatible(original_param, new_param, param_name):
-    if not isinstance(original_param, Parameter):
+    # We want to check the incoming parameter type even when the original parameter is not a tensor.
+    if not isinstance(new_param, Tensor):
+        raise_error(
+            "Unrecognized type for module parameter.",
+            f"Expected a tensor for parameter: '{param_name}', but got: {type_str_from_arg(new_param)}",
+        )
+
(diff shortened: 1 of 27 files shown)


tripy/tripy/frontend/module/parameter.py
```diff
@@ -80,18 +32,21 @@ def __init__(self, shape: Optional[Sequence[int]], dtype: "tripy.dtype") -> None
...
         if shape is None:
-            _is_shape_known = False
+            is_shape_known = False
             shape = []
-        super().__init__(reshape(arange(math.prod(shape), dtype), shape))
+        tensor = reshape(arange(math.prod(shape), dtype), shape)
+        self.__dict__ = tensor.__dict__
 
-        self._shape = shape
+        self._shape = list(shape)
```
slyubomirsky, line 42: When is the `list` conversion needed?

tripy/tripy/frontend/module/module.py
```diff
@@ -17,21 +17,47 @@
...
-    if not isinstance(original_param, Parameter):
+    # We want to check the incoming parameter type even when the original parameter is not a tensor.
+    if not isinstance(new_param, Tensor):
+        raise_error(
+            "Unrecognized type for module parameter.",
+            f"Expected a tensor for parameter: '{param_name}', but got: {type_str_from_arg(new_param)}",
+        )
+
+    if not isinstance(original_param, Tensor):
+        # Allow values to be initialized to non-tensor types and changed later.
```
slyubomirsky, line 39: When do we set it to a non-tensor type? Only for the initial value? (Is it `None`?) Seems like odd to have an exception to the invariant.

slyubomirsky (approved):
A very convenient simplification :+1:

tripy/tripy/frontend/module/parameter.py
pranavm-nvidia: Sometimes we initialize with a tuple

tripy/tripy/frontend/module/module.py
pranavm-nvidia: In Tripy, only when it's `None` (before the attribute has been set for the first time). But we don't want to be unnecessarily restrictive in user code. e.g. there's no reason to disallow:
```py
module.x = "example"
module.x = tp.Tensor(...)
```

slyubomirsky: That's true, perhaps we might find uses for non-tensor-valued properties.

PR #419, opened 2024-12-02
Adds an OptionsBundle class to allow options classes to subscribe to option groups

pranavm-nvidia:
Introduces an `OptionsBundle` class that will make it easier to share groups of
options across different options classes. This class essentially allows you to bundle
together option providers like `DebugOptions` so that they don't need to be managed
individually.

### DIFF
--- a/mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/StableHloToExecutable.h
+++ b/mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/StableHloToExecutable.h
@@ -33,6 +33,7 @@
 
 #include "mlir-executor/Runtime/API/API.h"
 #include "mlir-executor/Support/Status.h"
+#include "mlir-tensorrt-dialect/Utils/OptionsBundle.h"
 #include "mlir-tensorrt/Compiler/Client.h"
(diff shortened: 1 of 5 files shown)


mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/StableHloToExecutable.h
```diff
@@ -50,7 +51,7 @@ namespace mlirtrt::compiler {
 
 class StableHloToExecutableTask;
 
-struct StableHLOToExecutableOptions : public mlir::OptionsContext {
+struct StableHLOToExecutableOptions : public mlir::OptionsBundle<DebugOptions> {
```
pranavm-nvidia, line 54: If we prefer not to change the inheritance hierarchy, then I can make the `OptionsBundle` a member instead (that's how I have it in the first commit).

christopherbate:
This looks really good, thanks

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/StableHloToExecutable.h
pranavm-nvidia: I've squashed the commits into one.

Issue #408, opened 2024-11-26
Merge dtype constraints and convert_to_tensors decorators


pranavm-nvidia:
Our data type constraints decorator needs to be applied before `convert_to_tensors` so that the inputs it receives are have already been converted to tensors that it can retrieve data types correctly. 

Additionally, having multiple decorators complicates the stack information, making it very hard to follow. If we had one decorator that did tensor conversions and data type checks, it would resolve both of these issues. 

I considered merging all of this into the `export.public_api` decorator, but that won't work since we may need to apply this decorator to methods of classes (which we don't want to have to do for `public_api`).

slyubomirsky:
Implemented

PR #420, opened 2024-12-03
Combine the functionality of the `convert_to_tensors` decorator and the `dtypes` constraint.

slyubomirsky:
Address #408. Since all of the functions that used the `convert_to_tensors` decorator also applied a `dtypes` constraint, the approach here is to simply include the option to do conversions in the `dtypes` constraint.


tripy/tripy/utils/ast.py
```diff
@@ -123,9 +123,14 @@ def get_arg_candidate_column_offsets(
...
+                # For methods, the `self` argument is omitted from ast.Call.args
                 if "self" in arg_names:
-                    # For methods, the `self` argument is omited from ast.Call.args
-                    arg_node = node.args[index - 1]
+                    index -= 1
+                # If the final argument is a starred object, then we treat any args
+                # past the end as pointing to the starred object (this would be a variadic call,
+                # and the starred object would be a catchall)
+                if index >= len(node.args) and isinstance(node.args[-1], ast.Starred):
+                    arg_node = node.args[-1]
```
slyubomirsky, line 133: This is a case that simply didn't come up before. Here, it results from `slice_helper(input_tensor, *args)` in the slice cases. I am not exactly sure why it did not come up with the previous approach, since the conditions should have been met there too. Perhaps the source code is being fetched when it should not have?

slyubomirsky: Ah, this has to do with `STACK_DEPTH_OF_BUILD` in the tensor constructor. We may have to adjust it because the maximum number of stack frames is, I think, more than 4 now. For one, the tensor constructor now includes a call to `raw_init`, so it should be at least 5. Are there any other stack frames we should consider? I don't think these changes should add any since they just replace existing uses of the conversion decorator.

tripy/docs/post0_developer_guides/how-to-add-new-ops.md
```diff
@@ -192,9 +191,9 @@ from tripy.types import ShapeLike
 # `autodoc_options` parameter.
 @export.public_api(document_under="tensor_operations")
 
-# The `convert_to_tensors` decorator automatically converts compatible
-# arguments, like `TensorLike` or `ShapeLike`s, into tensors.
-@frontend_utils.convert_to_tensors()
+# We can use the `constraints.dtypes` decorator to automatically convert
+# compatible arguments, like `TensorLike` or `ShapeLike`s, into tensors.
+@constraints.dtypes(convert_tensor_and_shape_likes=True)
```
pranavm-nvidia, line 196: Can we rename/move the decorator to more accurately reflect its functionality and also to convey that we should not add more decorators if we can instead add functionality to this one? Maybe something like `@interface`?

tripy/tripy/constraints.py
```diff
@@ -58,18 +124,104 @@ def get_arg_dtype(arg, func_name, arg_name) -> utils.Result["tripy.dtype"]:
     return utils.Result.ok(arg_dtype)
 
 
+def convert_input_types(func, args, kwargs, conversion_targets, conversion_preprocess_func, constraints, shape_likes):
+    from tripy.common.datatype import bool as tp_bool
+    from tripy.common.datatype import floating, integer
+    from tripy.frontend.dimension_size import DimensionSize
+    from tripy.frontend.tensor import Tensor
+    from tripy.frontend.utils import tensor_from_shape_like
```
pranavm-nvidia, line 132: Seems like this helper is only used in this file. Can we move it here?

tripy/tripy/constraints.py
```diff
@@ -58,18 +124,104 @@ def get_arg_dtype(arg, func_name, arg_name) -> utils.Result["tripy.dtype"]:
...
+    return new_args, new_kwargs
+
+
 def dtypes(
     constraints: Dict[str, str] = {},
     variables: Dict[str, List[str]] = {},
     exceptions: List[Dict[str, str]] = [],
     aliases: List[str] = [],
+    convert_tensor_and_shape_likes: bool = False,
+    conversion_targets: Set[str] = None,
```
pranavm-nvidia, line 219: Does it make sense to unify these? We can have:
```py
convert_to_tensors: Union[bool, Set[str]]
```
where `convert_to_tensors=True` means that we should automatically choose what to convert while `convert_to_tensors=[x, y, z]` means we should convert only `x`, `y`, and `z`.

tripy/tripy/constraints.py
```diff
@@ -78,29 +230,68 @@ def dtypes(
...
+
+        # if no dtype constraints have been specified at all, do not add to the table so we don't generate invalid tests
+        if constraints or variables or exceptions:
+            for key in [func.__qualname__] + aliases:
+                TYPE_VERIFICATION[key] = verif_info
 
         @functools.wraps(func)
         def wrapper(*args, **kwargs):
+            if convert_tensor_and_shape_likes:
+                args, kwargs = convert_input_types(
```
pranavm-nvidia, line 291: Would be good if there's a way we can avoid doing `utils.merge_function_arguments` twice. Maybe we can call it here (conditionally, so we don't incur the cost if tensor conversion/dtype checking is disabled) and pass it as an input to `convert_input_types`? Then if `merged_args` is already set, we don't need to call it again in the next block.

tripy/tripy/utils/ast.py
pranavm-nvidia: Could you add a test for `STACK_DEPTH_OF_BUILD`? Its purpose is to exclude frames for helpers like `build_internal`, `raw_init`, etc. such that the first frame we see for `Tensor`s is the API (e.g. `full`) where `Op.build` was called.

pranavm-nvidia: Would also be good to add a unit test for `get_arg_candidate_column_offsets`. I think all the logic in this file is very fragile and not well tested.

tripy/tripy/constraints.py
slyubomirsky: The main impediment is that we want to update the args during the type conversion and this would require also updating the merged args data structure. I will see if there will be an easy way to keep the merged args up to date.

pranavm-nvidia: Sure, if it's non-trivial it's fine to keep it as-is.

tripy/tripy/utils/ast.py
slyubomirsky: Perhaps our behavior was wrong before. 4 does appear to be the correct number from the cases I've looked at, so we probably should have run into that case earlier.

tripy/tripy/constraints.py
```diff
@@ -58,69 +124,214 @@ def get_arg_dtype(arg, func_name, arg_name) -> utils.Result["tripy.dtype"]:
...
+                arg.stack_info = original_stack_info
+
+        add_arg(arg)
+
+    return new_args, new_kwargs, new_merged_args
+
+
+def interface(
+    dtype_constraints: Dict[str, str] = {},
     variables: Dict[str, List[str]] = {},
```
pranavm-nvidia, line 227: We should probably also add the `dtype_` prefix to `variables`.

tripy/docs/post0_developer_guides/how-to-add-new-ops.md
```diff
@@ -192,9 +191,12 @@ from tripy.types import ShapeLike
 # `autodoc_options` parameter.
 @export.public_api(document_under="tensor_operations")
 
-# The `convert_to_tensors` decorator automatically converts compatible
-# arguments, like `TensorLike` or `ShapeLike`s, into tensors.
-@frontend_utils.convert_to_tensors()
+# We can use the `constraints.interface` decorator to specify constraints on
```
pranavm-nvidia, line 194: Might want to rename `constraints` -> `wrappers` or something generic like that. `constraints` doesn't really capture the scope of the decorator.

tripy/docs/post0_developer_guides/how-to-add-new-ops.md
```diff
@@ -192,9 +191,12 @@ from tripy.types import ShapeLike
...
+# inputs and perform transformations on them, like automatically converting
+# compatible arguments (e.g., `TensorLike` or `ShapeLike`s) into tensors.
+# We will aim to include most constraints and transformations in this decorator
+# so as to avoid layering too many decorators.
+@constraints.interface(convert_tensor_and_shape_likes=True)
```
pranavm-nvidia, line 199: I think `convert_to_tensors` is a more accurate reflection of what we're doing. The types that we support (i.e. `TensorLike` and `ShapeLike`) are just a detail that might change over time. OTOH the behavior (i.e. converting things to `Tensor`s) should remain consistent.

tripy/tests/frontend/trace/ops/test_slice.py
```diff
@@ -29,6 +29,14 @@ def test_op_func_all_partial(self):
         assert isinstance(a, tp.Tensor)
         assert isinstance(a.trace_tensor.producer, Slice)
 
+    def test_slice_of_inline_output(self):
+        a = tp.Tensor([1, 2, 3, 4])
+        # The start and stop params use clamp bound, but the step parameter doesn't.
+        # The result is that the stack traces for the slice params are of different lengths.
```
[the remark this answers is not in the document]
slyubomirsky: I was actually getting an out of bounds failure for a case that looks like this in the Python REPL but I can't reproduce it in a unit test. However, in this case, the stack trace for the step parameter is shorter than that for the start and stop, so it does at least exercise the logic for checking the frame index in there. I haven't been able to find a test case that causes the out of bounds failure outside of the REPL, unfortunately.

tripy/docs/post0_developer_guides/how-to-add-new-ops.md
slyubomirsky: That is true, only part of that file deals with dtype constraints now.

tripy/tests/frontend/trace/ops/test_slice.py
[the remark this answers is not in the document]
slyubomirsky: This is not the only time I've noticed stack trace oddities when using the REPL vs writing unit tests, for the record.

pranavm-nvidia (approved):

PR #421, opened 2024-12-03
Fix get_stack_info for python3.12

parthchadha:


tripy/tripy/utils/stack_info.py
```diff
@@ -119,14 +119,17 @@ def get_stack_info(include_code_index: int = None) -> StackInfo:
             source_info._dispatch_target = frame.f_locals["key"]
 
         try:
-            # In Python 3.11, frames contain column offset information.
-            frame.f_code.co_positions
-        except AttributeError:
+            if not hasattr(frame.f_code, "co_positions"):
+                raise AttributeError
```
pranavm-nvidia, line 123: I think we can just use `getattr` or `frame.f_code.co_positions` here. It's effectively the same as the condition.

tripy/tripy/utils/stack_info.py
```diff
@@ -119,14 +119,17 @@ def get_stack_info(include_code_index: int = None) -> StackInfo:
...
+
+            # For 3.11+, positions is an iterator of (line, end_line, column, end_column)
+            for pos in frame.f_code.co_positions():
+                if pos[0] == frame.f_lineno:
+                    _, _, start, end = pos
+                    source_info.column_range = (start, end)
+                    break
+        except (AttributeError, ValueError):
```
pranavm-nvidia, line 131: It doesn't seem like anything in the code above could cause a `ValueError`. Can we keep the implementation in the `else` block like we had it before so that we don't hide errors?

pranavm-nvidia (approved):

PR #422, opened 2024-12-03
Allows `DimensionSize`s to be used in `InputInfo`

pranavm-nvidia:
Introduces a new `IntLike` type alias that covers integers and `DimensionSize`s and updates `InputInfo` to accept shape arguments of this new type. Note that the `DimensionSize`s must be evaluated in order to populate the profile correctly.

This allows for using the shapes of eager mode tensors when compiling.

### DIFF
--- a/tripy/tripy/types.py
+++ b/tripy/tripy/types.py
@@ -33,15 +33,26 @@ TensorLike = export.public_api(
     module=sys.modules[__name__],
     symbol="TensorLike",
     doc="""
-        Type annotation for a parameter that is either a Tripy :class:`Tensor` or a Python number that can be automatically converted into one.
+        A Tripy :class:`Tensor` or a Python number that can be automatically converted into one.
         """,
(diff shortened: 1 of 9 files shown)


tripy/tripy/types.py
```diff
@@ -33,15 +33,26 @@
     module=sys.modules[__name__],
     symbol="TensorLike",
     doc="""
-        Type annotation for a parameter that is either a Tripy :class:`Tensor` or a Python number that can be automatically converted into one.
+        A Tripy :class:`Tensor` or a Python number that can be automatically converted into one.
         """,
 )(Union["tripy.Tensor", numbers.Number])
 
+
+IntLike = export.public_api(
```
parthchadha, line 41: When we allow FP shape tensors we will have to change the alias name again. Maybe use `NumberLike`?

slyubomirsky (approved):

parthchadha (approved):

PR #426, opened 2024-12-04
Adds ExecutorOptions and DeviceOptions options providers

pranavm-nvidia:
- Moves `executorIndexBitwidth` and `executorUsePackedMemRefCConv` into its
    own options bundle.

- Adds a `DeviceOptions` provider and updates the OptionsContext and OptionsProviders
    to use `llvm::Error`s instead of `mlirtrt::Status` since the latter is not accessible
    to the OptionsContext.


mlir-tensorrt/compiler/lib/Compiler/OptionsProviders.cpp
```diff
@@ -0,0 +1,64 @@
...
+//===----------------------------------------------------------------------===//
+///
+/// Data structures and functions for manipulating compiler options.
+///
+//===----------------------------------------------------------------------===//
+#include "mlir-tensorrt/Compiler/OptionsProviders.h"
+#include "cuda_runtime_api.h"
+#include "llvm/Support/Error.h"
+
+// TODO (pranavm): Check if we can just reuse `DeviceInfo.cpp`?
```
pranavm-nvidia, line 28: This logic looks almost the same as what's in `DeviceInfo.cpp`. If we can share code between them that might be good.

mlir-tensorrt/compiler/lib/Compiler/OptionsProviders.cpp
```diff
@@ -0,0 +1,64 @@
...
+  int64_t smVersion = ccMajor * 10 + ccMinor;
+  this->computeCapability = smVersion;
+  this->maxSharedMemoryPerBlockKb = properties.sharedMemPerBlock / 1024;
+  this->maxRegistersPerBlock = properties.regsPerBlock;
+  return llvm::Error::success();
+}
+
+llvm::Error mlirtrt::compiler::DeviceOptions::finalize() {
+  if (shouldInferFromHost) {
+    // TODO (pranavm): How to check whether options were provided?
```
pranavm-nvidia, line 58: Not sure how to validate the options here. I don't know if we have access to the actual CLI options

mlir-tensorrt/compiler/lib/Compiler/OptionsProviders.cpp
```diff
@@ -0,0 +1,64 @@
...
+    // Does llvm::cl have a notion of mutually exclusive options like Python's
+    // argparse?
+    return inferDeviceOptionsFromHost();
```
pranavm-nvidia, line 61: Need to figure out how to raise an error if conflicting device options were provided.

mlir-tensorrt/compiler/lib/Compiler/StableHloToExecutable.cpp
```diff
@@ -482,6 +436,7 @@ StableHloToExecutableTask::compileStableHLOToExecutable(
 //===----------------------------------------------------------------------===//
 
 namespace {
+// TODO (pranavm): Any way to reuse `DeviceOptions` here?
```
pranavm-nvidia, line 439: Not quite sure how to deduplicate `ClusteringPipelineCliOpts` and `DeviceOptions`

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsProviders.h
```diff
@@ -27,13 +27,21 @@
...
+  virtual ~OptionsProvider() {}
+  virtual void addToOptions(mlir::OptionsContext &context) = 0;
+
+  virtual llvm::Error finalize() { return llvm::Error::success(); }
+};
+
 /// DebugOptions are options that are common to different compiler API
 /// interfaces.
-struct DebugOptions {
+struct DebugOptions : public OptionsProvider {
```
pranavm-nvidia, line 44: I can use CRTP here if we don't want vtable overhead

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsProviders.h
```diff
@@ -57,19 +65,57 @@ struct DebugOptions {
...
+    context.addOption("device-max-registers-per-block", maxRegistersPerBlock,
+                      llvm::cl::init(0));
+    context.addOption("device-infer-from-host", shouldInferFromHost,
+                      llvm::cl::init(true),
+                      llvm::cl::desc("Infers device information from host"));
+  }
+
+  llvm::Error finalize() override;
+
+  // TODO (pranavm): This should ideally be private but needs to be set from
```
pranavm-nvidia, line 107: Need to figure out if I can make this option private if we are able to unify `DeviceOptions` and `ClusteringPipelineCliOpts`.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsProviders.h
christopherbate: If we didn't need a base class to require `DebugOptions` to have `addToOptions`, then why did we need one to add `finalize`?

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsProviders.h
```diff
@@ -0,0 +1,121 @@
...
+
+  void addToOptions(mlir::OptionsContext &context) override {
+    context.addOption(
+        "device-compute-capability", computeCapability, llvm::cl::init(64),
+        llvm::cl::desc("Sets the device compute capbility. Only relevant "
+                       "if '--device-infer-from-host=false'"));
+    context.addOption("device-max-shared-memory-per-block-kb",
+                      maxSharedMemoryPerBlockKb, llvm::cl::init(0));
+    context.addOption("device-max-registers-per-block", maxRegistersPerBlock,
+                      llvm::cl::init(0));
```
christopherbate, line 99: Set default value to 65536 (limit for all arch since sm60).

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsProviders.h
```diff
@@ -0,0 +1,121 @@
...
+
+  /// Target device maximum 4-byte register sper block.
+  uint64_t maxRegistersPerBlock;
+
+  void addToOptions(mlir::OptionsContext &context) override {
+    context.addOption(
+        "device-compute-capability", computeCapability, llvm::cl::init(64),
+        llvm::cl::desc("Sets the device compute capbility. Only relevant "
+                       "if '--device-infer-from-host=false'"));
+    context.addOption("device-max-shared-memory-per-block-kb",
```
christopherbate, line 96: Let's set default value to `48` (which was value for sm60, all other arch have equal or higher limit since then)

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsRegistry.h
```diff
@@ -71,14 +72,15 @@ optionsCreateFromArgs(const CompilerClient &client,
...
-    return inferStatus;
+  llvm::Error finalizeStatus = result->finalize();
+
+  std::optional<std::string> errMsg{};
+  llvm::handleAllErrors(
+      std::move(finalizeStatus),
+      [&errMsg](const llvm::StringError &err) { errMsg = err.getMessage(); });
+
+  if (errMsg)
+    return getInternalErrorStatus(errMsg->c_str());
```
christopherbate, line 83: Prefer to add a bit more context. I think you can use printf-style formatting with this function:

```
return getInternalErrorStatus("failed to initialize options: %s", errMsg->c_str());
```

mlir-tensorrt/compiler/lib/Compiler/OptionsProviders.cpp
christopherbate: I gave this some thought and want to rework the logic a bit (this isn't on your plate or a requirement for this PR). 

This setup essentially is only allowing two mutually exclusive paths (either infer all values from CUDA runtime API or manually specify the values). I think in the future we will want something more flexible, e.g. "infer all values" or "infer all values but override others" or "populate all values with defaults for sm_XY", etc. 

Anyway, it will require a bit more thought so we can leave as future work.

mlir-tensorrt/compiler/lib/Compiler/StableHloToExecutable.cpp
```diff
@@ -299,16 +251,18 @@ void StableHloToExecutableTask::buildPostClusteringPipeline(
   populateExtensionPasses(pm, opts, Phase::ExecutorLowering);
 
   ConvertCUDAToExecutorPassOptions cudaToExecutorOpts;
-  cudaToExecutorOpts.indexBitwidth = opts.executorIndexBitwidth;
-  cudaToExecutorOpts.usePackedMemRefCConv = opts.executorUsePackedMemRefCConv;
+  cudaToExecutorOpts.indexBitwidth =
+      opts.get<ExecutorOptions>().executorIndexBitwidth;
```
christopherbate, line 255: Nit: we can probably update the names to leave off the `executor` prefix since it is now obvious

mlir-tensorrt/compiler/lib/Compiler/StableHloToExecutable.cpp
christopherbate: A million dollar question :joy:. In all seriousness, I don't think you can right now. We're butting up against limitations of C++ (no builtin reflection) and the design of the upstream `PassPipelineOptions` (no inheritance scheme/limited flexibility). You would need to make the upstream `PassPipelineOptions` more flexible at least.

pranavm-nvidia:
> Can you rebase and move off draft form?

@christopherbate Yes, but I'm still waiting for #419 to merge since this is dependent on those changes.

mlir-tensorrt/compiler/lib/Compiler/StableHloToExecutable.cpp
pranavm-nvidia: Ah yes, I did that with device options but forgot to do so for the executor. Will fix.

mlir-tensorrt/test/python/mlir_tensorrt_compiler/compiler_api/test_options_context.py
```diff
@@ -27,4 +27,4 @@
...
-# CHECK: --tensorrt-timing-cache-path= --device-infer-from-host=true --debug-only= --executor-index-bitwidth=64 --entrypoint=main --plan-clustering-disallow-host-tensors-in-tensorrt-clusters=false --tensorrt-workspace-memory-pool-limit=1073741824 --device-max-registers-per-block=65536 --tensorrt-strongly-typed=false --tensorrt-layer-info-dir= --device-compute-capability=86 --debug=false --mlir-print-ir-tree-dir= --disable-tensorrt-extension=false --tensorrt-builder-opt-level=3 --tensorrt-engines-dir=
+# CHECK: Options[--tensorrt-timing-cache-path= --device-infer-from-host=true --debug-only= --executor-index-bitwidth=64 --entrypoint=main --device-max-registers-per-block=65536 --tensorrt-workspace-memory-pool-limit=1073741824 --plan-clustering-disallow-host-tensors-in-tensorrt-clusters=false --tensorrt-strongly-typed=false --device-compute-capability=86 --tensorrt-layer-info-dir= --debug=false --mlir-print-ir-tree-dir= --disable-tensorrt-extension=false --tensorrt-engines-dir= --tensorrt-builder-opt-level=3 --device-max-shared-memory-per-block-kb=48]
```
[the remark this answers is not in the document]
pranavm-nvidia: But we do want at least a sanity check that the options are being printed correctly right? Maybe I can limit it to just a few options that we can explicitly set?

mlir-tensorrt/tensorrt/include/mlir-tensorrt-dialect/Utils/Options.h
```diff
@@ -155,6 +155,9 @@ class OptionsContext : public llvm::cl::SubCommand {
   /// as a cache key.
   virtual std::optional<llvm::hash_code> getHash() const;
 
+  /// Post-process options after parsing.
+  virtual llvm::Error finalize() = 0;
```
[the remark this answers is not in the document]
pranavm-nvidia: Yes, I'll elaborate on this before I open up the PR for review.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsProviders.h
pranavm-nvidia: @christopherbate It's so we can have a default implementation since most of the classes don't actually need to implement it. We could alternatively use some template magic to not call `finalize` at all if it's not defined:
```cpp
if constexpr (hasFinalize<OptionsProviderT>())
    result->finalize();
```
That's not too difficult to do with SFINAE (and obviously we'd factor it out so it doesn't pollute the call-sites).

mlir-tensorrt/compiler/lib/Compiler/OptionsProviders.cpp
pranavm-nvidia: How about just making the default value an empty optional? Then for any argument that isn't provided, we can overwrite with the inferred value. That wouldn't be difficult to do I think and I could include it in this PR.

mlir-tensorrt/compiler/lib/Compiler/StableHloToExecutable.cpp
pranavm-nvidia: Ah that's unfortunate. Will leave it as-is for now.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsProviders.h
```diff
@@ -0,0 +1,121 @@
...
+
+  /// Target device max shared memory per block (kilobytes)
+  int64_t maxSharedMemoryPerBlockKb;
+
+  /// Target device maximum 4-byte register sper block.
+  uint64_t maxRegistersPerBlock;
+
+  void addToOptions(mlir::OptionsContext &context) override {
+    context.addOption(
+        "device-compute-capability", computeCapability, llvm::cl::init(64),
```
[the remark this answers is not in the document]
pranavm-nvidia: Fixed

mlir-tensorrt/compiler/lib/Compiler/OptionsProviders.cpp
```diff
@@ -0,0 +1,64 @@
...
+//===----------------------------------------------------------------------===//
+#include "mlir-tensorrt/Compiler/OptionsProviders.h"
+#include "cuda_runtime_api.h"
+#include "llvm/Support/Error.h"
+
+// TODO (pranavm): Check if we can just reuse `DeviceInfo.cpp`?
+llvm::Error mlirtrt::compiler::DeviceOptions::inferDeviceOptionsFromHost() {
+  cudaDeviceProp properties;
+  cudaError_t err = cudaGetDeviceProperties(&properties, 0);
+  if (err != cudaSuccess)
```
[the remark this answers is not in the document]
pranavm-nvidia: Done

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsProviders.h
pranavm-nvidia: Updated.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsProviders.h
pranavm-nvidia: Updated.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsRegistry.h
pranavm-nvidia: Done

mlir-tensorrt/compiler/lib/Compiler/OptionsProviders.cpp
pranavm-nvidia: @christopherbate I guess the question here is whether the compiler is allowed to depend on the executor?

christopherbate: Yes, compiler depends on the `executor` project

christopherbate: Declaring dependency on `MLIRTensorRTSupportDeviceInfo` is fine

mlir-tensorrt/compiler/lib/Compiler/OptionsProviders.cpp
christopherbate: Let's push this to a later point in time, I'd rather have something sketched out first rather doing adhoc. It has to work with all parts of the project, including internal.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsProviders.h
christopherbate: That sounds like a good idea

mlir-tensorrt/test/python/mlir_tensorrt_compiler/compiler_api/test_options_context.py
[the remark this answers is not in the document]
christopherbate: ok, yeah, I might have been too heavy handed there. I'd say just checking 1-2 items is best, otherwise you'll need to update it every time we update the options

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsProviders.h
pranavm-nvidia: Added

mlir-tensorrt/executor/lib/Support/DeviceInfo.cpp
```diff
@@ -23,26 +23,33 @@
 
 using namespace mlirtrt;
 
+namespace {
```
christopherbate, line 26: nit: We use the LLVM convention (https://llvm.org/docs/CodingStandards.html#anonymous-namespaces), which is to declare functions `static` rather than wrapping in anonymous namespace. Only local classes/structs would get wrapped in anonymous namespace.

christopherbate (approved):

mlir-tensorrt/executor/lib/Support/DeviceInfo.cpp
pranavm-nvidia: Oh I see, wasn't aware. Fixed now.

PR #429, opened 2024-12-06
Fixes type checking, nanoGPT

pranavm-nvidia:
[Adds support for various types in our type checking logic](https://github.com/NVIDIA/TensorRT-Incubator/commit/3c95dafa9e4c21c8fab6ab7e2e83ecb98c664b83)

Adds support for dictionaries, tuple, and callables in our type checking.
Also removes a fallback path where we would skip type checking if we weren't
able to handle an annotation. Since we're only doing type checking on Tripy APIs,
where we control the type annotations, we should not have this fallback path but
should instead add support for whatever types we use.

Also updates various parts of the code to respect type annotations. Many of these
were working before since the type annotation would just be ignored.

[Updates pipelines to remove mentions of the "manual" test cadence](https://github.com/NVIDIA/TensorRT-Incubator/commit/f9d2d9bd59e14fe239e7a11e5c4c3923af3487b3)

[Updates nanoGPT to not store certain tensors as members](https://github.com/NVIDIA/TensorRT-Incubator/commit/a92c2fa523c0d142cb4bb1e3f9ebd88bdfa87fac)

### DIFF
--- a/tripy/examples/nanogpt/model.py
+++ b/tripy/examples/nanogpt/model.py
@@ -67,7 +67,6 @@ class CausalSelfAttention(tp.Module):
             tp.tril(tp.ones((config.block_size, config.block_size), dtype=config.dtype)),
             (1, 1, config.block_size, config.block_size),
         )
-        self.zeros = tp.zeros((1, 1, self.seq_len, self.seq_len), dtype=config.dtype)
 
     def __call__(self, x: tp.Tensor):
         B, T = x.shape[0:2]
@@ -87,11 +86,7 @@ class CausalSelfAttention(tp.Module):
 
         k_t = tp.transpose(k, -2, -1)
         att = (q @ k_t) * (1.0 / math.sqrt(self.embedding_size // self.num_heads))
-        att = tp.masked_fill(
-            att,
-            self.bias[:, :, :T, :T] == self.zeros[:, :, :T, :T],
-            float("-inf"),
-        )
+        att = tp.masked_fill(att, self.bias[:, :, :T, :T] == 0, float("-inf"))
 
         att = tp.softmax(att, dim=-1)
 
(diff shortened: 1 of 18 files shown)


tripy/tripy/backend/api/input_info.py
```diff
@@ -63,25 +63,8 @@ def __init__(self, shape: Sequence[Union[IntLike, Tuple[IntLike, IntLike, IntLik
...
-                            f" but got: {len(elem)} values in shape: {shape}. "
-                        ],
-                    )
-            else:
-                raise_error(
-                    "Shape values should be either a single integer-like value or a 3-element tuple specifying min/opt/max bounds.",
-                    [f"Shape: {shape} contains an invalid element: {elem}"],
-                )
+
+            assert len(elem) == 3 and all(is_int_like(val) for val in elem)
```
pranavm-nvidia, line 67: This is now taken care of by type checking, hence why it's just an assertion now.

tripy/tripy/frontend/module/sequential.py
```diff
@@ -12,30 +12,31 @@
...
-from dataclasses import dataclass
 import copy
+from dataclasses import dataclass
+from typing import Any, Callable, Dict, Iterator, Tuple, Union
 
 from tripy import export
 from tripy.common.exception import raise_error
 from tripy.frontend.module import Module
 
+ModuleLike = Union[Module, Callable[["tripy.Tensor"], "tripy.Tensor"]]
```
pranavm-nvidia, line 23: This is just for convenience and is automatically expanded wherever relevant (e.g. the rendered docs)

tripy/tripy/frontend/module/layernorm.py
```diff
@@ -43,7 +43,7 @@ class LayerNorm(Module):
     dtype: datatype.dtype
     r"""The data type used to perform the operation."""
 
-    normalized_shape: Tuple[int]
+    normalized_shape: Sequence[int]
```
pranavm-nvidia, line 46: Note that `Tuple` specifies a fixed number of arguments. For sequences of arbitrary length, we should use `Sequence` instead.

tripy/examples/nanogpt/model.py
```diff
@@ -137,16 +132,15 @@ def __init__(self, config):
...
         self.wpe = tp.Embedding(config.block_size, config.embedding_size, dtype=config.dtype)
-        self.h = [Block(config) for _ in range(config.num_layers)]
+        self.h = tp.Sequential(*[Block(config) for _ in range(config.num_layers)])
         self.ln_f = tp.LayerNorm(config.embedding_size)
-        self.pos = tp.reshape(tp.arange(0, config.seq_len, dtype=tp.int32), (1, config.seq_len))
 
     def __call__(self, idx):
         tok_emb = self.wte(idx)  # token embeddings of shape (batch_size, seq_len, embedding_size)
-        pos_emb = self.wpe(self.pos[:, : idx.shape[1]])  # position embeddings of shape (seq_len, embedding_size)
+        pos = tp.unsqueeze(tp.arange(idx.shape[1], dtype=tp.int32), 0)
```
parthchadha, line 140: I think this is incorrect, `config.seq_len` will be the max seq len and `idx.shape[1]` will be the current sequence length.

pranavm-nvidia: Ah so we need to constrain it to the `config.seq_len`?

pranavm-nvidia: Updated.

parthchadha (approved):

PR #435, opened 2024-12-09
[Compiler/TensorRT] Add tensorrt-to-executable task

pranavm-nvidia:


mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/TensorRTToExecutable.h
```diff
@@ -0,0 +1,85 @@
...
+// TODO (pranavm): Figure out a better way to reuse TRT translation options -
+// maybe move to options providers?
+struct TensorRTOptions
+    : public mlirtrt::compiler::OptionsProvider<TensorRTOptions> {
+  mlir::tensorrt::TensorRTTranslationOptions options;
+
+  void addToOptions(mlir::OptionsContext &context) {
+    options.addToOptions(context);
+  }
+};
```
pranavm-nvidia, line 47: I can move `TensorRTTranslationOptions` to make them an options provider if that makes sense to do.

mlir-tensorrt/compiler/include/mlir-tensorrt/Registration/RegisterMlirTensorRtPasses.h
```diff
@@ -52,6 +53,12 @@ inline void registerAllMlirTensorRtPasses() {
   mlir::registerTransformsPasses();
   mlir::registerConvertPDLToPDLInterp();
 
+  // TODO (pranavm): Check if this needs to be conditional - the TRT passes
+  // above are not.
+#ifdef MLIR_TRT_TARGET_TENSORRT
+  mlirtrt::compiler::registerTensorRTToExecutableTask();
+#endif
```
pranavm-nvidia, line 60: I'm not sure what exactly needs to be guarded. In a lot of places we guard TRT things with `MLIR_TRT_TARGET_TENSORRT` but that doesn't seem to be the case for the TRT passes registered above.

mlir-tensorrt/compiler/lib/Compiler/CMakeLists.txt
```diff
@@ -19,6 +20,8 @@ add_mlir_tensorrt_library(MLIRTensorRTCompilerStableHloToExecutable
     StableHloToExecutable.cpp
     # TODO: TensorRTExtension should be an independent library.
     TensorRTExtension/TensorRTExtension.cpp
+    # TODO (pranavm): TensorRTToExecutable should probably be a separate library
+    TensorRTToExecutable.cpp
```
pranavm-nvidia, line 24: Given that there's an open TODO for `TensorRTExtension.cpp` I'm wondering if this is more complicated than it seems.

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include "mlir-executor/Executor/Transforms/Passes.h"
+#include "mlir-tensorrt/Compiler/OptionsRegistry.h"
+#include "mlir-tensorrt/Compiler/PassManagerUtils.h"
+
+using namespace mlirtrt::compiler;
+
+TensorRTToExecutableOptions::TensorRTToExecutableOptions(
+    TaskExtensionRegistry extensions) {
+  // TODO (pranavm): Do we need to support extensions?
+}
```
pranavm-nvidia, line 33: Not sure if we want to require extensions for all options types or if we need to handle both cases in the options registry. If it's the former, then I can just assert that the extensions are empty here (or maybe even just add support?). If it's the latter, we could have a `setExtensions` method so it becomes optional instead of having it part of the constructor.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/PassManagerUtils.h
```diff
@@ -0,0 +1,32 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//===----------------------------------------------------------------------===//
+
+#include "mlir-tensorrt/Compiler/OptionsProviders.h"
+#include "mlir/Pass/PassManager.h"
+
+using namespace mlirtrt::compiler;
```
christopherbate, line 24: We should never ever put `using namespace ...` into a header file

mlir-tensorrt/compiler/include/mlir-tensorrt/Registration/RegisterMlirTensorRtPasses.h
christopherbate: The TensorRT dialect and MLIR passes are always built. It doesn't depend on actually having TensorRT binaries or headers to build against (or at least, that was the idea, we haven't been enforcing it). The `MLIR_TRT_TARGET_TENSORRT` is basically a guard for anything  that relies on having actual TRT headers or libraries to link against, for example the translation from MLIR to TensorRT, hence the `TARGET_TENSORRT` in the name.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/TensorRTToExecutable/Passes.td
```diff
@@ -0,0 +1,35 @@
...
+//===----------------------------------------------------------------------===//
+#ifndef MLIR_TENSORRT_COMPILER_TENSORRTTOEXECUTABLE_PASSES
+#define MLIR_TENSORRT_COMPILER_TENSORRTTOEXECUTABLE_PASSES
+
+include "mlir/Pass/PassBase.td"
+
+//===----------------------------------------------------------------------===//
+// OutlineTensorRTOpPass
+//===----------------------------------------------------------------------===//
+// TODO: what are the dependent dialects? what are the options?
```
yizhuoz004, line 28: What are the required dependent dialects/options?

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp
```diff
@@ -0,0 +1,250 @@
...
+  result = tensorrt::TensorRTModuleOp::create(moduleOp.getLoc(), "trt_engines");
+  symbolTable.insert(result, op->getParentOp() == moduleOp ? Block::iterator(op)
+                                                           : Block::iterator{});
+  return result;
+}
+
+/// Helper function to call the `makeRegionIsolatedFromAbove` to capture all
+/// required arguments into the InlineGroupOp region.
+// static LogicalResult
+// makeIsolatedFromAboveImpl(RewriterBase &rewriter, plan::InlineGroupOp regionOp,
```
yizhuoz004, line 120: Is this function still useful?

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp
```diff
@@ -0,0 +1,250 @@
...
+          OutlineTensorRTOpPass> {
+public:
+  using Base::Base;
+  void runOnOperation() override {
+    ModuleOp module = getOperation();
+
+    SymbolTableCollection symbolTable;
+    IRRewriter rewriter(&getContext());
+    // what are these? are they needed?
+    // DataFlowSolver solver;
```
yizhuoz004, line 183: We just decide if an op is tensorrt dialect or not. is `solver` still needed then?

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp
```diff
@@ -0,0 +1,250 @@
...
+    // solver.load<TensorKindAnalysis>(symbolTable);
+    // if (failed(solver.initializeAndRun(module)))
+    //   return signalPassFailure();
+
+    FailureOr<ClusteringOpts> opts = getTensorRTClusteringOptions(module);
+    if (failed(opts)) {
+      emitError(module.getLoc()) << "failed to create clustering options";
+      return signalPassFailure();
+    }
+    // What do they do here?
```
yizhuoz004, line 195: Did not figure out this part, `regionOps` is not used either.

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/TensorRTToExecutable/Passes.td
yizhuoz004: Also I got error during compilation:
```
error: ‘createOutlineTensorRTOpPass’ was not declared in this scope
```
Did I missing anything for the tablegen config?

yizhuoz004: Fixed, there's one missing header in `TensorRTToExecutable.cpp`

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/TensorRTToExecutable/TensorRTToExecutable.h
```diff
@@ -0,0 +1,107 @@
...
+// TODO (pranavm): Figure out a better way to reuse TRT translation options -
+// maybe move to options providers?
+struct TensorRTOptions : public OptionsProvider<TensorRTOptions> {
+public:
+  using OptionsProvider::OptionsProvider;
+  mlir::tensorrt::TensorRTTranslationOptions options;
+
+  TensorRTOptions(mlir::OptionsContext &ctx) : OptionsProvider(ctx) {}
+
+  void addToOptions(mlir::OptionsContext &context) {
```
yizhuoz004, line 57: Where should I call this function? Currently I got `no option registered with name tensorrt-builder-opt-level` when testing the bindings.

yizhuoz004: Also fixed by calling it in the constructor.

mlir-tensorrt/python/bindings/Compiler/CompilerPyBind.cpp
```diff
@@ -72,6 +72,19 @@ class PyStableHLOToExecutableOptions
           mtrtStableHloToExecutableOptionsDestroy};
 };
 
+/// Python object type wrapper for `MTRT_TensorRTToExecutableOptions`.
```
pranavm-nvidia, line 75: @yizhuoz004 I don't think we need Python bindings for specific options types anymore. We had added a generic `OptionsContext` API that can be used to construct any options type. See `mlir-tensorrt/compiler/test/python/mlir_tensorrt_compiler/compiler_api/test_options_context.py` for an example.

yizhuoz004: I see, I'll make the changes, thanks

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/OptionsRegistry.h
```diff
@@ -78,7 +78,7 @@ optionsCreateFromArgs(mlir::MLIRContext *context,
       [&errMsg](const llvm::StringError &err) { errMsg = err.getMessage(); });
 
   if (errMsg)
-    return getInternalErrorStatus("failed to initialize options: %s",
+    return getInternalErrorStatus("failed to initialize options: {0}",
                                   errMsg->c_str());
```
christopherbate, line 82: Thanks, `errMsg` can also change to just `*errMsg`

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/TensorRTToExecutable/Passes.h
```diff
@@ -0,0 +1,49 @@
+//===- Passes.h ----------------------------------------------===//
+//
+// SPDX-FileCopyrightText: Copyright 2024 NVIDIA CORPORATION & AFFILIATES.
```
christopherbate, line 3: 2025

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/TensorRTToExecutable/Passes.td
```diff
@@ -0,0 +1,34 @@
+//===- Passes.td ----------------------------------------------------------===//
+//
+// SPDX-FileCopyrightText: Copyright 2024 NVIDIA CORPORATION & AFFILIATES.
```
christopherbate, line 3: 2025 here and in other new files

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/TensorRTToExecutable/TensorRTToExecutable.h
```diff
@@ -0,0 +1,105 @@
...
+#include "mlir-tensorrt-dialect/Target/TranslateToTensorRT.h"
+
+#include "mlir-executor/Runtime/API/API.h"
+#include "mlir-executor/Support/Status.h"
+#include "mlir-tensorrt-dialect/Utils/Options.h"
+#include "mlir-tensorrt-dialect/Utils/OptionsBundle.h"
+#include "mlir-tensorrt/Compiler/Client.h"
+#include "mlir-tensorrt/Compiler/Extension.h"
+#include "mlir-tensorrt/Compiler/OptionsProviders.h"
+#include "mlir/IR/BuiltinOps.h"
```
christopherbate, line 36: Several of these headers can be elided 
https://llvm.org/docs/CodingStandards.html#include-as-little-as-possible

mlir-tensorrt/compiler/include/mlir-tensorrt/Registration/RegisterMlirTensorRtPasses.h
```diff
@@ -53,6 +54,12 @@ inline void registerAllMlirTensorRtPasses() {
   mlir::registerTransformsPasses();
   mlir::registerConvertPDLToPDLInterp();
 
+  // TODO (pranavm): Check if this needs to be conditional - the TRT passes
+  // above are not.
```
christopherbate, line 58: This should not go here. Check where the callsites are for `registerStableHlotoExecutable` task

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+    Passes.cpp
+
+    PARTIAL_SOURCES_INTENDED
+
+    DEPENDS
+    MLIRTensorRTTensorRTToExecutableIncGen
+
+    LINK_LIBS PUBLIC
+    MLIRIR
+    MLIRTensorRTRegistration
```
christopherbate, line 12: This catch-all library has been removed in a commit which hasn't been sync'd up yet. Just remove this and add in dependent libraries explicitly.

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+  PassPipelineRegistration<TensorRTToExecutablePassPipelineOptions>(
+      "tensorrt-compilation-pipeline", "apply compilation post-clustering",
+      [](OpPassManager &pm,
+         const TensorRTToExecutablePassPipelineOptions &opts) {
+        TensorRTToExecutableTask::buildPostClusteringPipeline(pm, opts);
+      });
+}
+
+#endif // MLIR_TRT_ENABLE_HLO
```
christopherbate, line 225: Not dependent on ENABLE_HLO

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/TensorRTToExecutable.cpp
```diff
@@ -0,0 +1,191 @@
...
+  pm.addPass(createDropNestedModulesPass());
+}
+
+void TensorRTToExecutableTask::populatePassManager(
+    mlir::PassManager &pm, const TensorRTToExecutableOptions &options) {
+  buildTensorRTClusteringPipeline(pm, options);
+
+  buildPostClusteringPipeline(pm, options);
+
+  mlir::executor::ConvertStdToExecutorPassOptions stdToExecOpts;
```
christopherbate, line 131: You will want to add a `host-target` flag which can take options `executor`, `llvm` or `emitc`. This executor lowering pass is just for the `executor` option. The pipeliens for the other branches will be sync'd up Friday

christopherbate (changes requested):

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp
```diff
@@ -0,0 +1,225 @@
...
+/// clustered into one TensorRT function (which is eventually translated to a
+/// engine).
+static FailureOr<ClusteringOpts>
+getTensorRTClusteringOptions(Operation *op) {
+  ClusteringOpts opts;
+  opts.mergeIndependentClusters = [](Operation *, ClusterRange, Operation *,
+                                     ClusterRange) { return true; };
+  opts.clusterTarget = Attribute{};
+  opts.isClusterableOp = [](Operation *op) {
+    if (llvm::isa<tensorrt::TensorRTDialect>(op->getDialect()))
```
christopherbate, line 55: This can be simplified to just `llvm::isa_and_present<tensorrt::TensorRTDialect>(op->getDialect());`

Note that in certain cases in artificially constructed tests, it's possible for `op->getDialect()` to return null, so you should use the `_and_present` otherwise plain `isa` will assert.

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp
christopherbate: Here you need to run 

```
FailureOr<SmallVector<Cluster>> clusters = 
mlir::analyzeAndClusterOperations(module, opts);
```

In the loop below, just loop over these clusters and pass them to the `outline` function (instead of `InlineGroupOp`

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp
```diff
@@ -0,0 +1,225 @@
...
+  // Create the function. Symbol name de-duplication occurs with insert into the
+  // symbol table.
+  result = tensorrt::TensorRTModuleOp::create(moduleOp.getLoc(), "trt_engines");
+  symbolTable.insert(result, op->getParentOp() == moduleOp ? Block::iterator(op)
+                                                           : Block::iterator{});
+  return result;
+}
+
+static FailureOr<tensorrt::CallAllocOp>
+outlineOp(RewriterBase &rewriter, tensorrt::TensorRTModuleOp trtModule, plan::InlineGroupOp op) {
```
christopherbate, line 118: Here's what I recommend here to make this work:

Change the last `op` parameter to `const Cluster &cluster`, then at the start call 

```
auto inlineGroupOp = cast<plan::InlineGroupOp>(mlir::createRegionOpFromCluster(
      cluster, rewriter,
      [](OpBuilder &b, Location loc, TypeRange types, Attribute target) {
        auto regionOp = b.create<plan::InlineGroupOp>(loc, types, target);
        b.setInsertionPointToStart(&regionOp.getRegion().emplaceBlock());
        b.create<plan::YieldOp>(loc);
        return regionOp;
      }));
```

christopherbate: (Then use `inlineGroupOp` where you have  `op` below except where noted.

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/TensorRTToExecutable.cpp
```diff
@@ -0,0 +1,191 @@
...
+              llvm::iterator_range(options), err);
+
+        llvm::Error finalizeStatus = result.finalize();
+        std::optional<std::string> errMsg{};
+        llvm::handleAllErrors(std::move(finalizeStatus),
+                              [&errMsg](const llvm::StringError &err) {
+                                errMsg = err.getMessage();
+                              });
+
+        if (errMsg)
```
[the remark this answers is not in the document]
christopherbate: (`registerCompilationTaskWithNoExtensions`)

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/TensorRTToExecutable.cpp
```diff
@@ -0,0 +1,191 @@
...
+  buildPostClusteringPipeline(pm, options);
+
+  mlir::executor::ConvertStdToExecutorPassOptions stdToExecOpts;
+  stdToExecOpts.indexBitwidth = options.get<ExecutorOptions>().indexBitwidth;
+  stdToExecOpts.usePackedMemRefCConv = true;
+  mlir::executor::buildExecutorLoweringPipeline(pm, stdToExecOpts);
+}
+
+void mlirtrt::compiler::registerTensorRTToExecutableTask() {
+  registerOption(
```
christopherbate, line 138: `registerOption` no longer needed

mlir-tensorrt/compiler/include/mlir-tensorrt/Registration/RegisterMlirTensorRtPasses.h
yizhuoz004: Fixed, thanks

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp
```diff
@@ -0,0 +1,215 @@
...
+        b.setInsertionPointToStart(&regionOp.getRegion().emplaceBlock());
+        b.create<plan::YieldOp>(loc);
+        return regionOp;
+      }));
+
+  // Make the region isolated from above. This captures the input operands.
+  SmallVector<Value> inputs =
+      makeRegionIsolatedFromAbove(rewriter, inlineGroupOp.getRegion());
+  
+  tensorrt::TensorRTModuleOp trtModule = getOrCreateTensorRTModuleOp(inlineGroupOp);
```
yizhuoz004, line 128: @christopherbate I had to move `getOrCreateTensorRTModuleOp` here to make this pass work, otherwise in `createOutlinedFunc`, the `Symboltable(op)` will hit seg fault. Could you check if this is correct?

An example run, input:
```
func.func @trt_gather_default1(%arg0: tensor<10x20x30xf32>, %arg1: tensor<2x5xi32>,
                %arg2: tensor<10x2x5x30xf32>) -> tensor<10x2x5x30xf32> {
  %0 = tensorrt.gather {
    axis = 1 : i64
  } ins(%arg0, %arg1 : tensor<10x20x30xf32>, tensor<2x5xi32>) -> tensor<10x2x5x30xf32>
  %1 = tensorrt.element_wise <kSUM>(%0, %arg2 : tensor<10x2x5x30xf32>, tensor<10x2x5x30xf32>) -> tensor<10x2x5x30xf32>
  return %1 : tensor<10x2x5x30xf32>
}
```
Output:
```
module {
  func.func @trt_gather_default1(%arg0: tensor<10x20x30xf32>, %arg1: tensor<2x5xi32>, %arg2: tensor<10x2x5x30xf32>) -> tensor<10x2x5x30xf32> {
    %0 = tensorrt.call_alloc @trt_engines::@tensorrt_cluster(%arg0, %arg1, %arg2 : tensor<10x20x30xf32>, tensor<2x5xi32>, tensor<10x2x5x30xf32>) -> tensor<10x2x5x30xf32>
    return %0 : tensor<10x2x5x30xf32>
  }
  tensorrt.module @trt_engines {
    func.func @tensorrt_cluster(%arg0: tensor<10x20x30xf32>, %arg1: tensor<2x5xi32>, %arg2: tensor<10x2x5x30xf32>) -> tensor<10x2x5x30xf32> {
      %0 = tensorrt.gather {axis = 1 : i64} ins(%arg0, %arg1 : tensor<10x20x30xf32>, tensor<2x5xi32>) -> tensor<10x2x5x30xf32>
      %1 = tensorrt.element_wise <kSUM>(%0, %arg2 : tensor<10x2x5x30xf32>, tensor<10x2x5x30xf32>) -> tensor<10x2x5x30xf32>
      return %1 : tensor<10x2x5x30xf32>
    }
  }
}
```

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp
```diff
@@ -0,0 +1,215 @@
...
+  // Tag the function with a UnitAttr for identifying the different kinds of
+  // functions based on the cluster type.
+  return cast<FunctionOpInterface>(outlinedFunc.getOperation());
+}
+
+/// Given the `op`, find the closest ModuleOp and check if the module has a
+/// `tensorrt.module` operation in it. If it does, then return the existing
+/// `tensorrt.module` operation. Otherwise, create a new `tensorrt.module`.
+static tensorrt::TensorRTModuleOp getOrCreateTensorRTModuleOp(Operation *op) {
+  auto moduleOp = op->getParentOfType<ModuleOp>();
```
christopherbate, line 92: You should pass in `ModuleOp moduleOp` to this function instead of trying to `getParent` here. That's probably what caused segfault.

Then down below you can also just change the symbol table insertion to `symbolTable.insert(result)`.

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp
christopherbate: Commented above on how to fix. You should move it out of `outlineOp` back to its original position in order to avoid performing multiple linear scans unnecessarily.

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp
yizhuoz004: Thanks!

mlir-tensorrt/compiler/include/mlir-tensorrt/Compiler/TensorRTToExecutable.h
yizhuoz004: Let's do that in another PR later.

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp
```diff
@@ -0,0 +1,211 @@
...
+namespace {
+
+//===----------------------------------------------------------------------===//
+// OutlineTensorRTOpPass
+//===----------------------------------------------------------------------===//
+
+/// ClusteringOpts that identifies groups of TensorRT operations and will be
+/// clustered into one TensorRT function (which is eventually translated to a
+/// engine).
+static FailureOr<ClusteringOpts> getTensorRTClusteringOptions(Operation *op) {
```
christopherbate, line 46: Static functions don't go in anonymous namespaces, see https://llvm.org/docs/CodingStandards.html#anonymous-namespaces

christopherbate (approved):
There are no tests? You need to add a couple basic tests, probably one from Python API to invoke "tensorrt-to-executable" and 1-2 from `mlir-tensorrt-opt` to invoke the registered passes/pipelines

PR #437, opened 2024-12-11
Move internal changes

shelkesagar29:
This PR moves the following internal commits to OSS,

**Validate any stride for empty tensors**

Empty tensors can have a variety of strides which are all equally
meaningless. The updated validation check allows any stride if the
tensor has a dimension of size zero.

**Exporting APIs for StableHlo PassManager and TRT Executable**

This changes breaks compiler.compiler_stablehlo_to_executable into two separate APIs in such a way:
exe = compiler.compiler_stablehlo_to_executable(client, module.operation, opts)

is changed to

pipeline = compiler.StableHloPipeline(client, opts)
exe = compiler.getExecutable(pipeline, module.operation)


christopherbate (approved):
lgtm

PR #439, opened 2024-12-11
Always construct memref value in storage op

yizhuoz004:


tripy/tripy/common/utils.py
```diff
@@ -68,6 +68,14 @@ def convert_list_to_array(values: List[Any], dtype: str) -> bytes:
 
     return array.array(TYPE_TO_FORMAT[dtype], values)
 
+def get_array_supported_types() -> Tuple["tripy.common.datatype"]:
```
pranavm-nvidia, line 71: Is this being used anywhere?

tripy/tripy/frontend/trace/ops/storage.py
```diff
@@ -56,30 +56,27 @@ def __init__(
...
         else:
-            # If the input was a sequence, we need to copy it so that we don't take changes made
-            # to the list after the Storage op was constructed.
-            self.data = copy.copy(data)
             self.dtype = dtype if dtype else common_utils.get_element_type(data)
             self.shape = tuple(utils.get_shape(data))
+            self.data = memref.create_memref(
+                shape=self.shape,
+                dtype=self.dtype,
+                array=common_utils.convert_list_to_array(utils.flatten_list(data), dtype=self.dtype),
```
pranavm-nvidia, line 71: I thought `convert_list_to_array` works for nested lists too?

yizhuoz004: No, I have to flatten nested lists first.

tripy/tripy/common/utils.py
yizhuoz004: It's not used, let me remove it.

tripy/tripy/frontend/trace/ops/storage.py
```diff
@@ -56,30 +56,27 @@ def __init__(
...
 
         self.outputs[0].shape = list(self.shape)
 
     def str_skip_fields(self) -> Set[str]:
-        # skip data if i) it is a MemRefValue or ii) its volume exceeds threshold
-        if not isinstance(self.data, Sequence) or utils.should_omit_constant_in_str(self.shape):
-            return {"data"}
-        return set()
+        # skip data since it is always a memref value
+        return {"data"}
```
pranavm-nvidia, line 79: We actually need to be able to print small constants since otherwise we will have to lift them out as inputs. Is the problem here just that `memref` doesn't have a `__str__`? I wonder if we should add that in the MLIR-TRT bindings or just have special behavior here (we could either attach the new method dynamically, i.e. `runtime.MemRefValue.__str__ = ...` or have some type of custom behavior for `Storage.__str__`).

tripy/tripy/backend/mlir/memref.py
```diff
@@ -21,6 +21,7 @@
 import mlir_tensorrt.runtime.api as runtime
 
 from tripy.backend.mlir import utils as mlir_utils
+from tripy.common import datatype
```
pranavm-nvidia, line 24: This import probably isn't required.

tripy/tripy/frontend/tensor.py
```diff
@@ -150,7 +151,7 @@ def raw_init(
                 data = memref.create_memref_view(data)
             Storage.build_internal([], [instance.trace_tensor], data)
         else:
-            Storage.build_internal([], [instance.trace_tensor], data, dtype, device)
+            Storage.build_internal([], [instance.trace_tensor], data, None, device)
```
pranavm-nvidia, line 154: Is there ever a case where we actually want to set the datatype in `Storage`? If not I think we can drop the argument from the constructor.

tripy/tripy/frontend/tensor.py
```diff
@@ -26,6 +26,7 @@
 from tripy import export, utils
 from tripy.backend.mlir import memref
 from tripy.common import datatype
+from tripy.common import utils as common_utils
```
pranavm-nvidia, line 29: This import should be unnecessary too.

tripy/tripy/frontend/tensor.py
yizhuoz004: The only case is for empty tensor where we cannot infer dtype from list elements. Is there a better solution for it?

yizhuoz004: Oh empty tensor can be casted as well, I can remove it then.

pranavm-nvidia: Can we default it to FP32 and then rely on our casting? That's effectively what's happening here when you pass `None`

tripy/tripy/frontend/trace/ops/storage.py
yizhuoz004: Yes, `memref` does not have `__str__`. Better add that for MLIR-TRT bindings. One WAR we can do is when the data volume is small enough, we store the str representation, we can get that by `str(data)`, it works for both dlpack tensors and python lists. I've created an issue #442 for this.

pranavm-nvidia: So you mean just add a new field like `data_str` in `Storage`? We will need this to make caching work

yizhuoz004: Yes

yizhuoz004: Do you want it to be included in this PR?

pranavm-nvidia: We could do a separate PR, but it will need to happen before caching.

farazkh80: I can include it in caching PR

pranavm-nvidia: Sure, that works.

Issue #409, opened 2024-11-26
Improve behavior of evaluation during compile


pranavm-nvidia:
When a tensor is evaluated during compile, we currently raise an error or print a warning. However, we could make Tripy functionally correct even in the case of evaluation by simply not updating the frontend tensor's op to `Storage`. This would preserve the computation graph and make compilation work correctly. 

This might be as simple as adding one condition to `Tensor.eval()`:
```py
if not self.trace_tensor.is_compile_tracer:
    Storage.build_internal([], [self.trace_tensor], data)
    ...
```

We *will* need the warnings though since there may be cases where the evaluated result is erroneously used later in the graph - e.g.
```py
batch = int(x.shape[0]) # Eval happens here
tp.ones((batch, ...)) # Dynamic shapes broken here
```
We could also suppress warnings in some cases, e.g. if the tensor is only being printed. A simple way to achieve that would be to add a `suppress_warnings` parameter to `eval` and set it to `True` when calling it from `__repr__` - we will need to pass it through `tolist`, so we will likely want a `_tolist_helper` since the public `tolist` method should not have this option exposed. 

We can, however, drop the errors that are thrown in `compile`, which means we can also drop the `eval_stack_info` field of `TraceTensor`. 

Finally, to test it, we should verify the following are true when evaluating while compiling:
1. We never raise an error
2. The frontend tensor op is never updated in-place (i.e. it should not be turned into a `Storage` tensor)
3. We do not emit warnings when the evaluation is triggered by `__repr__` (i.e. we can safely assume the output is unused later in the graph)

pranavm-nvidia:
We may actually still want warnings in all cases since the extra evaluations will trigger compilation, which could make tracing extremely slow. 
A way to mitigate this could be to store the evaluated tensors somehow such that only other evaluated tensors use their values while the compiler still traces them like non-evaluated ops. 

For example, if we have a graph like:
```
A -> B -> C -> D
```
and we print `B` and `C`, we should only need to compile the `A->B` part once during evaluation and compile `C` separately.
However, in the final compiled executable, we still want the full graph, not just the `->D` part.

slyubomirsky:
Per discussion, this probably is not a priority since we could not find cases where evaluating during the compilation is useful, especially since it is already permitted for cases like printing (when the result is not used). See #443.

PR #443, opened 2024-12-12
[Tripy] Permit `eval` while tracing, but do not update trace.

slyubomirsky:
Addresses #409. Evaluation while tracing still gives a warning but does not alter the graph, so it does not produce any errors.

### DIFF
--- a/tripy/nvtripy/backend/api/compile.py
+++ b/tripy/nvtripy/backend/api/compile.py
@@ -25,6 +25,11 @@ from nvtripy.backend.mlir import Compiler as MLIRCompiler
 from nvtripy.common.exception import raise_error
 from nvtripy.frontend import Tensor, Trace
 
+# If a tensor is evaluated during the trace (and hence replaced with a Storage node),
+# we should keep track of which ones are evaluated so we could revert them tracing to
+# avoid modifying the compiled model.
+REVERT_GRAPH_AFTER_COMPILING = None
+
 
 # TODO (#230): Support collections of tensors in args/kwargs
 @export.public_api(document_under="compiling_code/compile.rst")
@@ -165,7 +170,15 @@ def compile(
     # as `InputInfo`s, but the order needs to match the signature of the original function.
     compiled_arg_names = [name for name in signature.parameters.keys() if name in input_names]
 
+    # If we evaluate any tensors when calling func(), we set up REVERT_GRAPH_AFTER_COMPILING
+    # to record any changes made to the graph (done by BaseTraceOp when it changes the trace tensors' producers)
+    # so we can revert them before proceeding.
+    global REVERT_GRAPH_AFTER_COMPILING
+    REVERT_GRAPH_AFTER_COMPILING = []
     trace_outputs = utils.make_list(func(*new_args, **new_kwargs))
+    for trace_tensor, orig_producer in REVERT_GRAPH_AFTER_COMPILING:
+        trace_tensor.producer = orig_producer
+    REVERT_GRAPH_AFTER_COMPILING = None
 
     if not trace_outputs:
         raise_error(
--- a/tripy/nvtripy/backend/api/compile.py
+++ b/tripy/nvtripy/backend/api/compile.py
@@ -184,13 +184,6 @@ def compile(
     trace_inputs = [trace_input_map[name] for name in compiled_arg_names]
     trace = Trace(trace_outputs, trace_inputs, shapes=shapes)
 
-    for op in trace.ops:
-        for tensor in op.inputs + op.outputs:
-            if tensor.is_compile_tracer and tensor.eval_stack_info is not None:
-                raise_error(
-                    "Cannot evaluate a tensor while compiling.", ["Tensor was evaluated here:", tensor.eval_stack_info]
-                )
-
     flat_ir = trace.to_flat_ir()
     mlir = flat_ir.to_mlir()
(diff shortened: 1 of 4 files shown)


slyubomirsky:
I am not fully certain how the [storage mechanism described](https://github.com/NVIDIA/TensorRT-Incubator/issues/409#issuecomment-2501952349) in the subsequent comment on #409 should work. If the main issue is recompilation caused by the fact that we not updating the trace graph if we evaluate while tracing, perhaps one approach could be to insert the Storage op anyway but restore the original graph after compilation.

tripy/tripy/frontend/tensor.py
```diff
@@ -265,8 +269,15 @@ def __repr__(self) -> str:
 
         data_list = self.tolist()
 
-        assert isinstance(self.trace_tensor.producer, Storage)
-        data_shape = self.trace_tensor.producer.shape
+        if isinstance(self.trace_tensor.producer, Storage):
```
pranavm-nvidia, line 272: Does the memref have shape information? If so, it would probably be cleaner to use the memref directly rather than `tolist()`.

tripy/tripy/frontend/tensor.py
```diff
@@ -224,11 +224,15 @@ def eval(self) -> runtime.MemRefValue:
         assert len(data) == 1, "Expects only one output from mlir_tensorrt.compiler executor"
         data = data[0]
 
-        Storage.build_internal([], [self.trace_tensor], data)
+        # If we are tracing, we should not replace the tensor with a storage op. However, even so,
+        # evaluating while tracing could introduce errors (e.g., if the evaluated value is used to create a constant
+        # that is in the graph) and is likely also to slow down compilation, so we still give warnings in that case.
+        if not self.trace_tensor.is_compile_tracer:
+            Storage.build_internal([], [self.trace_tensor], data)
```
pranavm-nvidia, line 231: The problem with this is if we evaluate multiple tensors in a compiled graph, we'll get quadratic time complexity. What I suggested in https://github.com/NVIDIA/TensorRT-Incubator/issues/409#issuecomment-2501952349 is to store the evaluated result but only use it for evaluation and not tracing. I'm not sure exactly how that would work, but I expect it would require changes to how we trace during `eval()`.

pranavm-nvidia:
> I am not fully certain how the [storage mechanism described](https://github.com/NVIDIA/TensorRT-Incubator/issues/409#issuecomment-2501952349) in the subsequent comment on #409 should work. If the main issue is recompilation caused by the fact that we not updating the trace graph if we evaluate while tracing, perhaps one approach could be to insert the Storage op anyway but restore the original graph after compilation.

Could you elaborate on this approach? Sounds promising.

slyubomirsky:
I haven't tried implementing this yet, but if we could record which ops we changed during the trace, we could avoid recompilation and then change them back afterwards.

tripy/nvtripy/frontend/tensor.py
```diff
@@ -224,11 +225,21 @@ def eval(self) -> runtime.MemRefValue:
...
 
+        # If we are tracing for compilation, we record the change we make so that we can revert it before producing
+        # the final executable. However, even so, evaluating while tracing could introduce errors
+        # (e.g., if the result of evaluation is used to create a constant that is in the graph),
+        # so we still give warnings in that case.
+        #
+        # In principle, we could avoid inserting the storage op altogether, but that could slow down other
+        # evaluations that occur during the trace.
+        if REVERT_GRAPH_AFTER_COMPILING is not None:
+            REVERT_GRAPH_AFTER_COMPILING.append((self.trace_tensor, self.trace_tensor.producer))
```
slyubomirsky, line 236: This might be a bit of a clumsy approach to keeping around the old producers (required a global var from the compiler). I wasn't sure where else the data structure could reside, since evaluation is handled as a method on tensors. It's a tricky issue because the tensor evaluated could be anywhere in the middle of the graph and the interface to `compile` is an opaque `Callable`.

tripy/nvtripy/backend/api/compile.py
```diff
@@ -25,6 +25,11 @@
 from nvtripy.common.exception import raise_error
 from nvtripy.frontend import Tensor, Trace
 
+# If a tensor is evaluated during the trace (and hence replaced with a Storage node),
+# we should keep track of which ones are evaluated so we could revert them tracing to
+# avoid modifying the compiled model.
+REVERT_GRAPH_AFTER_COMPILING = None
```
pranavm-nvidia, line 31: Shouldn't this be a property of the trace rather than something global?

tripy/nvtripy/frontend/tensor.py
pranavm-nvidia: What about storing it on the trace tensors themselves? `original_producer` or something like that?

pranavm-nvidia: I guess in that case we would need to do a DFS from the output and swap the producer before stepping further

slyubomirsky: Yeah searching the graph might be necessary if there isn't anywhere else we could record the tensors that have been evaluated.

slyubomirsky:
Closing per discussion: We've decided that there aren't any cases we could think of where evaluating while tracing is useful, especially since evaluating without using the result (e.g., just printing) already worked. We can revisit if such a case does present itself.

PR #446, opened 2024-12-13
Adds float8 support back to nanoGPT

pranavm-nvidia:


pranavm-nvidia:
Cannot merge as this is still failing

PR #461, opened 2024-12-18
Reduce nanogpt quantization calib size to speed up test

yizhuoz004:


tripy/examples/nanogpt/quantization.py
```diff
@@ -43,10 +45,10 @@ def modelopt_quantize(model_hf, quant_mode):
     else:
         raise NotImplementedError(f"Unsupported quantization mode: {quant_mode}")
 
-    calib_size = 512
+    calib_size = 64
     batch_size = 1
     if quant_mode == "int4-weight-only":
-        calib_size = 32
+        calib_size = 16
```
pranavm-nvidia, line 51: Why do we use a smaller size for INT4?

tripy/examples/nanogpt/quantization.py
```diff
@@ -43,10 +45,10 @@ def modelopt_quantize(model_hf, quant_mode):
     else:
         raise NotImplementedError(f"Unsupported quantization mode: {quant_mode}")
 
-    calib_size = 512
+    calib_size = 64
```
pranavm-nvidia, line 48: Does this impact the accuracy?

pranavm-nvidia (approved):

tripy/examples/nanogpt/quantization.py
yizhuoz004: int4 quantization is very slow, modelopt recommends using a smaller calib size with a bigger batch size.

tripy/examples/nanogpt/quantization.py
yizhuoz004: No, the outputs are all the same.

PR #470, opened 2025-01-07
Fix sam2 sample artifact removal

yizhuoz004:


tripy/tests/test_examples.py
```diff
@@ -84,7 +86,7 @@ def __str__(self):
     Example(["nanogpt"]),
     Example(
         ["segment-anything-model-v2"],
-        artifact_names=["truck.jpg", "bedroom", "saved_engines/", "output/", "checkpoints/"],
+        artifact_names=["truck.jpg", "bedroom", "saved_engines/", "output/", "checkpoints/*.pt"],
```
pranavm-nvidia, line 89: Just curious - was this broken before and not removing the `checkpoints/` directory? Or are we just making sure that checkpoints are actually generated here?

pranavm-nvidia (approved):

parthchadha (approved):

tripy/tests/test_examples.py
yizhuoz004: This was removing the whole `checkpoints` directory, but we want to keep `checkpoints/download_ckpt.sh` there and only removing the `*.pt` files.

pranavm-nvidia:
The perf regression looks like noise due to not being able to lock the clocks. I think this is good to merge. If we see this happen more frequently, we might want to consider bumping the threshold to 10%.

PR #475, opened 2025-01-11
Move internal changes

christopherbate:
## [tensorrt] Fix various edge cases in 'tensorrt-broadcast-elimination'

This change fixes several edge cases in 'tensorrt-broadcast-elimination'
for a particular pattern that commutes 'tensorrt.broadcast' and
'tensorrt.collapse_rank'. Previously it did not correctly handle multiple
collapsed unit dimensions and would produce an erroneous assertion when
the collapsed input is dynamic. Several additional regression tests
are added.

## [tensorrt] Fix nvinfer1::Dims max rank assertion

Fixes an assertion which limited translation of MLIR to TensorRT to
programs with tensors of rank < 8 when it should be <= 8 to align
with TensorRT op verification and the actual nvinfer API. Adds an
additional translation test.

## Fix options configuration to enable global flags in "compiler task" API

### DIFF
--- a/mlir-tensorrt/CMakeLists.txt
+++ b/mlir-tensorrt/CMakeLists.txt
@@ -253,6 +253,14 @@ include(AddMLIRPython)
 include(MLIRDetectPythonEnv)
 mlir_configure_python_dev_packages()
 
+# Declare the Python source targets ahead-of-time since the sources may be built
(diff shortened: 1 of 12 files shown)


shelkesagar29 (approved):
LGTM.

PR #478, opened 2025-01-17
SAM2: Save engines to different paths according to model type

yizhuoz004:


tripy/examples/segment-anything-model-v2/sam2/build_sam.py
```diff
@@ -323,7 +324,7 @@ def build_sam2_base(
 
     # Get component configurations and initialize cache
     components = get_component_configs(model, cfg)
-    model_cache = SAM2ModelCache()
+    model_cache = SAM2ModelCache(model_type)
```
pranavm-nvidia, line 327: Do we need something fancier like hashing the config? Or do we expect the config not to change?

pranavm-nvidia (approved):

tripy/examples/segment-anything-model-v2/sam2/build_sam.py
yizhuoz004: Right, we don't expect config to change.

PR #479, opened 2025-01-17
Updates Tripy to work with Python 3.9, moves files into utils

pranavm-nvidia:
- Updates the `Dockerfile` to use a Python 3.9 container (the lowest version we support) so we can be sure we're not relying on features that are not available there.

     This unforunately means that we need to drop `lldb` from the container
     since it depends on Python 3.10. However, `gdb` is still available.

- Moves various files into `utils`. We were previously unable to do this because the `__init__.py` file of `utils` would import lots of things, creating circular dependencies. Now, the top-level `utils` module will not import everything, meaning there is no risk of circular dependencies.

- Makes some minor changes in the code to make Python 3.9 work (e.g. using `._name` instead of `.__name__`)

### DIFF
--- a/tripy/Dockerfile
+++ b/tripy/Dockerfile
@@ -1,10 +1,10 @@
-FROM ubuntu:22.04
+FROM python:3.9
 
 LABEL org.opencontainers.image.description="Tripy development container"
 
 WORKDIR /tripy
 
-SHELL ["/bin/bash", "-c"]
+ENTRYPOINT ["/bin/bash"]
 
 # Setup user account
 ARG uid=1000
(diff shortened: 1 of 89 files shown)


tripy/Dockerfile
```diff
@@ -30,21 +30,5 @@ RUN pip install build .[docs,dev,test,build] \
...
-    echo "deb-src http://apt.llvm.org/jammy/ llvm-toolchain-jammy-$LLVM_VERSION main" >> /etc/apt/sources.list.d/llvm.list && \
-    wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key 2>/dev/null > /etc/apt/trusted.gpg.d/apt.llvm.org.asc && \
-    wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | gpg --dearmor - | tee /usr/share/keyrings/kitware-archive-keyring.gpg >/dev/null && \
-    echo 'deb [signed-by=/usr/share/keyrings/kitware-archive-keyring.gpg] https://apt.kitware.com/ubuntu/ jammy main' | tee /etc/apt/sources.list.d/kitware.list >/dev/null && \
-    apt-get update && \
-    apt-get install -y ${LLVM_PACKAGES} && \
-    apt-get clean -y && \
-    rm -rf /var/lib/apt/lists/* && \
-    ln -s /usr/bin/lldb-17 /usr/bin/lldb
-
```
yizhuoz004, line 48: Small concern if we drop lldb -- the dev container is mainly for tripy devs, and lldb is useful to debug MLIR-TRT. Or we can add this section to `Debugging.md`.

Also will this affect MLIR-TRT wheels? I need to double check if they are sensitive to python versions.

pranavm-nvidia: @yizhuoz004 how much does lldb add compared to gdb?

MLIR-TRT posts wheels for 3.9+, so that should be fine.

yizhuoz004: Haven't used gdb for this project, but from @jhalakpatel it seems lldb is quite essential to debug with llvm.

pranavm-nvidia: I see, maybe we can have a separate debugging container? I think it's more important to develop in a 3.9 container so we don't break older Python versions - using `lldb` will probably be the less common case. If/when we drop 3.9 support, then we'd be able to merge the containers again.

yizhuoz004: Sounds good

PR #480, opened 2025-01-18
Refactors quantization guide 

pranavm-nvidia:
[Refactors quantization guide](https://github.com/NVIDIA/TensorRT-Incubator/pull/480/commits/a62af1053a712cd5066becc130f34b6bf5f1d980)
- Refactors quantization guide to be more concise
- Updates `nvidia-modelopt` version so we don't need a special package index
- Reduces threshold for summary mode when pretty printing tensors.
- Adds a `# doc: no-output` tag to omit output from code examples.

### DIFF
--- a/tripy/docs/pre0_user_guides/01-quantization.md
+++ b/tripy/docs/pre0_user_guides/01-quantization.md
@@ -1,154 +1,235 @@
 # Quantization
 
+**Quantization** reduces memory and compute requirements by running operations in low precision:
+- **Scaling** is required to translate to/from low precision.
+- **Scaling factors** are chosen such that they minimize accuracy loss.
+- They can be either:
+    - Loaded into quantization-enabled {class}`nvtripy.Module`s, or
+    - Used with {func}`nvtripy.quantize`/{func}`nvtripy.dequantize`.
+
+:::{seealso}
+The
+[TensorRT developer guide](https://docs.nvidia.com/deeplearning/tensorrt/developer-guide/index.html#working-with-int8)
+explains quantization in more detail.
+:::
 
 
-## Using Quantized Modules
+## Post-Training Quantization With ModelOpt
 
-Various modules predefined by Tripy support quantization. For example, the {class}`nvtripy.Linear`
-module includes two arguments to configure the quantization mode. Let's construct the following
-quantized linear module:
+If the model was not trained with quantization-aware training (QAT), we can use
+[TensorRT ModelOpt](https://nvidia.github.io/TensorRT-Model-Optimizer/index.html)
+to do **calibration** to determine scaling factors.
 
-```py
-# doc: print-locals quant_linear
-quant_linear = tp.Linear(
-    4,
-    2,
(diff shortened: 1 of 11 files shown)


tripy/docs/pre0_user_guides/01-quantization.md
```diff
@@ -1,154 +1,140 @@
 # Quantization
 
+**Quantization** can reduce a model's memory and compute requirements by running operations in a lower precision.
 
+- **Scaling** is required to translate to/from low precision.
```
pranavm-nvidia, line 5: Should I mention per-tensor/per-dimension/blockwise here?

pranavm-nvidia: It is already in the TRT developer guide that we link to below, so not sure how much I want to duplicate here. OTOH it would be good to provide at least a basic understanding of how quantization works as it pertains to using it in Tripy.

yizhuoz004: We can add a link to `tp.quantize`, where the basic logic of quantization is explained, e.g. per-tensor / per-channel etc.

yizhuoz004: The quantization guide better also provide some basic information like that, right now it directly jumps to using a "quantized module" -- we could also explain when to use `tp.quantize` v.s. quantized modules.

pranavm-nvidia: @yizhuoz004 I think the section in the TRT dev guide is better since it isn't attached to a specific API. A conceptual explanation is really what I'm looking for. If we point to `tp.quantize` there might be too many distractions (e.g. behavior/limitations specific to that one API rather than a general explanation of quantization)

yizhuoz004: Yes it makes sense.

pranavm-nvidia: Yes, I was just debating whether I should add that now or in a separate PR.

tripy/docs/pre0_user_guides/01-quantization.md
```diff
@@ -1,154 +1,230 @@
...
 ```
 
-Let's convert the `amax` to the scaling factor and load it to a compatible {class}`nvtripy.Linear` module:
+:::{seealso}
+`load_quant_weights_from_hf` in the [nanoGPT weight loader](source:/examples/nanogpt/weight_loader.py)
+is an example of loading scaling factors for an entire model.
+:::
+
 
+### Manually
```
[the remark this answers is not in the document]
pranavm-nvidia: @yizhuoz004 mentioned the two options at the top. Can you take another look?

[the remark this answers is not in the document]
yizhuoz004: Looks good, thanks!

yizhuoz004 (approved):

PR #489, opened 2025-02-05
Overhaul backend function execution for improved performance and flex…

jhalakpatel:
…ibility

This PR replaces the DPS-style calling convention with a non-DPS approach, eliminating the requirement for call sites to preallocate output buffers. This change enables us to bypass the computation of output shapes and advance allocation of output buffers, laying the groundwork for supporting data-dependent shapes where network outputs can have dynamic dimensions.

The underlying compiler stack has been enhanced to avoid allocating oversized buffers and eliminate an extra device-to-device copy operation from TensorRT-allocated memory to MLIR-TRT managed memory.

Additionally, we've improved the copy operation to support copying to host memory. This enhancement removes the need to track output device allocations for device-to-host copies. Previously, copy outputs were restricted to device allocations; now they can be allocated on both device and host.

Tests have been updated to align with the new calling convention, ensuring compatibility and correctness.

Other changes:
Fix type constraints tests
Address review comments

### DIFF
--- a/tripy/nvtripy/backend/api/executable.py
+++ b/tripy/nvtripy/backend/api/executable.py
@@ -15,7 +15,7 @@
 import base64
 import inspect
 from dataclasses import dataclass
-from typing import Sequence, Tuple, Union
+from typing import Sequence, Tuple, Union, Callable
 
 import mlir_tensorrt.runtime.api as runtime
 from nvtripy import export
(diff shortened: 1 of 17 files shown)


tripy/nvtripy/backend/api/executable.py
```diff
@@ -46,21 +46,19 @@ class Executable:
     """
 
     # The constructor is intentionally undocumented because it is not meant to be called by users.
-    # TODO(#155): output_devices is not needed after they can be queried from executable
```
pranavm-nvidia, line 49: Does this PR also resolve 155?

tripy/tests/integration/test_dequantize.py
```diff
@@ -29,6 +29,7 @@ class TestDequantize:
     @pytest.mark.parametrize(
         "dtype", [tp.float32, tp.float16, pytest.param(tp.bfloat16, marks=skip_if_older_than_sm80)]
     )
+    @pytest.mark.skip("StableHLO QDQ broken")
```
pranavm-nvidia, line 32: Can we xfail these instead of skip? That way we will remember to turn them on again when we switch to TRT dialect.

pranavm-nvidia:
Closing stale PRs

PR #490, opened 2025-02-06
Add script to generate tripy integration test stablehlo ir

jhalakpatel:
This script can be run periodically (or during CI/CD) to generate Tripy integration test IRs. For now, we generate stablehlo IR but script itself is IR agonistic. This will generate a file `test_tripy_integration.py` which can be copied to `mlir-tensorrt/compiler/test/python/IntegrationTests/test_tripy_integration.py` and committed to mitigating Tripy related regressions in MLIR-TensorRT.

### DIFF
--- a/tripy/scripts/generate_tripy_integration_ir.py
+++ b/tripy/scripts/generate_tripy_integration_ir.py
@@ -1,0 +1,123 @@
+import os
+import glob
+import tempfile
+import subprocess
+import re
+import shutil
+
+def construct_test_names(output):
+    relevant_output = output.split('\n\n=======================')[0]
+    module_match = re.search(r'<Module (.*?)>', relevant_output)
+    module_name = module_match.group(1) if module_match else ''
+    class_match = re.search(r'<Class (.*?)>', relevant_output)
+    class_name = class_match.group(1) if class_match else ''
+    function_matches = re.findall(r'<Function (.*?)>', relevant_output)
+    
+    test_names = [
+        f"/tripy/{module_name}::{class_name}::{func}" 
+        for func in function_matches
+    ]
+    return test_names
+
+def run_single_test(test_name):
+    with tempfile.TemporaryDirectory() as test_dir:
+        # Set environment for this test
+        test_env = os.environ.copy()
+        test_env['TRIPY_MLIR_DEBUG_ENABLED'] = '1'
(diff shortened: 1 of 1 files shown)


tripy/scripts/generate_tripy_integration_ir.py
```diff
@@ -0,0 +1,123 @@
...
+import mlir_tensorrt.compiler.api as api
+from mlir_tensorrt.compiler.ir import *
+
+# Store assemblies in a dictionary with test names as keys
+TEST_ASSEMBLIES = {
+""")
+        
+        for test_name, asm in test_assemblies.items():
+            f.write(f'    "{test_name}": """{asm}""",\n')
+
```
pranavm-nvidia, line 81: How large is the generated code usually? Wondering if we might need to split them across tests instead of just test files?

tripy/scripts/generate_tripy_integration_ir.py
```diff
@@ -0,0 +1,123 @@
...
+import subprocess
+import re
+import shutil
+
+def construct_test_names(output):
+    relevant_output = output.split('\n\n=======================')[0]
+    module_match = re.search(r'<Module (.*?)>', relevant_output)
+    module_name = module_match.group(1) if module_match else ''
+    class_match = re.search(r'<Class (.*?)>', relevant_output)
+    class_name = class_match.group(1) if class_match else ''
```
jhalakpatel, line 13: This might be incorrect if we have more than one class per module.

tripy/scripts/generate_tripy_integration_ir.py
jhalakpatel: I have not generated the code for the complete test suit (WIP). I will update once done.

jhalakpatel: Generated code is about 30k lines and takes around 15 minutes to run end to end in MLIR-TRT. May be we can perform some deduplication i.e. just include first test case in a class. or may be just include a single test per module.

pranavm-nvidia: Does a naive deduplication help at all? e.g. just comparing the generated StableHLO textual representation directly

pranavm-nvidia:
Closing stale PRs

PR #494, opened 2025-02-11
[Compiler] Add force-entrypoints-return-allocs option to TRT task

yizhuoz004:
Add force-entrypoints-return-allocs option to TRT task, also update the integration test to use non-DPS style calling convention.


jhalakpatel (approved):
LGTM!

christopherbate (approved):

PR #498, opened 2025-02-13
[Allocator] Fix OOM issues in non-DPS mode

jhalakpatel:
Summary:
- Restore querying of `rankOffset` from the output descriptor, which was inadvertently removed in a previous merge. Update the related test `tensorrt-runtime-to-executor.mlir`.
- Enhance debugging logs to provide clearer error messages.
- Revert `Allocator::track` to its previous implementation for cases where the incoming pointer is not managed internally by `Allocator`.
- Modify `Allocator::track` to correctly handle cases where an incoming pointer is already managed internally. This adjustment is necessary as `Allocator` now tracks pointers internally when they are returned as function results.
- Refine `Allocator::safeDeallocate` to ensure it only releases pointers that are managed internally.
- Correct a typo in `Allocator::safeDeallocate` for `PointerType::pinned_host`, which previously caused an error. Adjust to log a message deferring memory deallocation to `PinnedMemoryAllocator`.
- Address an issue in external memref creation by avoiding retracking of pointers already managed by `Allocator`. This prevents redundant tracking as externally managed.
- Ensure that when populating arguments for an Enqueue function, the session allocator tracks pointers as internally managed if they are managed by the client allocator. Ensure the session tracker does not assume ownership for deallocation.
- Prevent `~Allocator()` from releasing pointers that have already been released internally.

### DIFF
--- a/mlir-tensorrt/compiler/lib/Conversion/TensorRTRuntimeToExecutor/TensorRTRuntimeToExecutor.cpp
+++ b/mlir-tensorrt/compiler/lib/Conversion/TensorRTRuntimeToExecutor/TensorRTRuntimeToExecutor.cpp
@@ -375,6 +375,10 @@ struct ConvertEnqueueAllocToCall
     for (auto [idx, result] : llvm::enumerate(op.getResults())) {
       MemRefType memrefType = cast<MemRefType>(result.getType());
       unsigned rank = memrefType.getRank();
+
+      // Skip the rank offset that is populated by the callee.
+      outputDescOffset++;
+
       Value devicePtrOffset = b.create<executor::GetOffsetOp>(
           i64Type, structType,
           ArrayRef<OpFoldResult>{
--- a/mlir-tensorrt/compiler/test/Conversion/TensorRTRuntimeToExecutor/tensorrt-runtime-to-executor.mlir
+++ b/mlir-tensorrt/compiler/test/Conversion/TensorRTRuntimeToExecutor/tensorrt-runtime-to-executor.mlir
@@ -79,24 +79,26 @@ func.func @convert_enqueue_alloc(%arg0: memref<?xf32, #device>,
 //       CHECK:     %[[v12:.+]] = executor.table.get %[[v0]][3] : <!executor.ptr<device>, !executor.ptr<device>, i64, i64, i64>
 //       CHECK:     %[[v13:.+]] = executor.table.create(%[[v8]], %[[c0_i64]], %[[c2_i64]], %[[v9]], %[[v10]], %[[v11]], %[[c0_i64]], %[[c1_i64]], %[[v12]] : !executor.ptr<device>, i64, i64, i64, i64, !executor.ptr<device>, i64, i64, i64) : <!executor.ptr<device>, i64, i64, i64, i64, !executor.ptr<device>, i64, i64, i64>
 //       CHECK:     executor.call @_trtrt_enqueue_alloc(%[[v3]], %[[v2]], %[[v4]], %[[v13]]) : (!executor.ptr<host>, !executor.ptr<host>, !executor.ptr<host>, !executor.table<!executor.ptr<device>, i64, i64, i64, i64, !executor.ptr<device>, i64, i64, i64>) -> ()
-//       CHECK:     %[[v14:.+]] = executor.load %[[v4]] + %[[v6]] : (!executor.ptr<host>, i64) -> i64
-//       CHECK:     %[[v15:.+]] = executor.inttoptr %[[v14]] : (i64) -> !executor.ptr<device>
-//       CHECK:     %[[v16:.+]] = executor.getoffset[0, 2] : () -> i64, !executor.table<i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64>
-//       CHECK:     %[[v17:.+]] = executor.load %[[v4]] + %[[v16]] : (!executor.ptr<host>, i64) -> i64
+//       CHECK:     %[[v14:.+]] = executor.getoffset[0, 2] : () -> i64, !executor.table<i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64>
+//       CHECK:     %[[v15:.+]] = executor.load %[[v4]] + %[[v14]] : (!executor.ptr<host>, i64) -> i64
+//       CHECK:     %[[v16:.+]] = executor.inttoptr %[[v15]] : (i64) -> !executor.ptr<device>
 //       CHECK:     %[[v18:.+]] = executor.getoffset[0, 3] : () -> i64, !executor.table<i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64>
 //       CHECK:     %[[v19:.+]] = executor.load %[[v4]] + %[[v18]] : (!executor.ptr<host>, i64) -> i64
-//       CHECK:     %[[v20:.+]] = executor.table.create(%[[v15]], %[[v15]], %[[c0_i64]], %[[v17]], %[[v19]] : !executor.ptr<device>, !executor.ptr<device>, i64, i64, i64) : <!executor.ptr<device>, !executor.ptr<device>, i64, i64, i64>
-//       CHECK:     %[[v21:.+]] = builtin.unrealized_conversion_cast %[[v20]] : !executor.table<!executor.ptr<device>, !executor.ptr<device>, i64, i64, i64> to memref<?xf32, #executor.memory_type<device>>
-//       CHECK:     %[[v22:.+]] = executor.getoffset[0, 4] : () -> i64, !executor.table<i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64>
-//       CHECK:     %[[v23:.+]] = executor.load %[[v4]] + %[[v22]] : (!executor.ptr<host>, i64) -> i64
-//       CHECK:     %[[v24:.+]] = executor.inttoptr %[[v23]] : (i64) -> !executor.ptr<host>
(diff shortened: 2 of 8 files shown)


mlir-tensorrt/executor/lib/Runtime/API/API.cpp
```diff
@@ -648,19 +674,18 @@ mlirtrt::Status runtime::safeDeallocate(AllocTracker &tracker, uintptr_t ptr,
...
     tracker.untrack(ptr);
     return Status::getOk();
   }
 
-  if (obj.type != PointerType::pinned_host) {
-    MTRT_DBG("Synchronously freeing pinned host memory {0}",
-             reinterpret_cast<void *>(obj.ptr));
-    RETURN_ERROR_IF_CUDART_ERROR(
-        cudaFreeHost(reinterpret_cast<void *>(obj.ptr)));
+  if (obj.type == PointerType::pinned_host) {
```
jhalakpatel, line 685: @christopherbate This looks like a bug which we missed to fix.

mlir-tensorrt/executor/lib/Runtime/API/API.cpp
```diff
@@ -457,29 +471,62 @@ void AllocTracker::decrementExternalCount(uintptr_t ptr) {
...
+             llvm::formatv("Pointer {0} in track is internally managed but "
+                           "should be released internally",
+                           info.ptr)
+                 .str()
+                 .c_str());
+      const PointerInfo &existingInfo = get(info.ptr);
+      bool infoMismatch = existingInfo.size != info.size ||
+                          existingInfo.type != info.type ||
+                          existingInfo.owner != info.owner;
+      if (infoMismatch) {
```
jhalakpatel, line 503: This is OK to have info mismatch since the pointer is internally released.

mlir-tensorrt/compiler/lib/Conversion/TensorRTRuntimeToExecutor/TensorRTRuntimeToExecutor.cpp
```diff
@@ -375,6 +375,13 @@ struct ConvertEnqueueAllocToCall
     for (auto [idx, result] : llvm::enumerate(op.getResults())) {
       MemRefType memrefType = cast<MemRefType>(result.getType());
       unsigned rank = memrefType.getRank();
+
+      Value rankOffset = b.create<executor::GetOffsetOp>(
```
christopherbate, line 379: Just do `outputDescOffset++`. It's not necessary to create extra SSA values that will just get removed later.

mlir-tensorrt/executor/include/mlir-executor/Support/Status.h
```diff
@@ -239,6 +239,13 @@ class StatusOr {
     }                                                                          \
   } while (false);
 
+#define REPORT_FATAL_ERROR_IF(cond, msg)                                       \
```
christopherbate, line 242: If the point of this is to replace `assert` that you were using before, then I think we should wrap this definition to check `NDEBUG`.

Also, please prefix new macros with `MTRT_`. I think we can simplify the name to something like `MTRT_CHECK`:

```
#ifndef NDEBUG
#define MTRT_CHECK(cond, msg) ......
#else
// In Release mode, compiles to a no-op.
#define MTRT_CHECK(cond, msg)
#endif

File: mlir-tensorrt/compiler/test/Conversion/TensorRTRuntimeToExecutor/tensorrt-runtime-to-executor.mlir [hunk 5]
```diff
@@ -79,24 +79,26 @@ func.func @convert_enqueue_alloc(%arg0: memref<?xf32, #device>,
...
-//       CHECK:     %[[v21:.+]] = builtin.unrealized_conversion_cast %[[v20]] : !executor.table<!executor.ptr<device>, !executor.ptr<device>, i64, i64, i64> to memref<?xf32, #executor.memory_type<device>>
-//       CHECK:     %[[v22:.+]] = executor.getoffset[0, 4] : () -> i64, !executor.table<i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64>
-//       CHECK:     %[[v23:.+]] = executor.load %[[v4]] + %[[v22]] : (!executor.ptr<host>, i64) -> i64
-//       CHECK:     %[[v24:.+]] = executor.inttoptr %[[v23]] : (i64) -> !executor.ptr<host>
-//       CHECK:     %[[v25:.+]] = executor.load %[[v4]] + %[[v7]] : (!executor.ptr<host>, i64) -> i64
-//       CHECK:     %[[v26:.+]] = executor.getoffset[0, 6] : () -> i64, !executor.table<i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64>
-//       CHECK:     %[[v27:.+]] = executor.load %[[v4]] + %[[v26]] : (!executor.ptr<host>, i64) -> i64
+//       CHECK:     %[[v20:.+]] = executor.getoffset[0, 4] : () -> i64, !executor.table<i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64>
+//       CHECK:     %[[v21:.+]] = executor.load %[[v4]] + %[[v20]] : (!executor.ptr<host>, i64) -> i64
```
christopherbate, line 88: Looks like you forgot to update the CHECK statements after your last update

mlir-tensorrt/compiler/test/Conversion/StablehloToTensorRT/stablehlo-to-tensorrt-trt10.mlir
```diff
@@ -241,15 +242,15 @@ func.func @scatter_slice_update_i1_axis1(%arg0: tensor<1x134xi1>, %arg1: tensor<
   return %0 : tensor<1x134xi1>
 }
 
-// CHECK-LABEL: @scatter_slice_update_i1
+// CHECK-LABEL: @scatter_slice_update_i1_axis1
 //       CHECK:     %[[v0:.+]] = tensorrt.slice %arg1[0, 1][1, 1][1, 1] : tensor<1x2xi32> to tensor<1x1xi32>
 //       CHECK:     %[[v1:.+]] = tensorrt.collapse_rank %[[v0]] : tensor<1x1xi32> to tensor<i32>
```
christopherbate, line 247: It looks like changes in this file are unrelated? Please drop them

mlir-tensorrt/compiler/test/python/mlir_tensorrt_runtime/test_create_memref.py
```diff
@@ -414,7 +415,7 @@ def memref_alloc():
 test_released_internally()
 
 # CHECK-LABEL: Test memref is released internally with an external reference
```
christopherbate, line 417: Looks like we should also update the naming/commentary of this test to "not released internally"?

mlir-tensorrt/compiler/test/Conversion/TensorRTRuntimeToExecutor/tensorrt-runtime-to-executor.mlir
jhalakpatel: @christopherbate This is correct since earlier, we needed an extra `getOffset` at line 82 to create v14 (i.e. offset [0, 2]). Earlier, we used v6, i.e. `executor.load %[[v4]] + %[[v6]]`, where v6 was already generated somewhere above in the IR since the offset was incorrectly `[0, 1]` instead of `[0, 2]`. Let me know if its still unclear.

mlir-tensorrt/compiler/test/Conversion/StablehloToTensorRT/stablehlo-to-tensorrt-trt10.mlir
```diff
@@ -241,15 +242,15 @@ func.func @scatter_slice_update_i1_axis1(%arg0: tensor<1x134xi1>, %arg1: tensor<
...
 //       CHECK:     %[[v2:.+]] = tensorrt.collapse_rank %arg2 : tensor<1x1x5xi1> to tensor<1x5xi1>
 //       CHECK:     %[[v3:.+]] = tensorrt.constant dense<1> : tensor<2xi32>
 //       CHECK:     %[[v4:.+]] = tensorrt.linspace[%[[v1]] : tensor<i32>] [ static] [%[[v3]] : tensor<2xi32>] : tensor<1x5xi32>
-//       CHECK:     %[[v5:.+]] = tensorrt.identity %arg0 : tensor<1x134xi1> to tensor<1x134xi32>
-//       CHECK:     %[[v6:.+]] = tensorrt.identity %[[v2]] : tensor<1x5xi1> to tensor<1x5xi32>
-//       CHECK:     %[[v7:.+]] = tensorrt.scatter_elements {axis = 1 : i64} data(%[[v5]] : tensor<1x134xi32>) indices(%[[v4]] : tensor<1x5xi32>) updates(%[[v6]] : tensor<1x5xi32>)
+//       CHECK:     %[[v5:.+]] = tensorrt.identity %[[v2]] : tensor<1x5xi1> to tensor<1x5xi32>
```
jhalakpatel, line 251: For some reason, the instructions are swapped for me when I test locally.

christopherbate: Will  need to change these to `CHECK-DAG`.

Anyway, it looks like these changes will cause failure in CI. I suspect you are somehow on an old LLVM commit. Can you drop the change, push, and let's see if CI passes?

christopherbate (approved):

PR #510, opened 2025-02-20
Ports `cast` and `shape` operations

pranavm-nvidia:
[Ports the cast operation to the TRT dialect](https://github.com/NVIDIA/TensorRT-Incubator/commit/ecad27e9dd196c3a7ad7a7898fc6fc9fd361d9c1)

[Ports shape operation to TRT dialect](https://github.com/NVIDIA/TensorRT-Incubator/commit/6340a21de921bc12f7b31cf857394cbf6746685b)

Also adds an optimization for eager mode where the output of `Shape` will
be stored if we evaluate a `GetDimensionSize` so that subsequent `GetDimensionSize`s
do not need to evaluate the entire graph - otherwise, we recompile the graph
once per dimension of the shape.

### DIFF
--- a/tripy/nvtripy/frontend/ops/shape.py
+++ b/tripy/nvtripy/frontend/ops/shape.py
@@ -1,5 +1,5 @@
 #
-# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -19,7 +19,7 @@
 from nvtripy.common.datatype import DATA_TYPES
 from nvtripy.frontend.ops import utils as op_utils
 from nvtripy.frontend.ops._registry import register_tensor_method
-from nvtripy.trace.ops.shape import GetDimensionSize
+from nvtripy.trace.ops.shape import Shape, GetDimensionSize
 from nvtripy.types import ShapeLike
 from nvtripy.utils import wrappers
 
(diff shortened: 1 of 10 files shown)


tripy/nvtripy/frontend/ops/shape.py
```diff
@@ -49,7 +49,8 @@ def shape(self: "nvtripy.Tensor") -> ShapeLike:
     if all(dim >= 0 for dim in self.trace_tensor.shape) and not self.trace_tensor.is_compile_tracer:
         return self.trace_tensor.shape
 
+    shape = op_utils.create_op(Shape, [self])
     return [
-        op_utils.create_op(GetDimensionSize, [self], dim=index, always_cast_to_dimension_size=True)
+        op_utils.create_op(GetDimensionSize, [shape], dim=index, always_cast_to_dimension_size=True)
```
pranavm-nvidia, line 54: Wondering if I should change the name of `GetDimensionSize` - it's now basically just a special kind of slice op. Maybe `ExtractDimension` instead?

tripy/nvtripy/trace/ops/shape.py
```diff
@@ -28,11 +43,19 @@ def infer_rank(self):
...
-
         self.outputs[0].dtype = int32
 
-    def to_flat_ir(self, inputs, outputs):
-        from nvtripy.flat_ir.ops import GetDimensionSizeOp
-
-        GetDimensionSizeOp.build(inputs, outputs, self.dim)
+    def to_mlir(self, inputs, outputs):
+        # TODO (pranavm): Figure out if there's a cleaner way to do this.
+        # Might want to split this into multiple trace ops and compose them in the frontend.
```
yizhuoz004, line 50: I'm also wondering if `GetDimensionSize` is necessary as a trace op. Is it equivalant to `input.shape[dim]` in the frontend?

pranavm-nvidia: @yizhuoz004 maybe I can add a TODO a come back to this once I add Slice/Squeeze? The `slice` frontend op actually requires `shape`, so it's a bit circular

yizhuoz004: Sounds good

yizhuoz004 (approved):

PR #517, opened 2025-02-20
[TensorRT] Relax same-shape constraint for TRT QDQ ops

yizhuoz004:
Changes same-shape constraint for QDQ ops to same-rank, so that we can generate output result type with unknown dims.


shelkesagar29 (approved):

christopherbate (commented):
To accomplish this, you just need to add static type inference routine like all the other TRT ops. That will enable allowing type differences where one sim is static and another dynamic or both dynamic. Just removing any shape constraint isn't the right approach as it would allow two different static values

PR #522, opened 2025-02-21
[TensorRT] Add type inference for QDQ ops

yizhuoz004:
Adds type inference for tensorrt QDQ ops to ensure static dims of input/output types are the same.


christopherbate (approved):
Looks good, please just add an invalid test for each op to `tensorrt/test/Dialect/TensorRT/invalid.mlir`

PR #523, opened 2025-02-21
Avoids broadcast elimination for dynamic shapes

pranavm-nvidia:
When multiple input operands of an operation (e.g. elementwise, select) are being broadcasted along the same dimension to a larger one, we cannot eliminate the broadcast. The logic to do this was already in place, but did not account for dynamic dimensions.

### DIFF
--- a/mlir-tensorrt/tensorrt/lib/TensorRT/Transforms/BroadcastElimination.cpp
+++ b/mlir-tensorrt/tensorrt/lib/TensorRT/Transforms/BroadcastElimination.cpp
@@ -262,7 +262,12 @@ static bool broadcastAbsorbPreconditions(BroadcastOp broadcastOp) {
          broadcastOp.getBroadcastDimsPermutation().isIdentity();
 }
 
+static bool isOneOrDynamic(int64_t dimSize) {
+  return dimSize == 1 || ShapedType::isDynamic(dimSize);
+}
+
 namespace {
+
 /// A `tensorrt.element_wise` operation can pull in `tensorrt.broadcast`
 /// operations since the elementwise op has implicit broadcasting semantics.
 struct ElementwiseAbsorbBroadcast : public OpRewritePattern<ElementWiseOp> {
@@ -287,11 +292,13 @@ public:
       return failure();
 
     // You can't eliminate both broadcasts if the same unit-dim in both
-    // operands is being broadcast to a larger value. We can do some further
-    // simplification, but we leave that to other patterns.
+    // operands is being broadcast to a larger value. We can
+    // do some further simplification, but we leave that to other patterns.
+    // If there are dynamic shapes, we must be extra conservative.
     for (unsigned i = 0; i < input1.getType().getRank(); i++) {
-      if (input1.getType().getDimSize(i) == 1 &&
-          input2.getType().getDimSize(i) == 1 && op.getType().getDimSize(i) > 1)
+      if (isOneOrDynamic(input1.getType().getDimSize(i)) &&
+          isOneOrDynamic(input2.getType().getDimSize(i)) &&
+          op.getType().getDimSize(i) != 1)
         return failure();
     }
     rewriter.replaceOpWithNewOp<ElementWiseOp>(op, op.getType(), input1, input2,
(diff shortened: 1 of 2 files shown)


mlir-tensorrt/tensorrt/lib/TensorRT/Transforms/BroadcastElimination.cpp
```diff
@@ -263,6 +263,11 @@ static bool broadcastAbsorbPreconditions(BroadcastOp broadcastOp) {
 }
 
 namespace {
+
```
pranavm-nvidia, line 266: This change seems to be causing a regression in `broadcast_dynamic_expand_shape_regression` but I haven't looked into why yet.

mlir-tensorrt/tensorrt/lib/TensorRT/Transforms/BroadcastElimination.cpp
```diff
@@ -263,6 +263,11 @@ static bool broadcastAbsorbPreconditions(BroadcastOp broadcastOp) {
...
+static bool isOneOrDynamic(int64_t dimSize) {
+  return dimSize == 1 || dimSize < 0;
```
christopherbate, line 268: use `ShapedType::isDynamic(dimSize)`. The specific sentinel value representing dynamic shape is an implementation detail.

christopherbate (changes requested):
We can't merge with regression, are you going to look into it or do you need me do so?

pranavm-nvidia:
> We can't merge with regression, are you going to look into it or do you need me do so?

I am looking into it. I think it might be the test that's wrong - we can discuss more in today's sync.

mlir-tensorrt/tensorrt/test/Dialect/TensorRT/broadcast-elimination.mlir
```diff
@@ -310,40 +371,3 @@ func.func @broadcast_elim_matmul_vector(%arg0: tensor<?x?x128xf32>, %arg1: tenso
...
-
-// CHECK-LABEL: func.func @broadcast_dynamic_expand_shape_regression
-//  CHECK-SAME: (%[[arg0:.+]]: tensor<?x?x1x1xi1>, %[[arg1:.+]]: tensor<?x1xf16>, %[[arg2:.+]]: tensor<?x?x256x256xf16>, %[[arg3:.+]]: tensor<4xi32>) -> tensor<?x?x256x256xf16> {
-//       CHECK:     %[[v0:.+]] = tensorrt.reshape %[[arg1]] : tensor<?x1xf16> to tensor<1x1x?x1xf16>
-//       CHECK:     %[[v1:.+]] = tensorrt.select ins(%[[arg0]], %[[arg2]], %[[v0]] : tensor<?x?x1x1xi1>, tensor<?x?x256x256xf16>, tensor<1x1x?x1xf16>) -> tensor<?x?x256x256xf16>
-//       CHECK:     return %[[v1]] : tensor<?x?x256x256xf16>
-
-// -----
-
-func.func @broadcast_dynamic_expand_shape_regression(%arg0: tensor<?x?x1x1xi1>, %arg1: tensor<?x1x?xf16>, %arg2: tensor<?x?x256x256xf16>, %arg3: tensor<4xi32>) -> tensor<?x?x256x256xf16> {
```
pranavm-nvidia, line 330: I've removed these two tests because I don't think they add any coverage beyond what `broadcast_select_negative_dynamic_input` provides. If that's not the case, I can update and re-enable them.

mlir-tensorrt/tensorrt/lib/TensorRT/Transforms/BroadcastElimination.cpp
```diff
@@ -263,6 +263,11 @@ static bool broadcastAbsorbPreconditions(BroadcastOp broadcastOp) {
 }
 
 namespace {
+
+static bool isOneOrDynamic(int64_t dimSize) {
```
christopherbate, line 267: static functions go outside anonymous namespace 

https://llvm.org/docs/CodingStandards.html#restrict-visibility

mlir-tensorrt/tensorrt/test/Dialect/TensorRT/broadcast-elimination.mlir
```diff
@@ -310,40 +371,3 @@ func.func @broadcast_elim_matmul_vector(%arg0: tensor<?x?x128xf32>, %arg1: tenso
 
 
 // -----
-
-func.func @broadcast_dynamic_expand_shape_regression(%arg0: tensor<?x?x1x1xi1>, %arg1: tensor<?x1xf16>, %arg2: tensor<?x?x256x256xf16>, %arg3: tensor<4xi32>) -> tensor<?x?x256x256xf16> {
```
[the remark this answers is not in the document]
christopherbate: To clarify: the pattern being tested is just for the simplification of the broadcast via reshape/transpose. So we could simplify the test by removing the elementwise operation and chest checking the result of  the broadcast transformation.

christopherbate (approved):

PR #524, opened 2025-02-24
Port resize ops

yizhuoz004:


tripy/nvtripy/trace/ops/resize.py
```diff
@@ -34,55 +36,51 @@ class Resize(BaseTraceOp):
...
-        from nvtripy.flat_ir.ops import ResizeCubicOp, ResizeLinearOp, ResizeNearestOp
-
-        if self.scales:
-            from nvtripy.common.datatype import float32, int32
-            from nvtripy.flat_ir.ops import ConstantOp, ConvertOp, MulOp
-            from nvtripy.flat_ir.tensor import FlatIRTensor
-
-            # construct output_shape using scales
-            # inputs[1] is input[0].shape
-            # output_shape = (inputs[1].cast(fp32) * scales).cast(int32)
```
yizhuoz004, line 47: We used to always compute `output_shape` because the stablehlo pipeline needs this argument for shape inference. But now we can just pass `scales` when it's available.

tripy/nvtripy/trace/ops/resize.py
```diff
@@ -34,55 +36,51 @@ class Resize(BaseTraceOp):
...
+        assert len(inputs) == 1 or len(inputs) == 2, "Resize must have exactly 1 or 2 inputs."
+        selector_attr = tensorrt.ResizeSelectorAttr.get("kFORMULA")
+        if len(inputs) == 2:
+            output_shape = inputs[1]
+            scales_attr = None
+        else:
+            assert self.scales, "Resize scales must be provided when there is only 1 input."
+            output_shape = None
+            scales_attr = ir.DenseF32ArrayAttr.get(self.scales)
         if self.mode == "nearest":
```
pranavm-nvidia, line 49: Could we actually split these different modes into separate trace operation classes? You could have a factory function like we have for [unary](https://github.com/NVIDIA/TensorRT-Incubator/blob/trt-dialect/tripy/nvtripy/trace/ops/unary.py). 

I think it'd be good to have a 1:1 mapping of Trace ops to TRT dialect ops.

tripy/nvtripy/trace/ops/resize.py
```diff
@@ -34,55 +36,51 @@ class Resize(BaseTraceOp):
     def infer_dtypes(self):
         self.outputs[0].dtype = self.inputs[0].dtype
```
pranavm-nvidia, line 37: Can this be dropped? The default behavior is to propagate dtypes like this.

tripy/nvtripy/trace/ops/resize.py
yizhuoz004: Updated, I'm using a `ResizeBase` to handle some shared code.

tripy/nvtripy/trace/ops/resize.py
yizhuoz004: It seems when there are multiple inputs, we have to override the `infer_dtypes`: https://github.com/NVIDIA/TensorRT-Incubator/blob/main/tripy/nvtripy/trace/ops/base.py#L85

PR #531, opened 2025-02-24
Port pooling ops

yizhuoz004:


tripy/nvtripy/trace/ops/pooling.py
```diff
@@ -30,8 +32,8 @@ class Kind(enum.Enum):
         def __init__(self, op):
             self.op = op
 
-        MAX = "max"
-        AVG = "avg"
+        MAX = "kMAX"
+        AVG = "kAVERAGE"
```
pranavm-nvidia, line 36: Worth having separate ops for these IMO. Makes the trace a bit more readable and also makes the code slightly easier to reason about. A factory function similar to what I added for unary should do the trick.

tripy/nvtripy/trace/ops/pooling.py
```diff
@@ -43,93 +45,25 @@ def __init__(self, op):
     def infer_dtypes(self):
         self.outputs[0].dtype = self.inputs[0].dtype
```
pranavm-nvidia, line 46: Is this one needed? There's only 1 input right?

tripy/nvtripy/trace/ops/pooling.py
```diff
@@ -43,93 +45,25 @@ def __init__(self, op):
...
-                [reduce_out],
-                reduce_mode=self.kind.op,
-                window_dims=window_dims,
-                window_strides=window_strides,
-                padding=padding,
+    def to_mlir(self, inputs, outputs):
+        window_size_attr = ir.DenseI64ArrayAttr.get(self.kernel_dims)
+        stride_attr = ir.DenseI64ArrayAttr.get(self.stride)
+        pre_padding_attr = ir.DenseI64ArrayAttr.get([pad[0] for pad in self.padding])
+        post_padding_attr = ir.DenseI64ArrayAttr.get([pad[1] for pad in self.padding])
```
pranavm-nvidia, line 52: Maybe the caller should split up the padding? That will make the trace text representation closer to the TRT op

tripy/nvtripy/trace/ops/pooling.py
```diff
@@ -43,93 +45,25 @@ def __init__(self, op):
...
+        pooling_type_attr = tensorrt.PoolingTypeAttr.get(self.kind.op)
+        avg_excludes_padding_attr = None
+        # TODO: TensorRT defaults this value to be True, verify if this aligns with torch
```
pranavm-nvidia, line 55: I would assume excluding padding is generally the default behavior

tripy/nvtripy/trace/ops/pooling.py
yizhuoz004: Updated, do we also want to seperate the frontend APIs into different files?

tripy/nvtripy/trace/ops/pooling.py
yizhuoz004: Seems the other case in torch: https://pytorch.org/docs/stable/generated/torch.nn.AvgPool2d.html
`count_include_pad` is True by default.

tripy/nvtripy/trace/ops/pooling.py
yizhuoz004: I want to keep the padding representation consistent with torch, and also in other tripy ops (conv, pad etc).

tripy/nvtripy/trace/ops/pooling.py
yizhuoz004: I'm going to add this option to the frontend, question: shall we go with torch or trt convension?

tripy/nvtripy/trace/ops/pooling.py
pranavm-nvidia: Yeah, I've been doing that so far. Makes it a bit easier to find and prevents files from getting too long.

tripy/nvtripy/trace/ops/pooling.py
pranavm-nvidia: Right, but the translation can be done within the `tp.pooling` API right? I think we want the trace parameters to be as close to TRT parameters as possible

tripy/nvtripy/trace/ops/pooling.py
pranavm-nvidia: TRT and ONNX both default it to off. I would assume that most times you wouldn't want to skew the averages with padding values? OTOH if we assume most people are porting models from torch, then this discrepancy would be a source of bugs. Let's go with the torch convention for now?

tripy/nvtripy/trace/ops/pooling.py
yizhuoz004: Updated, thanks

tripy/nvtripy/frontend/ops/pooling/avgpool.py
```diff
@@ -115,6 +58,8 @@ def avgpool(
             to apply to the input along each spatial dimension before and after the dimension respectively,
             where :math:`M` is the number of spatial dimensions, i.e. :math:`M = \text{rank(input)} - 2`.
             Defaults to all 0.
+        count_include_pad: When True, the padded values will be included in the averaging operation. This
+            option can only be ``True`` when input datatype is :class:`int8`.
```
pranavm-nvidia, line 62: nit:
```suggestion
            option must be ``True`` if the input data type is :class:`int8`.
```
The current wording sounds like only `False` is allowed if the dtype is not int8

tripy/nvtripy/frontend/ops/pooling/avgpool.py
```diff
@@ -115,6 +58,8 @@ def avgpool(
             to apply to the input along each spatial dimension before and after the dimension respectively,
             where :math:`M` is the number of spatial dimensions, i.e. :math:`M = \text{rank(input)} - 2`.
             Defaults to all 0.
+        count_include_pad: When True, the padded values will be included in the averaging operation. This
```
pranavm-nvidia, line 61:
```suggestion
        count_include_pad: Whether to include padded values in the averaging operation. This
```

tripy/nvtripy/frontend/ops/pooling/maxpool.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    spatial_dims = len(kernel_dims)
+    if spatial_dims != 2 and spatial_dims != 3:
+        raise_error("Unsupported kernel_dims, must be 2D or 3D.", [f"Got kernel_dims={kernel_dims}"])
+
+    op_utils.check_conv_pooling_args(kernel_dims, stride, padding)
+    stride = utils.utils.default(stride, [1] * spatial_dims)
+    padding = utils.utils.default(padding, [(0, 0)] * spatial_dims)
+    pre_padding = [pad[0] for pad in padding]
+    post_padding = [pad[1] for pad in padding]
```
pranavm-nvidia, line 82: Maybe worth factoring this out into a helper like `transform_pooling_params`?

tripy/nvtripy/trace/ops/pooling.py
```diff
@@ -15,121 +15,54 @@
...
+            import nvtripy.common.datatype as tp_dtype
+
+            window_size_attr = ir.DenseI64ArrayAttr.get(self.kernel_dims)
+            stride_attr = ir.DenseI64ArrayAttr.get(self.stride)
+            pre_padding_attr = ir.DenseI64ArrayAttr.get(self.pre_padding)
+            post_padding_attr = ir.DenseI64ArrayAttr.get(self.post_padding)
+            pooling_type_attr = tensorrt.PoolingTypeAttr.get(pooling_type_name)
+            avg_excludes_padding_attr = None
+            if pooling_type_name != "kMAX":
+                avg_excludes_padding_attr = ir.BoolAttr.get(self.avg_excludes_padding)
```
pranavm-nvidia, line 49: Can we not set this unconditionally? Does it fail in the bindings?

yizhuoz004: Yes, see https://github.com/NVIDIA/TensorRT-Incubator/blob/main/mlir-tensorrt/tensorrt/lib/TensorRT/IR/Verification.cpp#L529

`avg_count_excludes_padding` only works with `kAVERAGE`

tripy/nvtripy/frontend/ops/pooling/avgpool.py
```diff
@@ -115,6 +58,8 @@ def avgpool(
             to apply to the input along each spatial dimension before and after the dimension respectively,
             where :math:`M` is the number of spatial dimensions, i.e. :math:`M = \text{rank(input)} - 2`.
             Defaults to all 0.
+        count_include_pad: Whether to include padded values in the averaging operation. This
+            option must be ``True`` if the input data type is :class:`int8`.
```
pranavm-nvidia, line 62: We can actually shorten this further:
```suggestion
        count_include_pad: Whether to include padded values in the average. 
            Must be ``True`` if the input data type is :class:`int8`.
```

tripy/nvtripy/frontend/ops/pooling/maxpool.py
yizhuoz004: Updated, the dimension checks have been performed previously in `op_utils.check_conv_pooling_args`

tripy/nvtripy/frontend/ops/pooling/avgpool.py
```diff
@@ -135,7 +80,17 @@ def avgpool(
         raise_error("Unsupported kernel_dims, must be 2D or 3D.", [f"Got kernel_dims={kernel_dims}"])
 
     op_utils.check_conv_pooling_args(kernel_dims, stride, padding)
-    stride = utils.utils.default(stride, [1] * spatial_dims)
-    padding = utils.utils.default(padding, [(0, 0)] * spatial_dims)
-
-    return op_utils.create_op(Pooling, [input], Pooling.Kind.AVG, kernel_dims, stride, padding)
+    if input.dtype == datatype.int8 and not count_include_pad:
+        raise_error("`count_include_pad` can only be `True` when input datatype is int8.")
```
pranavm-nvidia, line 84:
```suggestion
        raise_error("`count_include_pad` must be `True` when input datatype is int8.")
```

tripy/nvtripy/frontend/ops/pooling/maxpool.py
```diff
@@ -0,0 +1,80 @@
...
+        output = tp.maxpool(input, kernel_dims=(2, 2))
+
+        pool_torch = torch.nn.MaxPool2d((2, 2), stride=1) # doc: omit
+        expected = pool_torch(torch.from_dlpack(input).to("cpu")) # doc: omit
+
+        assert torch.allclose(torch.from_dlpack(output).to("cpu"), expected)
+    """
+    spatial_dims = len(kernel_dims)
+    if spatial_dims != 2 and spatial_dims != 3:
+        raise_error("Unsupported kernel_dims, must be 2D or 3D.", [f"Got kernel_dims={kernel_dims}"])
```
pranavm-nvidia, line 76: I think this check can also be pulled into `transform_pooling_params`? Might want to put it under `fronted/ops/pooling/utils.py` so it's better scoped and easier to find

pranavm-nvidia (approved):

PR #537, opened 2025-02-25
[TensorRT] Fix dynamic linspace translation, relax dynamic verifier

yizhuoz004:
Fixes dynamic linspace translation by passing a 1-D static shape tensor. Relaxes linspace op's verifier when the input step dimension is unknown.

### DIFF
--- a/mlir-tensorrt/tensorrt/lib/Target/TensorRTEncodingOpInterface/NetworkEncoder.cpp
+++ b/mlir-tensorrt/tensorrt/lib/Target/TensorRTEncodingOpInterface/NetworkEncoder.cpp
@@ -371,9 +371,8 @@ nvinfer1::ILayer *NvInferNetworkEncoder::addFillLayer(
     nvinfer1::Dims shapeDims = dynamicShape->getDimensions();
     assert(shapeDims.nbDims == 1 && shapeDims.d[0] > 0 &&
            "invalid shape tensor shape");
-    staticShape.nbDims = shapeDims.d[0];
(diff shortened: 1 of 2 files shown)


mlir-tensorrt/tensorrt/lib/Target/TensorRTEncodingOpInterface/NetworkEncoder.cpp
```diff
@@ -371,7 +371,7 @@ nvinfer1::ILayer *NvInferNetworkEncoder::addFillLayer(
     nvinfer1::Dims shapeDims = dynamicShape->getDimensions();
     assert(shapeDims.nbDims == 1 && shapeDims.d[0] > 0 &&
            "invalid shape tensor shape");
-    staticShape.nbDims = shapeDims.d[0];
+    staticShape.nbDims = shapeDims.nbDims;
```
pranavm-nvidia, line 374: This happens to work because shape tensors will always have rank 1, but IMO we should just hard code it. From the documentation:
```
For FillOperation::kLINSPACE, dimensions.nbDims must be 1 for static start/delta. 
```

yizhuoz004: Sure, updated

christopherbate:
@yizhuoz004 @shelkesagar29 This change was missing a corresponding test in `tensorrt/tensorrt/test/Dialect/TensorRT/invalid.mlir`. Please follow up with a test.

PR #546, opened 2025-02-26
Port Pad, add unary neg, remove Pad in Trace

yizhuoz004:


tripy/nvtripy/frontend/ops/pad.py
```diff
@@ -72,13 +75,22 @@ def pad(
             [f"Got mode={mode}, while supported modes are {supported_modes}"],
         )
 
-    padding_low, padding_high = list(zip(*pad))
+    padding_lows, padding_highs = list(zip(*pad))
+    padding_lows = op_utils.tensor_from_shape_like(padding_lows)
+    padding_highs = op_utils.tensor_from_shape_like(padding_highs)
+    starts = -padding_lows
+    input_shape = op_utils.create_op(Shape, [input])
```
pranavm-nvidia, line 82: We can just use `input.shape` here

yizhuoz004: `input.shape` returns a list of `DimensionSize`, I just want a single `Tensor` here

pranavm-nvidia: Ah got it. Could you add a comment so we don't accidentally change it in the future?

tripy/nvtripy/trace/ops/slice.py
```diff
@@ -24,10 +24,28 @@
 
 @dataclass(repr=False)
 class Slice(BaseTraceOp):
+
+    mode: str = None
```
pranavm-nvidia, line 28: Can we make separate trace ops for different modes similar to what we have for unary, binary, etc.? The mode changes the behavior significantly

yizhuoz004: Updated

PR #555, opened 2025-03-03
Updates TensorRT outlining to propagate shape profile attributes

pranavm-nvidia:


mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp
```diff
@@ -118,7 +132,7 @@ outlineOp(RewriterBase &rewriter, tensorrt::TensorRTModuleOp trtModule,
       makeRegionIsolatedFromAbove(rewriter, inlineGroupOp.getRegion());
 
   // Create the outlined function
-  FailureOr<FunctionOpInterface> func = createOutlinedFunc(
+  FailureOr<FunctionOpInterface> func = createOutlinedFunc(outerModule,
```
christopherbate, line 135: The function which contains the cluster can be found via `cluster.back().getParentOfType<func::FuncOp>()`.

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp
```diff
@@ -63,15 +63,29 @@ createOutlinedFunc(RewriterBase &rewriter, Location loc, Operation *module,
...
+  StringRef tensorrtShapeBoundsAttrName = mlir::tensorrt::TensorRTDialect::getShapeProfileArgAttrName();
+
+  // TODO (pranavm): Will there only ever be one outer FuncOp? How do we make this more robust?
+  func::FuncOp outerFunc;
+  for (auto func : outerModule.getOps<func::FuncOp>()) {
+      outerFunc = func;
+      break;
+  }
+  for (unsigned idx = 0; idx < outlinedFunc.getNumArguments(); idx++) {
+    const auto profile = outerFunc.getArgAttrOfType<tensorrt::ShapeProfileAttr>(idx, tensorrtShapeBoundsAttrName);
```
christopherbate, line 76: The arguments of the outlined function may not be the same or ordered the same as the function which contained the cluster.

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp
```diff
@@ -54,7 +54,7 @@ static FailureOr<ClusteringOpts> getTensorRTClusteringOptions(Operation *op) {
 /// `nameBase` but may have numbers appended in order to unique the name. The
 /// created function has argument/result types as indicated by the parameters.
 static FailureOr<FunctionOpInterface>
-createOutlinedFunc(RewriterBase &rewriter, Location loc, Operation *module,
+createOutlinedFunc(ModuleOp outerModule, RewriterBase &rewriter, Location loc, Operation *mod,
```
christopherbate, line 57: I would suggest to leave this function (`createOutlinedFunc`) unchanged.

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp
```diff
@@ -118,7 +132,7 @@ outlineOp(RewriterBase &rewriter, tensorrt::TensorRTModuleOp trtModule,
       makeRegionIsolatedFromAbove(rewriter, inlineGroupOp.getRegion());
```
christopherbate, line 132: You'll want to look at these `inputs` to figure out which of them:

a) has a dynamic shape (since the shape profile only matters for dynamic shape)
b) if it has dynamic shape, you'll want to check whether it is a block argument of the containing function.

So it would look something like this:


```
SmallVector<Attribute> profileAttrsPerInput;
for(Value v : inputs) {
  auto rtt = dyn_cast<RankedTensorType>(v.getType());
  if(!rtt || rtt.hasStaticShape()) {
   profileAttrsPerInput.push_back(Attribute{});
   continue;
  }

  auto blockArg = dyn_cast<BlockArgument>(v);
  if(!blockArg || blockArg.getOwner()->getParentOp() != funcContainingCluster) {
    // emit an error here because in this case , it is impossible to deduce the shape profiel without further analysis
  }
  
  int64_t funcArgNum = blockArg.getArgNumber();
  profileAttrsPerInput.push_back(
    funcContainingCluster.getArgAttrOfType<tensorrt::ShapeProfileAttr>(funcArgNum)
  );

  // Check `profileAttrsPerInput.back()` to see if it is non-null. If it is null, emit an error.
  
}

// use profileAttrsPerInput to populate attributes on the outlined func.
```

```

Comment (pranavm-nvidia) [reply mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp]: Updated accordingly.

File: mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp [hunk 5]
```diff
@@ -124,6 +124,40 @@ outlineOp(RewriterBase &rewriter, tensorrt::TensorRTModuleOp trtModule,
...
+        funcContainingCluster.getArgAttrOfType<tensorrt::ShapeProfileAttr>(
+            argIndex, tensorrtShapeBoundsAttrName));
+
+    if (!profileAttrsPerInput.back())
+      emitError(trtModule.getLoc())
+          << "Profile attribute of argument " << argIndex << " is not set";
+  }
+
+  for (unsigned idx = 0; idx < func->getNumArguments(); idx++) {
+    func->setArgAttr(idx, tensorrtShapeBoundsAttrName,
```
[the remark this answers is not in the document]
pranavm-nvidia: Note to self: I should be able to reproduce this in the unit test by making the input static and output dynamic and not providing shape profile attributes.

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp
```diff
@@ -124,6 +124,42 @@ outlineOp(RewriterBase &rewriter, tensorrt::TensorRTModuleOp trtModule,
...
+          << "Block argument is not part of the signature of the function "
+             "containing this TRT cluster";
+
+    int64_t argIndex = blockArg.getArgNumber();
+    profileAttrsPerInput.push_back(
+        funcContainingCluster.getArgAttrOfType<tensorrt::ShapeProfileAttr>(
+            argIndex, tensorrtShapeBoundsAttrName));
+
+    if (!profileAttrsPerInput.back())
+      emitError(trtModule.getLoc())
```
christopherbate, line 152: You're emitting an error here but not communicating that the pass should fail (if an error is emitted, then the pass should fail).

You can do `return emitError(...);` which is equivalent to  `emitError(...); return failure();`. 

Then the failure needs to be passed backwards to the `runOnOperation` function in the pass, where you can then do `signalPassFailure();`

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp
```diff
@@ -124,6 +124,42 @@ outlineOp(RewriterBase &rewriter, tensorrt::TensorRTModuleOp trtModule,
...
+    auto rtt = dyn_cast<RankedTensorType>(v.getType());
+    if (!rtt || rtt.hasStaticShape()) {
+      profileAttrsPerInput.push_back(Attribute{});
+      continue;
+    }
+
+    auto blockArg = dyn_cast<BlockArgument>(v);
+    if (!blockArg ||
+        blockArg.getOwner()->getParentOp() != funcContainingCluster)
+      emitError(trtModule.getLoc())
```
christopherbate, line 142: `return emitError(...) ...;`

mlir-tensorrt/compiler/lib/Compiler/TensorRTToExecutable/Passes.cpp
christopherbate: Might also want to give a more descriptive error message (these are visible to compiler user). For example, include the key of the attribute that is expected (`tensorrtShapeBoundsAttrName`)

mlir-tensorrt/compiler/test/Pipelines/TensorRTClustering/tensorrt-clustering.mlir
```diff
@@ -1,15 +1,28 @@
...
 
 // CHECK-LABEL: @trt_relu
 // CHECK-DAG:   %[[v0:.+]] = tensorrt.call_alloc @trt_engines::@tensorrt_cluster
 // CHECK-DAG: return %[[v0]]
-// CHECK-DAG: @tensorrt_cluster
+// CHECK-DAG: @tensorrt_cluster(%arg0: tensor<?xf16> {tensorrt.shape_profile = #tensorrt.shape_profile<min = [1], opt = [4], max = [7]>})
 // CHECK-DAG: %[[v1:.+]] = tensorrt.activation {activationType = #tensorrt.activation_type<kRELU>}
 // CHECK-DAG: return %[[v1]]
+
+// -----
```
christopherbate, line 17: To head off any future problems, you may want to add a more complicated test case here where the function has multiple arguments. Some variations might be: N arguments where a couple are unused, have the arguments be used in an order other than their list order (to try to create a clustering where the args to the cluster have a different order). 

You can also create a negative test case. Add `-verify-diagnostics` to the RUN command, then you can have a test like


```
// expected-error @below {{expected error message}}
func.func @no_shape_profile(%arg0: tensor<?xf32>) -> tensor<?xf32> {
   // body just like `trt_relu` test
   ...
 }
```

christopherbate (approved):

PR #556, opened 2025-03-04
[NFC] Add a negative unit test for dynamic tensorrt.linspace

yizhuoz004:
Adds a unit test to check if error is reported when step dim[0] is not equal to rank of output in tensorrt.linspace.


christopherbate (approved):
LGTM, thanks!

PR #558, opened 2025-03-05
Adds layer checks to prevent nullptr dereferences

pranavm-nvidia:
When a TensorRT `add<Layer>` function fails (e.g. due to invalid parameters), it will return a `nullptr`. Previously, we were not checking for this, which would cause segfaults later when we tried to dereference the `nullptr`.


mlir-tensorrt/tensorrt/include/mlir-tensorrt-dialect/TensorRT/IR/TensorRTOps.td
```diff
@@ -224,6 +224,8 @@ def TensorRT_ActivationOp : TensorRT_Op<"activation", [Pure,
     nvinfer1::IActivationLayer *layer = $net->addActivation(
                                                     *$input,
                                                     *$activationType);
+    if (!layer)
```
pranavm-nvidia, line 227: Leaving a comment open so I can also update `cast` once I merge that (I don't want to edit that PR since it's already passed the pipeline/code review).

christopherbate (approved):
LGTM, thanks

PR #560, opened 2025-03-06
Allows I64 tensors in `tensorrt.shape`

pranavm-nvidia:
Updates `tensorrt.shape` to allow accepting I64 input tensors. The shape operation doesn't really care about the data type of its input (only shape) and so all data types are supported.

### DIFF
--- a/mlir-tensorrt/tensorrt/test/Target/TensorRT/shape.mlir
+++ b/mlir-tensorrt/tensorrt/test/Target/TensorRT/shape.mlir
@@ -12,6 +12,16 @@ func.func @tensorrt_shape_op(
 }
 
 
+// CHECK-LABEL: @tensorrt_shape_op_int64
+//  CHECK-SAME: tensorrt.engine
+func.func @tensorrt_shape_op_int64(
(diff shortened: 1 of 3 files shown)


shelkesagar29 (approved):
LGTM. Thanks!

mlir-tensorrt/tensorrt/test/Target/TensorRT/shape.mlir
```diff
@@ -12,6 +12,16 @@ func.func @tensorrt_shape_op(
 }
 
 
+// CHECK-LABEL: @tensorrt_shape_op_int64
+//  CHECK-SAME: tensorrt.engine
+func.func @tensorrt_shape_op_int64(
```
christopherbate, line 17: I think this specific test would need to be predicated using "// REQUIRES: tensorrt-version-ge-10.0" in a separate file if one doesn't already exist for this op.

pranavm-nvidia: Makes sense, I have moved the test: https://github.com/NVIDIA/TensorRT-Incubator/pull/563

PR #561, opened 2025-03-06
Corrects resize validation for dynamic shapes

pranavm-nvidia:
When the output dimensions are dynamic and the `output_shape` parameter has not been provided, resize requires `scales` to be present. The previous checks were erroneously reporting errors if the input dimensions were dynamic, even though this is actually supported by TensorRT.

Also cleans up several error messages to be clearer.


christopherbate (approved):
LGTM, thanks

PR #562, opened 2025-03-06
Allows scalars in unary operations

pranavm-nvidia:
Current versions of TensorRT allow unary inputs to be scalar. This change restricts the 1D check to TensorRT 8 so that newer versions are able to accept scalar inputs.


christopherbate (approved):
LGTM, thanks!

PR #564, opened 2025-03-11
Enables BF16 in pooling

pranavm-nvidia:
The pooling layer in TRT supports all activation types, which includes BF16. For now, this change is restricted to TRT 10+, but likely would apply to TRT 9 as well.


shelkesagar29:
@pranavm-nvidia if you have time, I would appreciate checking BF16 uses for other layers too. So we can add all changes in one PR.

christopherbate (approved):

PR #566, opened 2025-03-11
Enable QDQ related tests, enforce scale to be constant

yizhuoz004:


tripy/nvtripy/frontend/ops/utils.py
```diff
@@ -230,6 +232,9 @@ def check_qdq_args(input, scale, dtype, dim, is_quantize):
             [f"scale has rank={scale.rank}."],
         )
 
+    if not isinstance(scale.trace_tensor.producer, Constant):
+        raise_error("Scale must be a constant, please evaluate scale tensor first.")
```
pranavm-nvidia, line 236: Could we mention `.eval()` specifically here? Also would be good to mention where the scale tensor came from:
```suggestion
        raise_error("Scale must be a constant. Please evaluate the tensor using `.eval()`.", [f"Note: scale was defined here:", scale])
```

pranavm-nvidia (approved):

PR #568, opened 2025-03-12
Updates default TensorRT version to 10.9

pranavm-nvidia:
Updates the default version we compile against to 10.9. Also corrects the build number to match the latest release.


christopherbate (commented):
Ci workflows need to be updated

christopherbate (approved):

pranavm-nvidia:
> Ci workflows need to be updated

Done

christopherbate:
Looks like some of the CMake changes need to be ported over from internal. Rebasing on `main` might fix, would need to check

PR #572, opened 2025-03-13
Document eval, tolist

yizhuoz004:


tripy/nvtripy/frontend/tensor.py
```diff
@@ -223,11 +235,26 @@ def eval(self) -> runtime.MemRefValue:
...
+
+        Returns:
+            The tensor represented as a nested list or a python number.
+
+        .. code-block:: python
+        :linenos:
+        :caption: eval
+
+            tensor = tp.ones((2, 2))
+            tensor_list = tensor.tolist()
```
pranavm-nvidia, line 252: We'll need to add a `# doc: print-locals tensor_list` (and also include `tensor` if you want that as well).

tripy/nvtripy/frontend/tensor.py
```diff
@@ -177,12 +176,24 @@ def ndim(self):
     def device(self):
         return self.trace_tensor.device
 
-    # TODO (pranavm): Document this and `tolist()`? Figure out more ergonomic way to get scalar value.
-    def eval(self) -> runtime.MemRefValue:
+    def eval(self) -> "nvtripy.Tensor":
+        """
+        Evaluates the tensor, making the tensor to be a constant.
```
pranavm-nvidia, line 181: Maybe something like:
```
Immediately evaluates this tensor. By default, tensors are evaluated lazily. 
```
There's not really a notion of a "constant" in the user API, e.g. `tp.ones(...)` looks like a constant but isn't actually one.

tripy/nvtripy/frontend/tensor.py
```diff
@@ -223,11 +235,26 @@ def eval(self) -> runtime.MemRefValue:
                 f"Note: Tensor was evaluated while compiling here: {str_from_stack_info(self.trace_tensor.eval_stack_info)}",
                 mode="once",
             )
+        return self
 
-        return data
+    def tolist(self) -> Union[List, numbers.Number]:
+        """
+        Returns the tensor as a nested list. If the tensor is a scalar, returns a python number.
```
pranavm-nvidia, line 242: Does it make sense to also have an `item()` API for scalars? On one hand, it's redundant with `tolist`, but on the other it is more clear.

PR #576, opened 2025-03-13
Improves slice error messages for unsupported types

pranavm-nvidia:


tripy/nvtripy/frontend/ops/slice.py
```diff
@@ -105,6 +106,17 @@ def to_positive_idx(slice_idx):
         else:
             assert isinstance(slice_idx, slice)
 
+            def check_type(name, arg):
```
yizhuoz004, line 109: Should we also remove tp.Tensor from the type annotation:
`index: Union["nvtripy.Tensor", slice, IntLike, Sequence[Union[slice, IntLike]]]` ?

pranavm-nvidia: @yizhuoz004 Tensors can be used in slice to perform a gather, i.e. `a[t]` - it's just that they can't be used for slice _bounds_, i.e. `a[0:t]` etc.

yizhuoz004 (approved):

PR #578, opened 2025-03-14
Add support for TensorRT Quickly Deployable Plugins

samurdhikaru:
- Overloads `tp.plugin` to allow the addition of TRT QDPs
- Relaxes the check for the defined plugin field length vs provided field length match
   - Many legacy plugins and QDPs are defined with variable length plugin fields where `getFieldNames()` of the plugin creator could report zero field length. So the check is enforced only when defined field length is positive.
- Adds two integration tests for QDPs


tripy/nvtripy/frontend/ops/plugin.py
```diff
@@ -75,3 +76,122 @@ def plugin(
         output_info,
         kwargs,
     )
+
+@export.public_api(document_under="operations/functions")
+def plugin(
+    op: trtp._lib.PluginDef,
```
pranavm-nvidia, line 82: Would it be possible to use the same string that we register the plugin with, i.e. `"example::elemwise_add_plugin"`? We can use `getattr` internally to grab the correct plugin from TRT. This way we keep TRT types out of the Tripy interface.

tripy/tests/integration/test_plugin_qdp.py
```diff
@@ -0,0 +1,184 @@
...
+            pads = pads,
+        )
+
+        ref = np.pad(x, [[0, 0], [0, 0], [pads[0], pads[1]], [pads[2], pads[3]]], "wrap")
+
+        assert cp.allclose(cp.from_dlpack(out), ref)
+
+if __name__ == "__main__":
+    test_elemwise_add_plugin()
+    test_circ_pad_plugin()
```
pranavm-nvidia, line 184: Can probably remove this part. You can do this via `pytest <path> -k <test_name>`

tripy/tests/integration/test_plugin_qdp.py
```diff
@@ -0,0 +1,184 @@
...
+        inp = tp.iota((2, 2))
+        out = tp.plugin(
+            trt.plugin.op.example.elemwise_add_plugin,
+            [inp],
+            block_size = 256,
+        )
+
+        assert cp.allclose(cp.from_dlpack(out), cp.from_dlpack(inp + 1))
+
+    def test_circ_pad_plugin():
```
pranavm-nvidia, line 165:
```suggestion
    def test_circ_pad_plugin(self):
```

tripy/docs/post0_developer_guides/integrating-tensorrt-python-plugins.md
```diff
@@ -0,0 +1,125 @@
...
+
+@trtp.register("example::elemwise_add_plugin")
+def add_plugin_desc(inp0: trtp.TensorDesc, block_size: int) -> trtp.TensorDesc:
+    """
+    Defines the plugin interface - inputs, outputs and attributes.
+    Args:
+    inp0: Input tensor descriptor
+    block_size: Block size for the Triton kernel
+    Returns:
+    Output tensor descriptor with same shape/dtype as input
```
pranavm-nvidia, line 39:
```suggestion
        inp0: Input tensor descriptor
        block_size: Block size for the Triton kernel
    Returns:
        Output tensor descriptor with same shape/dtype as input
```

tripy/docs/post0_developer_guides/integrating-tensorrt-python-plugins.md
```diff
@@ -0,0 +1,125 @@
...
+)
+
+assert cp.allclose(cp.from_dlpack(out), cp.from_dlpack(inp + 1))
+```
+
+### Supported attribute types
+
+QDPs support either scalar attributes of selected primitive Python types or 1-D Numpy arrays of selected Numpy data types. For Numpy array attributes, the corresponding argument in `@trtp.register` must be annotated with `numpy.typing.NDArray[dtype]`, where `dtype` is the expected Numpy data type. Please refer to the [TensorRT Plugin Guide](https://docs.nvidia.com/deeplearning/tensorrt/api/python_api/pluginGuide.html) for details on exact primitive and Numpy data types supported by QDPs.
+
+## Understanding the Implementation
```
pranavm-nvidia, line 101: I think this section can be converted to inline comments so we don't need to jump around to understand what's going on?

tripy/docs/post0_developer_guides/integrating-tensorrt-python-plugins.md
```diff
@@ -0,0 +1,125 @@
...
+The `@trtp.aot_impl` decorator returns an ahead-of-time (AOT) compiled kernel, along with information required to invoke the kernel such as kernel launch parameters and extra scalar arguments. This plugin uses OpenAI Triton to define the kernel and uses its AOT compilation capabilities to retrieve PTX for the kernel.
+
+### Tripy Integration
+
+Tripy's `tp.plugin()` function accepts:
+- The plugin op
+- A list of input tensors
+- Plugin attributes as keyword arguments
+
+Links:
```
pranavm-nvidia, line 123: Can put these under a seealso block:
```
:::{seealso}
...
:::
```

tripy/docs/post0_developer_guides/integrating-tensorrt-python-plugins.md
```diff
@@ -0,0 +1,125 @@
+# Integrating TensorRT Python Plugins with Tripy
```
pranavm-nvidia, line 1: I think this should be moved to `pre0_user_guides`. The idea is that user guides are meant for end users of Tripy while developer guides are meant for people contributing to Tripy.

tripy/nvtripy/flat_ir/ops/plugin.py
```diff
@@ -95,7 +96,10 @@ def plugin_field_to_attr(field_info: "compiler.PluginFieldInfo", values: Any) ->
         attrs = [ir.IntegerAttr.get(INT_TYPES[field_info.type](), value) for value in values]
 
     if len(attrs) != field_info.length:
-        return Result.err([f"Expected {field_info.length} value(s), but got {len(attrs)}."])
+        if field_info.length > 0:
+            return Result.err([f"Expected {field_info.length} value(s), but got {len(attrs)}."])
+        else:
+            logger.verbose(f"Expected {field_info.length} value(s), but got {len(attrs)}. May be ignored if using TensorRT quickly deployable plugins (QDP) or if it is variable length field.")
```
pranavm-nvidia, line 102: What does `field_info.length == 0` mean in the normal case? Would a plugin ever have a list field that expects no entries? If not, we can probably just drop the warning here

pranavm-nvidia: In that case the only change needed here is:
```py
if field_info.length > 0 and len(attrs) != field_info.length:
```

samurdhikaru: It is a legal thing to do, although realistic use cases are hard to imagine. Might be better to just drop the warning if it is indeed rare.

tripy/docs/post0_developer_guides/integrating-tensorrt-python-plugins.md
samurdhikaru: Moved to `pre0_user_guides`, thanks

tripy/docs/post0_developer_guides/integrating-tensorrt-python-plugins.md
samurdhikaru: Is `seealso` supported in Markdown?

tripy/docs/post0_developer_guides/integrating-tensorrt-python-plugins.md
samurdhikaru: Good point, revised

tripy/docs/post0_developer_guides/integrating-tensorrt-python-plugins.md
pranavm-nvidia: Not sure. We use the Myst parser to parse these guides before they are finally rendered, so our documentation supports a bunch of extra markup compared to Markdow.

samurdhikaru: Looks like it is supported: https://myst-parser.readthedocs.io/en/v0.17.0/_sources/api/parsers.md.txt

tripy/nvtripy/frontend/ops/plugin.py
```diff
@@ -75,3 +76,129 @@ def plugin(
...
+        assert cp.allclose(cp.from_dlpack(out), cp.from_dlpack(inp + 1))
+    """
+
+
+    namespace, name = op.split("::")
+
+    try:
+        trtp_op = getattr(getattr(trtp.op, namespace), name)
+    except AttributeError:
+        raise ValueError(f"Plugin {op} not found. Expected a plugin of the form '<namespace>::<name>'.")
```
pranavm-nvidia, line 163: We should use `raise_error` from `nvtripy.common.exception` here and below

tripy/nvtripy/flat_ir/ops/plugin.py
samurdhikaru: Dropped the warning

tripy/nvtripy/frontend/ops/plugin.py
```diff
@@ -17,7 +17,10 @@
 
 from typing import List, Sequence, Tuple, Union
 
+import numpy as np
```
pranavm-nvidia, line 20: Any way we can remove this NumPy dependency? Looks like we convert all NumPy array to lists anyways - could we just accept lists instead?

tripy/nvtripy/frontend/ops/plugin.py
```diff
@@ -75,3 +78,125 @@ def plugin(
...
+        if key not in trtp_op.input_attrs:
+            raise_error(f"Unexpected attribute {key} provided. Expected one of {trtp_op.input_attrs.keys()}.")
+
+        attr_annotation = trtp_op.input_attrs[key]
+
+        if isinstance(value, np.ndarray):
+            kwargs[key] = value.tolist()
+        else:
+            if attr_annotation is not type(value):
+                raise_error(f"Unexpected type '{type(value)}' for attribute '{key}'. Expected '{attr_annotation}'.")
```
pranavm-nvidia, line 178: Can use `type_str_from_arg` and `str_from_type_annotation` here to make sure the display of these types is clean.

tripy/nvtripy/frontend/ops/plugin.py
samurdhikaru: No, unfortunately it does not seem possible. The `output_descs = trtp_op.register_func(*input_descs, attrs)` call below has to conform to the QDP API: the `attrs` must have a NumPy array for any non-scalar attribute. So even if we accepted Python lists through the Tripy API, we'd have to convert it to a NumPy array before the `trtp_op.register_func()`.

pranavm-nvidia: Is it feasible to extend the TRT API to accept lists? That seems like a more natural representation than allowing NumPy arrays but constraining them to 1D.

samurdhikaru: The 1D restriction is temporary -- eventually we plan to support N-D NumPy arrays. The reason to accept NumPy was so that at that point we can make a smooth transition by not having to change the API.

tripy/nvtripy/frontend/ops/plugin.py
samurdhikaru: Done

pranavm-nvidia (approved):

PR #597, opened 2025-04-01
[API] Add `get_data_segments` API to Executable

yizhuoz004:
Adds 3 APIs related to data segment so that we can extract tensorrt engine from Executable.


mlir-tensorrt/executor/include/mlir-executor/Runtime/API/API.h
```diff
@@ -436,6 +436,8 @@ class ExecutableView {
   /// name exists.
   StatusOr<FunctionView> getFunction(std::string_view name) const;
 
+  size_t getNumDataSegments() const { return view->data_segments()->size(); }
```
pranavm-nvidia, line 439: Is this equivalent to `getConstants().size()?`

mlir-tensorrt/python/bindings/Utils.h
```diff
@@ -475,6 +475,39 @@ static inline void populateExecutableBindingInModule(py::module &m) {
...
+      "get_data_segment_name",
+      [](PyExecutable &self, int64_t index) -> std::optional<std::string> {
+        MTRT_StringView name;
+        MTRT_Status s = mtrtExecutableGetDataSegmentName(self, index, &name);
+        THROW_IF_MTRT_ERROR(s);
+        if (name.length == 0)
+          return std::nullopt;
+        return std::string(name.data, name.length);
+      },
+      "returns the name of the data segment");
```
pranavm-nvidia, line 510: Maybe we should bind this to a single function which return s a `List[Tuple[str, py::bytes]]`?

mlir-tensorrt/executor/include/mlir-executor/Runtime/API/API.h
yizhuoz004: Yes I think so, `getConstants` returns a vector of `DataSegmentInfo`.

mlir-tensorrt/python/bindings/Utils.h
yizhuoz004: Good idea, combined all 3 to a single API `get_data_segments`

mlir-tensorrt/executor/include/mlir-executor/Runtime/API/API.h
pranavm-nvidia: I'm wondering then whether we need to expose this separate API?

yizhuoz004: It's more straightforward and efficient than `getConstants`, see the implementation:
```
llvm::SmallVector<DataSegmentInfo> ExecutableView::getConstants() const {
  llvm::SmallVector<DataSegmentInfo> views;
  views.reserve(view->data_segments()->size());
  for (unsigned i = 0; i < view->data_segments()->size(); i++)
    views.push_back(view->data_segments()->Get(i));
  return views;
}
```

mlir-tensorrt/executor/lib/CAPI/Common/Common.cpp
```diff
@@ -129,6 +129,40 @@ MTRT_Status mtrtExecutableGetStorageView(MTRT_Executable executable,
...
 }
 
+MTRT_Status mtrtExecutableGetNumDataSegments(MTRT_Executable executable,
+                                             int64_t *result) {
+  const Executable *exe = unwrap(executable);
+  *result = exe->getNumDataSegments();
+  return mtrtStatusGetOk();
+}
+
+MTRT_Status mtrtExecutableGetDataSegment(MTRT_Executable executable,
```
pranavm-nvidia, line 139: Maybe this method should return both the data segment and its name?

mlir-tensorrt/executor/include/mlir-executor/Runtime/API/API.h
christopherbate: Yeah we have to do that currently since it's not obvious that the flatbuffer array data layout matches `ArrayRef<DataSegmentInfo>` or not. Since the construction of the vector is not performance bearing, we haven't optimized it. In the future it could be turned into returning an iterator over the poitners rather than a new vector over the pointers.

christopherbate, line 439: Just to be  cautious here, I would put `if(!view || !view->data_segments()) return 0;` Technically, all fields are optional. If no `data_segment` was in the executable, the `data_segments()` may return nullptr instead of an empty  array.

mlir-tensorrt/executor/lib/CAPI/Common/Common.cpp
```diff
@@ -129,6 +129,30 @@ MTRT_Status mtrtExecutableGetStorageView(MTRT_Executable executable,
...
+                                             int64_t *result) {
+  const Executable *exe = unwrap(executable);
+  *result = exe->getNumDataSegments();
+  return mtrtStatusGetOk();
+}
+
+MTRT_Status mtrtExecutableGetDataSegmentInfo(MTRT_Executable executable,
+                                             int64_t index,
+                                             MTRT_StringView *data,
+                                             MTRT_StringView *name) {
```
christopherbate, line 142: Please add `assert(data && name && "expected valid data and name output arguments");`

mlir-tensorrt/python/mlir_tensorrt_runtime/mlir_tensorrt/runtime/_mlir_libs/_api.pyi
```diff
@@ -49,6 +49,7 @@ class Executable:
     def __init__(self, buffer: bytes) -> None: ...
     def get_signature(self, arg0: str) -> PyFunctionSignature: ...
     def serialize(self) -> bytes: ...
+    def get_data_segments(self) -> list[tuple[str, bytes]]: ...
```
christopherbate, line 52: Thanks for adding this!

christopherbate (approved):
Left some minor comments, otherwise LGTM, thanks!

mlir-tensorrt/executor/include/mlir-executor/Runtime/API/API.h
yizhuoz004: Updated, thanks!

PR #600, opened 2025-04-04
Fixes a segfault when using the DLPack interface with a memref from `…

pranavm-nvidia:
…copyToHost`

In `copyToHost`, we were previously creating a memref with the device set to `nullptr`. When trying to use the DLPack interface, this would cause a segfault.

What was really intended was to set the device to an empty optional, since a host tensor does not have an associated device. This commit updates the code to correctly create an empty optional.

Also updates the test to trigger this code path by use the DLPack interface of the created memref.


christopherbate (approved):
LGTM, thanks!

PR #601, opened 2025-04-07
Reworks how constants are handled

pranavm-nvidia:
The `Constant` op is used in two ways:

1. To represent constants in the network with `tensorrt.constant`. In this case, the data must be in CPU memory.

2. To represent inputs to compiled executables. In this case, the data must be in GPU memory and we only use `Constant` for its memref, i.e. we do not actually materialize MLIR operations with `to_mlir()`.

This commit updates `Constant.to_mlir()` to reject GPU memory since it will only be called in case (1). Additionally, `Tensor.eval()` is updated to evaluate Constants in CPU memory so that they are copied to GPU memory.

In some cases, we don't want to perform this copy, so an extra `_eval_for_internal_methods()` is provided for internal use.


tripy/nvtripy/frontend/tensor.py
```diff
@@ -195,8 +221,10 @@ def eval(self) -> "nvtripy.Tensor":
             print(f"Tensor init_time took: {(init_time - start)  * 1000.0:.3f} ms")
             print(f"Tensor evaluation took: {(eval_time - init_time)  * 1000.0:.3f} ms")
         """
-        if isinstance(self.trace_tensor.producer, Constant):
+        if isinstance(self.trace_tensor.producer, Constant) and self.trace_tensor.producer.device.kind == "gpu":
             # Exit early if the tensor has already been evaluated.
+            # We can only do this for Constants that are already on the GPU, since otherwise
+            # we need to evaluate in order to perform a copy from the host to the device.
```
yizhuoz004, line 227: If the constant is on host, then evaluating it is equivalent to a h2d copy. Will it be more efficient than going through MLIR and building a TRT engine?

yizhuoz004 (approved):

tripy/nvtripy/frontend/tensor.py
pranavm-nvidia: @yizhuoz004 Yes, it would, but I think it's a rare enough case that the costs of maintaining a separate code path outweigh the benefits.

PR #604, opened 2025-04-07
Add OptionalParameter

yizhuoz004:


tripy/examples/nanogpt/weight_loader.py
```diff
@@ -114,7 +114,7 @@ def get_submodule(module, attr_name):
         if "ln" not in key:
             weight = weight.to(torch_dtype)
         param = tp.Tensor(weight.contiguous())
+        assert key in model.state_dict()
```
pranavm-nvidia, line 117: We should store the `model.state_dict()` somewhere. This is an expensive call because it recurses over all child modules.

tripy/nvtripy/frontend/module/linear.py
```diff
@@ -103,8 +103,10 @@ def __init__(
         self.weight_quant_dim = weight_quant_dim
         self.weight_scale = None
         self.input_scale = None
-        if quant_dtype is not None and weight_quant_dim is not None:
-            self.weight_scale = DefaultParameter(shape=[self.weight.shape[weight_quant_dim]], dtype=dtype)
+        if quant_dtype is not None:
+            weight_scale_shape = [] if weight_quant_dim is None else [self.weight.shape[weight_quant_dim]]
+            self.weight_scale = DefaultParameter(shape=weight_scale_shape, dtype=dtype)
+            self.input_scale = OptionalParameter(shape=[], dtype=dtype)
```
pranavm-nvidia, line 109: We also need to avoid using `input_scale` if it's still an `OptionalParameter` by the time we get to `__call__`

yizhuoz004: Maybe in `Module.__call__`, we replace all remaining `OptionalParameter` to `None`?
Currently I do `def __bool__(self): return False` but it's not robust.

pranavm-nvidia: Yeah, that sounds better. If we do `if tensor` it will trigger an eval, so we can't use that approach.

yizhuoz004: One downside of that approach is that it modifies the `state_dict`: after the call, the `OptionalParameters` are no longer in the state_dict. 
Since both `DefaultParameter` and `OptionalParameter` are decoupled from `Tensor`, maybe it's fine to use the following approach and rely on Tripy's error message when it's not used correctly?
```py
if optional_param: ... # the __bool__ returns false
if isinstance(optional_param, tp.Tensor):...
```

pranavm-nvidia: Yup, that would work too

tripy/nvtripy/frontend/module/module.py
```diff
@@ -37,26 +37,30 @@ def _check_param_compatible(original_param, new_param, param_name):
...
 
-    # We need to evaluate here anyway, so we map the entire shape to numbers upfront to save us from recomputing
-    # them again later.
-    original_shape = tuple(map(int, original_param.shape))
-    new_shape = tuple(map(int, new_param.shape))
-    if original_shape != new_shape:
-        is_compatible = utils.result.Result.err(
-            ["New parameter shape: ", new_shape, " is not compatible with current shape: ", original_shape]
-        )
+    skip_shape_comparison = (
```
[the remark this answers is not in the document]
pranavm-nvidia: Can we invert this to `do_shape_comparison`? I think the double negative below is a bit more confusing

tripy/nvtripy/frontend/module/module.py
```diff
@@ -37,26 +37,30 @@ def _check_param_compatible(original_param, new_param, param_name):
...
+            )
 
-    # Once we know the concrete shape of the parameter, we can update the trace tensor accordingly.
-    # This not only makes the trace more informative, but is actually required for some APIs, like
-    # addConvolutionND (to set kernelDims correctly).
-    new_param.trace_tensor.shape = new_shape
+        # Once we know the concrete shape of the parameter, we can update the trace tensor accordingly.
+        # This not only makes the trace more informative, but is actually required for some APIs, like
+        # addConvolutionND (to set kernelDims correctly).
+        new_param.trace_tensor.shape = new_shape
```
pranavm-nvidia, line 63: This part should be done unconditionally (the shape of the incoming parameter is always known)

tripy/nvtripy/frontend/module/module.py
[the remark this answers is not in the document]
yizhuoz004: Updated

tripy/examples/nanogpt/weight_loader.py
```diff
@@ -114,7 +115,7 @@ def get_submodule(module, attr_name):
         if "ln" not in key:
             weight = weight.to(torch_dtype)
         param = tp.Tensor(weight.contiguous())
+        assert key in expected_keys
         tripy_state_dict[key] = param
-
-    model.load_state_dict(tripy_state_dict)
+    model.load_state_dict(tripy_state_dict, strict=False)
```
pranavm-nvidia, line 120: I think we can re-enable `strict` now and remove the manual checks? It shouldn't see the input scales as extra keys

tripy/nvtripy/frontend/module/module.py
```diff
@@ -37,21 +37,27 @@ def _check_param_compatible(original_param, new_param, param_name):
...
     new_shape = tuple(map(int, new_param.shape))
-    if original_shape != new_shape:
-        is_compatible = utils.result.Result.err(
-            ["New parameter shape: ", new_shape, " is not compatible with current shape: ", original_shape]
-        )
+    do_shape_comparison = (
+        False
+        if isinstance(original_param, (DefaultParameter, OptionalParameter)) and not original_param.is_shape_known
+        else True
+    )
```
pranavm-nvidia, line 52:
```suggestion
    do_shape_comparison = not isinstance(original_param, (DefaultParameter, OptionalParameter)) or original_param.is_shape_known
```

tripy/nvtripy/frontend/module/module.py
```diff
@@ -37,21 +37,27 @@ def _check_param_compatible(original_param, new_param, param_name):
             f"Expected a tensor for parameter: '{param_name}', but got: {type_str_from_arg(new_param)}",
         )
 
-    if not isinstance(original_param, Tensor) and not isinstance(original_param, DefaultParameter):
+    if not isinstance(original_param, (Tensor, DefaultParameter, OptionalParameter)):
```
pranavm-nvidia, line 40: I wonder if we should use `ParameterBase` in this cases now?

tripy/examples/nanogpt/weight_loader.py
yizhuoz004: There would be missing keys, the `attn.bias` I mentioned yesterday.

tripy/nvtripy/frontend/module/module.py
yizhuoz004: Will python complain about no attribute `is_shape_known`?

pranavm-nvidia: Nope, it will short circuit in the first part of the condition

tripy/nvtripy/frontend/module/module.py
yizhuoz004: Yes, updated

pranavm-nvidia (approved):

PR #606, opened 2025-04-08
Migrates Tripy to use the tensorrt dialect instead of StableHLO

pranavm-nvidia:
Migrates Tripy to use the `tensorrt` dialect instead of StableHLO. The latter is a lower level representation that the TensorRT network API, meaning that we first break down high level Tripy operations and then pattern match them back up. Using the `tensorrt` dialect allows us to go directly to TensorRT without so many intervening layers of translation, which reduces complexity and bugs, and improves performance.

Broadly, this change does the following:
- Removes FlatIR
- Make Trace responsible for mapping to MLIR
    - Trace operations represent MLIR operations 1:1
- Makes the frontend API responsible for composing Trace operations


yizhuoz004 (approved):

pranavm-nvidia:
Created a new PR since GH seems to think there's a merge conflict (there isn't): https://github.com/NVIDIA/TensorRT-Incubator/pull/607

PR #608, opened 2025-04-09
Add `tensorrt_engine` to Executable API

yizhuoz004:


tripy/Dockerfile
```diff
@@ -32,3 +32,4 @@ RUN pip install build .[docs,dev,test,build] \
 
 # Export tripy into the PYTHONPATH so it doesn't need to be installed after making changes
 ENV PYTHONPATH=/tripy
+ENV LD_LIBRARY_PATH=/usr/local/lib/python3.9/site-packages/nvidia/cuda_runtime/lib:$LD_LIBRARY_PATH
```
pranavm-nvidia, line 35: Why is this needed?

tripy/nvtripy/backend/api/executable.py
```diff
@@ -299,6 +299,43 @@ def add(a, b):
         """
         json_utils.save(self, path)
 
+    def tensorrt_engine(self) -> bytes:
+        """
+        Returns the TensorRT engine from the executable if there's a valid one.
```
pranavm-nvidia, line 304: There should always be a valid one right? If there isn't it's a Tripy/MLIR-TRT bug

tripy/nvtripy/backend/api/executable.py
```diff
@@ -299,6 +299,43 @@ def add(a, b):
         """
         json_utils.save(self, path)
 
+    def tensorrt_engine(self) -> bytes:
```
pranavm-nvidia, line 302: I wonder if we should update the name to reflect that the engine is serialized?

tripy/nvtripy/backend/api/executable.py
```diff
@@ -299,6 +299,43 @@ def add(a, b):
...
+            assert isinstance(trt_engine, bytes)
+        """
+        data_segments = self._executable.get_data_segments()
+        if len(data_segments) != 1:
+            raise_error(
+                "Cannot get tensorrt engine from multiple clusters.",
+                [f"Found {len(data_segments)} clusters in the executable."],
+            )
+        trt_cluster = data_segments[0]  # tuple of (name, data)
+        if "tensorrt_cluster_engine" not in trt_cluster[0]:
```
pranavm-nvidia, line 335: Are we guaranteed that the name will always contain `tensorrt_cluster_engine`? Since we only expect there to be a single segment containing the engine, maybe it would be better not to check the name in case it changes in the future?

tripy/nvtripy/backend/api/executable.py
```diff
@@ -299,6 +299,43 @@ def add(a, b):
...
+            The serialized TensorRT engine in a ``bytes`` object.
+
+        .. code-block:: python
+            :linenos:
+            :caption: TensorRT engine
+
+            def add(a, b):
+                return a + b
+
+            # doc: no-print-locals compiled_add executable_file
```
pranavm-nvidia, line 316:
```suggestion
            # doc: no-print-locals compiled_add
```

tripy/nvtripy/backend/api/executable.py
```diff
@@ -299,6 +299,43 @@ def add(a, b):
         """
         json_utils.save(self, path)
 
+    def tensorrt_engine(self) -> bytes:
+        """
+        Returns the TensorRT engine from the executable if there's a valid one.
+
+        Returns:
+            The serialized TensorRT engine in a ``bytes`` object.
```
pranavm-nvidia, line 307:
```suggestion
            The serialized TensorRT engine as ``bytes``.
```

tripy/nvtripy/backend/api/executable.py
yizhuoz004: Right now it should always have `tensorrt_cluster_engine`, we can drop it. I was thinking if tripy could target different backends (e.g. DLA), then checking the name would be necessary.

tripy/nvtripy/backend/api/executable.py
yizhuoz004, line 304:
```suggestion
        Returns the serialized TensorRT engine from the executable.
```

tripy/nvtripy/backend/api/executable.py
yizhuoz004, line 302:
```suggestion
    def serialized_tensorrt_engine(self) -> bytes:
```

tripy/nvtripy/backend/api/executable.py
```diff
@@ -299,6 +299,43 @@ def add(a, b):
...
+            compiled_add = tp.compile(
+                add,
+                args=[
+                    tp.InputInfo(shape=((1, 2, 3),), dtype=tp.float32),
+                    tp.InputInfo(shape=((1, 2, 3),), dtype=tp.float32),
+                ],
+            )
+
+            trt_engine = compiled_add.tensorrt_engine()
+            assert isinstance(trt_engine, bytes)
```
yizhuoz004, line 326:
```suggestion
            trt_engine = compiled_add.serialized_tensorrt_engine()
            assert isinstance(trt_engine, bytes)
```

tripy/tests/backend/api/test_executable.py
```diff
@@ -116,3 +117,15 @@ def test_file_io(self, single_return_executable):
             out1 = single_return_executable(inp, inp)
             out2 = loaded_executable(inp, inp)
             assert tp.equal(out1, out2)
+
+    def test_tensorrt_engine(self, single_return_executable):
+        from polygraphy.backend.trt import EngineFromBytes, TrtRunner
+
+        trt_engine = single_return_executable.tensorrt_engine()
```
yizhuoz004, line 124:
```suggestion
        trt_engine = single_return_executable.serialized_tensorrt_engine()
```

tripy/nvtripy/backend/api/executable.py
yizhuoz004, line 316:
```suggestion
            # doc: no-print-locals compiled_add trt_engine
```

tripy/nvtripy/backend/api/executable.py
pranavm-nvidia: Yeah, a little wordy, but more accurate.

tripy/nvtripy/backend/api/executable.py
```diff
@@ -299,6 +299,41 @@ def add(a, b):
...
+            # doc: no-print-locals compiled_add trt_engine
+            compiled_add = tp.compile(
+                add,
+                args=[
+                    tp.InputInfo(shape=((1, 2, 3),), dtype=tp.float32),
+                    tp.InputInfo(shape=((1, 2, 3),), dtype=tp.float32),
+                ],
+            )
+
+            trt_engine = compiled_add.serialized_tensorrt_engine()
```
pranavm-nvidia, line 325: Should we also show how to deserialize it? Should be something like:
```py
runtime = trt.Runtime(trt.Logger())
engine = runtime.deserialize_cuda_engine(trt_engine)
```

tripy/tests/backend/api/test_executable.py
```diff
@@ -116,3 +117,15 @@ def test_file_io(self, single_return_executable):
...
+
+    def test_tensorrt_engine(self, single_return_executable):
+        from polygraphy.backend.trt import EngineFromBytes, TrtRunner
+
+        trt_engine = single_return_executable.serialized_tensorrt_engine()
+        load_engine = EngineFromBytes(trt_engine)
+        with TrtRunner(load_engine) as runner:
+            inp_data0 = np.random.rand(2, 2).astype(np.float32)
+            inp_data1 = np.random.rand(2, 2).astype(np.float32)
+            output = runner.infer(feed_dict={"arg0": inp_data0, "arg1": inp_data1})["result0"]
```
pranavm-nvidia, line 129: This is weird - the executable is supposed to preserve the parameter names of the compiled function. I wonder why they're turning into arg0 and arg1

pranavm-nvidia (approved):

tripy/tests/backend/api/test_executable.py
yizhuoz004: `arg0` and `arg1` come from the MLIR I think, e.g. `%arg0`
```
module @"ins_%t8_outs_%t10_0" {
  func.func @main(%arg0: tensor<1x1x8x8xi4>) -> tensor<?x?x?x?xf32> {
    %cst_f32 = tensorrt.constant dense<1.000000e+00> : tensor<f32>
    %0 = tensorrt.dequantize in(%arg0 : tensor<1x1x8x8xi4>) scale(%cst_f32 : tensor<f32>) -> tensor<?x?x?x?xf32>
    return %0 : tensor<?x?x?x?xf32>
  }
}
```

tripy/nvtripy/backend/api/executable.py
yizhuoz004: I prefer to leave that part to TensorRT doc, maybe we can add a note in this API's doc which links to TRT python API? We should be able to expect user who calls this API to have adequate knowledge about TRT and TRT engines.

pranavm-nvidia: Yeah, that's a good idea. Maybe in the body of the docstring we can add:
```py
.. seealso:: Refer to the `TensorRT developer guide <https://docs.nvidia.com/deeplearning/tensorrt/latest/inference-library/python-api-docs.html#deserializing-a-plan>_` for details on how to work with serialized TensorRT engines. 
```

tripy/tests/backend/api/test_executable.py
pranavm-nvidia: Ah ok, would have been nice if we could preserve the names, but I don't see an easy way to do that. I'll file an issue

PR #622, opened 2025-05-06
Updates Executables to track the InputInfo they were compiled with

pranavm-nvidia:


tripy/nvtripy/common/shape_bounds.py
```diff
@@ -16,16 +16,44 @@
...
 from dataclasses import dataclass
-from typing import Sequence
+from typing import Tuple
+from nvtripy.utils import json as json_utils
 
 from nvtripy.types import IntLike
 
 
 @dataclass
 class ShapeBounds:
```
pranavm-nvidia, line 26: Need to decide where to document this.

PR #624, opened 2025-05-07
Update layernorm to directly use TRT API

akhilg-nv:


tripy/nvtripy/frontend/module/layernorm.py
```diff
@@ -24,6 +24,41 @@
...
+
+@wrappers.interface(
+    dtype_constraints={"input": "T1", "weight": "T1", "bias": "T1", wrappers.RETURN_VALUE: "T1"},
+    dtype_variables={"T1": ["float32", "float16", "bfloat16"]},
+)
+def layernorm(
+    input: "nvtripy.Tensor",
+    weight: "nvtripy.Tensor",
+    bias: "nvtripy.Tensor",
+    eps: float = 1e-5,
```
pranavm-nvidia, line 40: We can drop the default value here since it should always be provided by the module.

tripy/nvtripy/trace/ops/layernorm.py
```diff
@@ -0,0 +1,42 @@
...
+    normalized_shape: Sequence[int]
+    eps: float = 1e-5
+
+    infer_rank = op_utils.InferRankPolicies.same_as_input()
+
+    def infer_dtypes(self):
+        self.outputs[0].dtype = self.inputs[0].dtype
+
+    def to_mlir(self, inputs, outputs):
+        rank = len(outputs[0].shape)
```
pranavm-nvidia, line 38: Alternatively: `outputs[0].type.rank`

PR #631, opened 2025-05-14
Move internal changes

shelkesagar29:
This MR moves the following internal changes to OSS.

commit 86ad3b690c1112ead4417574899792eebe88b797
Author: Christopher Bate <cbate@nvidia.com>
Date:   Wed May 14 17:42:33 2025 +0000

    [compiler] Use "--tensorrt-builder-opt-level=0" to speed up some integration tests

    In addition, don't unnecessarily use "--tensorrt-strongly-typed=false".
    Speeds up a couple tests significantly.

commit 2ec4d0a98d4c9f27aac868c6b612fff381d772c6
Author: Yuan Yao <yuayao@nvidia.com>
Date:   Mon May 12 16:16:25 2025 -0700

    [executor] Declare locals in entry block for values with external uses.

    Previously, values defined in non-entry blocks with uses outside were
    emitted as Lua globals in the Executor MLIR-to-Lua translation, causing
    correctness issues in coroutine contexts due to shared global state.
    This change declares such values as locals in the entry block, which are
    initialized to `nil`.

commit 8b5784ccc1269b1f059db7edb0daad4090926a99
Author: Chris Bate <cbate@nvidia.com>
Date:   Fri May 9 08:57:35 2025 -0700

    [compiler] Eliminate the `CompilationTaskOptionsBase::finalize` method

    Previously, we used a separate method to calculate overrides after
    parsing the pass options. This was only used for calculating the
    device information from the host if specified. However, we can just
    use the LLVM CL options callback to do this.

commit 542ce2a166c7a237d00caf16e8e126bf355fe25d
Author: Chris Bate <cbate@nvidia.com>
Date:   Tue May 6 14:01:18 2025 -0700

    [compiler] Fix enforcement of entrypoint buffer result allocation policy

    In the `plan-alloc-tensors` pass, we previously were not always inserting IR
    that would "force entrypoints to return allocs" when the corresponding
    option was set. In particular, certain edge cases, like the direct return
    of block arguments, would compile to a form where the compiler would
    drop the returned value.

    This change fixes the issue and enfornces materialization of a copy
    when the returned value is equivalent to a block argument.

    Note that this change does not handle enforcing a policy regarding
    whether function arguments are mutable or not.

commit 7dd0d45623ba33a92fbc2d6d0f928abafc857609
Author: Sagar Shelke <sagshelke@nvidia.com>
Date:   Tue Apr 29 14:22:16 2025 -0700

    [compiler] Add `MaterializeDenseResourceElementsAttr` pass in `StablehloExt` dialect

    This MR adds `MaterializeDenseResourceElementsAttr` pass in `StablehloExt` dialect. It
    converts `DenseElementsAttr` with elements more than `elementCountThreshold` to
    `DenseResourceElementsAttr`. Newer `DenseResourceElementsAttr` can be released from the
    context when not referred, unlike `DenseElementsAttr` which are immortal.

    Valid MLIR test is added.

commit ba5e5f6a1aedfbbc47db295c8b6c4c5df7e30605
Author: Chris Bate <cbate@nvidia.com>
Date:   Thu Apr 24 03:14:51 2025 -0700

    [compiler] Change fallback behavior for bounds detection in 'plan-create-closed-regions'

    In the 'plan-create-closed-regions' pass, we should not raise an error
    if we cannot detect whether a region input is a TensorKind through the
    TensorKindAnalysis. In certain edge cases, e.g. when the result of a
    `scf.while` is not used, the TensorKindAnalysis may not provide useful
    information since it is a sparse backwards dataflow analysis. Generally,
    however, these values don't require bounds, so we can just populate a
    "none" bounds by default if the kind cannot be detected.

commit c314f361ccda2648b4ab533f23a3619c24c53c3e
Author: Chris Bate <cbate@nvidia.com>
Date:   Thu Apr 24 00:21:53 2025 -0700

    [compiler] Fix handling of edge case during `stablehlo.dot_general`

    `stablehlo.dot_general` can have a result element type different
    from the operand element types if certain rules are met. The result
    element type should be a larger bitwidth than the operands  We were
    not accounting for this in the converter. Since TensorRT's manual
    states that the accumulator element type can be controlled using
    casts on the operands, we handle this by casting the operands to
    the result element type.

commit 1eda12cd118e6f42fb9106d4d555094981676f3b
Author: Chris Bate <cbate@nvidia.com>
Date:   Wed Apr 23 22:42:18 2025 -0700

    [compiler] Remove duplicate handling of edge cases between StablehloToTensorRT and preprocessing

    Previously, we had some very specific logic for optimizing certain edge cases
    during `stablehlo.dot_general` conversion to TensorRT. This included
    things like conversion to pure elementwise multiplication, handling of multiple
    outer-product dimensions (but strangely only for RHs). Most of this logic was
    introduced to handle failure cases as they arose during testing of
    specific models. All of these edge cases are now handled specifically
    during the Stablehlo preprocessing pass `stablehlo-canonicalize-dot-general`,
    so we no longer need special logic during the Stablehlo-to-TensorRT
    conversion, and it can now be removed.
    I moved relevant test cases into the `stablehlo-canonicalize-dot-general`
    pass tests from the `convert-stablehlo-to-tensorrt` pass tests.

commit 56432616b43cf72b03335d32eb1f33f5348ca4e4
Author: Chris Bate <cbate@nvidia.com>
Date:   Wed Apr 23 21:39:17 2025 -0700

    [compiler] StablehloToTensorRT: don't use F32 approximation for i32 dot ops

    Previously, we used f32 as an approximation when we encountered i32 operands
    in `stablehlo.dot|dot_general` operations. Since there is no obvious
    use case to support this approximation path using TensorRT, I removed the
    logic in the converter (only one instance - `stablehlo.dot` -
    actually had test coverage).

commit ee06c18680a156cb1815d7e40ef4f71dd1ae6ec1
Author: Chris Bate <cbate@nvidia.com>
Date:   Wed Apr 23 20:36:07 2025 -0700

    [compiler] NFC: declare helper method for common cast-replace logic in 'convert-stablehlo-to-tensorrt'

commit 542cc2dcf2f1a28d6831add54065864d6649c36b
Author: Chris Bate <cbate@nvidia.com>
Date:   Wed Apr 23 19:25:34 2025 -0700

    [compiler] NFC: cleanup repetitive TensorRT major version variable definitions

    This change cleans up repetitive declarations of 'targetTrtMajorVersion'
    by just declaring it as a member variable of the base pattern. It cuts
    out about 100 lines of code. Similarly, we can pass the 'targetTrtMajorVersion'
    to the creation of the rewrite utility rather than passing it to the
    `checkAndCreate` commands, reducing more repitition.

commit 14b8768d4493c1962da5472a83a0bfa34b4a1d2d
Author: Chris Bate <cbate@nvidia.com>
Date:   Wed Apr 16 13:01:52 2025 -0700

    [compiler] Fix handling of Dataflow state updates in the 'plan-clustering' pass

    Previously, we were unnecessarily re-running the entire dataflow solver
    during the clustering pass. This change improves the handling of dataflow
    state updates, adds a targeted regression test, and removes some unnecessary
    constant duplication from the 'plan-clustering' pass.

    The "solver-aware rewrite listener" from the 'plan-alloc-tensors' pass is
    now factored out into a common header file, and both passes use the listener
    to make state updates. Some minor cleanup to the 'plan-alloc-tensors' pass
    is also included. Explicit dependence of 'plan-clustering' on Stablehlo
    is removed.

commit 9d56ec5914d3bea0fb9b2f3dc643e781aeaf6cde
Author: Sagar Shelke <sagshelke@nvidia.com>
Date:   Fri Apr 11 16:17:35 2025 -0700

    [Conversion/StablehloToTensorRT] Avoid creating TensorRT conv ops with negative padding.

    TensorRT convolution op doesn't support negative padding. Previously,
    StablehloToTensorRT conversion pass was creating TRT conv ops
    with negative padding. This MR fixes this issue by first slicing inputs to
    make effective padding zero.

    Check for when `stablehlo.convolution` op represents deconvolution op
    is simplified to presence of non-unit `lhs_dilation`.

    A new test case is added and existing test cases are modified to support
    changes.

commit c552b9089dca66e4a98df471bc0ca04cf43c91c4
Author: Sagar Shelke <sagshelke@nvidia.com>
Date:   Fri Apr 11 15:41:23 2025 -0700

    [TensorRTToEmitC] Fix issues and enhance emitc conversion

    This MR enhances emitc conversion by,
    1. Adding support for multiple layers in `NvInferAdaptor.h`
    2. Fix multiple bugs
    3. Remove special deleter `UniquePtr`
    4. At the top of every generated c++ file, there will be comment
    explaining code structure and build instructions. Done via
    `VerbatimOp` op.

    Update tests where needed.

commit 59b6d5b9d6689816347cea6a3613a6e494d8f17b
Author: Chris Bate <cbate@nvidia.com>
Date:   Mon Apr 7 11:46:56 2025 -0700

    [compiler] Fix some edge cases in `stablehlo-ext-canonicalize-convolution`

    Fix an issue where the `stablehlo.convolution` rewrite patterns were
    not handling certain edge cases:

    - No spatial dimensions (can canonicalize to `stablehlo.dot_general`)
    - Spatial dimensions are not contiguous

    This change adds some additional logic and new tests. I added a pattern
    to make some effort to turn the "empty spatial dimensions" case into a
    canonical `stablehlo.dot_general` pattern, but it will not handle cases
    where the result needs to be transpsoed. The case where the spatial
    dimensions are not contiguous is also not changed. In the future, we
    should make an effort to canonicalize these cases as well.

commit 87f85654b20213e9e75b672ff2088d9c25c5d56a
Author: Chris Bate <cbate@nvidia.com>
Date:   Mon Apr 7 10:22:18 2025 -0700

    [compiler] Fix `stablehlo-to-tensorrt` conversion for `stablehlo.exponential_minus_one` (bf16)

    This change fixes an issue in the `stablehlo-to-tensorrt` conversion for
    `stablehlo.exponential_minus_one` when the input is of type `bf16`. The
    conversion was previously using a `tensorrt.identity` operation to cast
    the input to `f32` in order to the approximation. However, this was only
    done if the input type was not `f16`, ignoring other lower precision float
    types. This then could cause a crash because the approximation code assumed
    a `f32` tensor type. This change fixes the issue and adds a regression
    test.

commit 19d54f2518098b94231647ad1209c28216533494
Author: Chris Bate <cbate@nvidia.com>
Date:   Sat Apr 5 13:59:41 2025 -0700

    [compiler] Fix crash in 'stablehlo.log_plus_one' conversion to TensorRT

    The conversion was using the wrong type for a `FloatAttr` when the
    input had an element type other than `f32`. Fixed and added tests for
    TRT 10 and TRT 8 conversion cases.

commit dc1362a60cf1061e128aa0b89f5ee2b846b3cd55
Author: Chris Bate <cbate@nvidia.com>
Date:   Fri Apr 4 10:52:05 2025 -0700

    [compiler] Fix crash in `plan-create-closed-regions`

    This change fixes an issue where we crashed while creating a bounds
    attribute for a tensor classified as having host access. In fact,
    we should only query the bounds analysis for tensors that have host or
    index element types. We do not currently support deriving bounds for
    other types of tensors.

commit 1700bc38192f52c78a5909ecca7f31cce777f976
Author: Chris Bate <cbate@nvidia.com>
Date:   Fri Apr 4 10:51:45 2025 -0700

    [compiler] Fix crash in `convert-stablehlo-to-tensorrt`

    This change fixes a crash in `convert-stablehlo-to-tensorrt` in the
    pattern that converts `stablehlo.sort` to `tensorrt.top_k`. The
    pattern match logic performed a `dyn_cast` on a potentially null pointer,
    which caused a crash/assertion. This change fixes the issue and adds
    a regression test.

commit 42d74af68a3fc924f0938312537d8a8c48a96c1e
Author: Chris Bate <cbate@nvidia.com>
Date:   Fri Apr 4 09:49:47 2025 -0700

    [compiler] Fix crash in `stablehlo-canonicalize-dot-general`

    This change fixes two failure cases in `stablehlo-canonicalize-dot-general`:

    1. Failure to check conditions for invoking `Type::getIntOrFloatBitWidth`.
    2. Incorrect IR generation when rewriting `stablehlo.dot_general` to
       `stablehlo.multiply` if the operands have a different element type
       than the result.

commit 21b46472a07f073e6ac3914f72ffdec7d59972c5
Author: Chris Bate <cbate@nvidia.com>
Date:   Fri Apr 4 09:16:25 2025 -0700

    [compiler] Fix failure in `plan-alloc-tensors` when complex element types are present

    Fixes a failure in `plan-alloc-tensors` when complex element types are present.
    The code previously always checked `tensor::BitcastOp::areCastCompatible`, but
    this method will crash when the element type is not an integer or floating point
    type. Adds an additional test case involving complex element types.

commit 30659163c958620ce48df3e9e8cb818aac94f9a1
Author: Chris Bate <cbate@nvidia.com>
Date:   Thu Apr 3 20:01:39 2025 -0700

    [compiler] Fix `convert-stablehlo-to-tensorrt` for certain kinds of `stablehlo.scatter` ops

    This change fixes a crash in the `convert-stablehlo-to-tensorrt` pass
    when encountering a `stablehlo.scatter` op with an empty
    `scatter_dims_to_operand_dims` array.

commit ff55dfafb08a6330374fd6ff205676c5afc1000b
Author: Chris Bate <cbate@nvidia.com>
Date:   Thu Apr 3 10:21:46 2025 -0700

    [compiler] Fix DPS function rewrite in `plan-alloc-tensors` to account for shape differences

    This change fixes a bug in the DPS function rewrite where the logic
    was not taking into account the potential need to reshape the new
    DPS output argument to match the type of the `tensor.empty` that it
    replaces.

commit d3d3469100d5e4735c063ad65ca58ab7be6f10c8
Author: Sagar Shelke <sagshelke@nvidia.com>
Date:   Tue Apr 1 11:02:40 2025 -0700

    [TensorRTToEmitC] Add support for topk and fix a bug

    This MR makes the following changes in TensorRTToEmitC pass,
    1. Fix a bug where during creation of `emitc::OpaqueCall` op, number of results
    was always assumed to be one. This is not always correct. This is fixed by
    creating number of results opaque tensors by querying op under conversion.
    2. Add support for topk op conversion.

    Valid test is added.

commit 97b9501f7da2607de0dfc3e57b45729864cd83cd
Author: Chris Bate <cbate@nvidia.com>
Date:   Mon Mar 31 16:09:18 2025 -0700

    [executor] Workaround Flatbuffers alignment bug

    We seem to be hitting a bug in the Flatbuffers library which only appears
    in very specific conditions (translating two `func.func` plus one
    `executor.data_segment` and some other conditions). The symptom is that
    the `FunctionSignature` components don't have the correct alignment. The
    problem appears to be alleviated if we ensure that the "name" string of
    the function is serialized second. Regression test is added.

commit a478f84219352447fc1d12d6899e1707d0803e09
Author: Chris Bate <cbate@nvidia.com>
Date:   Fri Mar 28 14:55:53 2025 -0700

    [executor] Fail conversion of `memref.load|store` if memref is not in host-visible memory

    This change adds a check to the conversion of `memref.load|store` to the
    Executor dialect to ensure that the memref is in host-visible memory. If
    the memref is not in host-visible memory, the conversion will fail with an
    error. Generally, the compiler should eliminate such cases immediately
    after/during bufferization, but recently during some experimentation
    I found that there are some edge cases that could result in creation of
    such ops during lower-level passes (e.g. during `memref-expand-ops`). To
    fully address this, we need to check after bufferization for other
    operations that may be lowered into load/store of device memory (e.g.
    `memref.reshape` where the `shape` operand is in device memory).

commit 707275d76d3c8ea3b419eec7ce958afdacc9ed88
Author: Sagar Shelke <sagshelke@nvidia.com>
Date:   Fri Mar 28 13:43:14 2025 -0700

    [Dialect/CUDA] Allow complex f32 and f64 types in `cuda.launch` operation

    This MR allows complex f32 and f64 types in `cuda.launch` operation.
    Valid MLIR test is added.

commit 52c8200008a5a81742425400fabfd9de9d8fdf0b
Author: Chris Bate <cbate@nvidia.com>
Date:   Fri Mar 28 12:52:52 2025 -0700

    [compiler] Fix Plan Dialect 'input-kind' pass options

    This change fixes the 'input-kind' pass options for the Plan dialect
    passes. The options did not include the correct `llvm::cl::values` in
    the option definition, which could cause a crash if the textual pipeline
    was being dumped (e.g. when specifying `-mlir-pass-pipeline-crash-reproducer`).

commit 3a16555299e8def6f8d78ff507be072eb143d675
Author: Chris Bate <cbate@nvidia.com>
Date:   Wed Mar 26 15:41:25 2025 -0700

    [compiler] Update `memref.global|get_global` lowerings

    This change updates the lowering of `memref.global` and `memref.get_global`
    to use the updated `executor.data_segment` operation. Since we added
    support in the runtime for initializing the data segments in the appropriate
    address space, it simplifies things a bit in the compiler. The next
    change will be to enable more general use of the `executor-alloc-tensors`,
    which needed this change.

commit f2e479a230088c7c2a3e23e69b28933356468732
Author: Chris Bate <cbate@nvidia.com>
Date:   Wed Mar 26 14:27:53 2025 -0700

    [executor] Add "uninitialized_size" attribute to DataSegment in Executable schema

    This change adds support for translating the "uninitialized" attribute
    of the DataSegmentOp into the Executable schema.

commit fbb58c4efa5a0bfff5382592530e73811eafb89a
Author: Chris Bate <cbate@nvidia.com>
Date:   Wed Mar 26 11:10:19 2025 -0700

    [executor] Refactor serialization during translation to use common utilities

    This refactor moves the logic for serializing ElementsAttr into a common
    utility function, `serializeElementsAttr`, which is accepts a generic
    interface for serializing the data. The method allows handling the common
    kinds of ElementsAttr (DenseIntOrFPElementsAttr, including splat, and
    DenseResourceElementsAttr).

commit f26af3f1edba0bfe8bd595a1bc755f136adf5192
Author: Chris Bate <cbate@nvidia.com>
Date:   Wed Mar 26 08:49:36 2025 -0700

    [compiler] Fix TensorKindOpInterface for Stablehlo ops

    Previously, the implementation of TensorKindOpInterface for Stablehlo ops was
    incorrect. We were not checking whether the operands are candidates as host
    tensors (due to size constraints). If they cannot be candidates as host tensors,
    we should treat them as device tensors.

commit 6bc0f3e58f42ce75d1cb07f0468b38c8da99558c
Author: Sagar Shelke <sagshelke@nvidia.com>
Date:   Tue Mar 25 11:59:38 2025 -0700

    [tensorrt/Dialect] Remove `tensorrt.identity84` op

    This MR removes `tensorrt.identity84` op and tests/uses
    around it because lowest TRT version we aim for is 8.6. Even on customer
    side, most are on minimum 8.6.

commit 3abcce9382fac1b5327547ce98ac865eb13c1743
Author: Sagar Shelke <sagshelke@nvidia.com>
Date:   Tue Mar 25 11:06:50 2025 -0700

    [HotFix] Fix lit command in failing test

    After 35e4338c95432bab4a381e28a52cf65cfb14a9bb, test
    `workaround-oob-slice.mlir` fails because `tensorrt-apply-bug-wars`
    pass now has become `tensorrt-apply-wars`.
    This MR corrects lit command.

commit 35e4338c95432bab4a381e28a52cf65cfb14a9bb
Author: Sagar Shelke <sagshelke@nvidia.com>
Date:   Tue Mar 25 10:36:15 2025 -0700

    [tensorrt/lib] Rename and update `ApplyBugWorkarounds` pass

    This MR makes the following changes,
    1. Rename `ApplyBugWorkarounds` pass to `ApplyWorkarounds` pass to represent
    TensorRT generic as well as bug specific workarounds.
    2. Update pass description to remove pattern information. New pass is expected
    to get updated with new patterns frequently. Documenting patterns in pass description
    will bloat the description.
    3. Loosen up verification of `tensorrt.identity` to allow `x -> ui8 -> x`
    conversions where x doesn't necessarily have to be of type {f32, f16}.
    4. If `tensorrt.identity` op is doing `ui8 -> x` OR  `x -> ui8` conversion where `x`
    is not in {f32, f16}, rewrite it as `ui8 -> fp32 -> x` OR `x -> fp32 -> ui8` respectively.
    This is done in newly added `IdentityUI8` pattern.
    5. Rename modules where necessary.

commit 1d99e7b9addbf00e745aca6b0c6b07a6daf912d5
Author: Chris Bate <cbate@nvidia.com>
Date:   Tue Mar 25 08:04:01 2025 -0700

    [executor] Fix requiring CUDA for some translation tests

    Fixes issue where CUDA is initialized for `executor-runner` even if
    user is just using `--dump-function-signatures` to inspect the executable
    instead of running it.

commit 6d620614becc732b181bd3ba8bf863849ff0b390
Author: Chris Bate <cbate@nvidia.com>
Date:   Mon Mar 24 15:35:10 2025 -0700

    NFC: Align the naming of different folders

    This change updates naming of different folders to be consistent:
    - Dialect/StableHloExt -> Dialect/StablehloExt
    - Interface/ -> Interfaces/

commit b0b9c398b97873a8adf38542483eb8a2bb18f284
Author: Christopher Bate <cbate@nvidia.com>
Date:   Mon Mar 24 22:17:31 2025 +0000

    NFC: remove two dead files that escaped delation from previous changes


christopherbate (approved):
LGTM, should be fine to merge after CI succeeds.

shelkesagar29:
Closing in favor of creating a new one with newer changes.

PR #632, opened 2025-05-14
Support ellipsis in slice

yizhuoz004:


tripy/nvtripy/frontend/ops/slice.py
```diff
@@ -27,14 +27,17 @@
 from nvtripy.utils.types import type_str_from_arg
 from nvtripy.utils.utils import make_list
 
+EllipsisType = type(Ellipsis)
```
yizhuoz004, line 30: `from types import EllipsisType` does not work for py 3.9.

pranavm-nvidia (approved):

tripy/nvtripy/frontend/ops/slice.py
```diff
@@ -95,8 +98,15 @@ def __getitem__(
...
     index = make_list(index)
+    ellipsis_count = index.count(Ellipsis)
+    if ellipsis_count > 1:
+        raise_error("An index can only have a single ellipsis ('...')")
     if len(index) > self.rank:
         raise_error(f"Input tensor has a rank of {self.rank} but was attempted to be sliced with {len(index)} indices")
+    if ellipsis_count:
+        ellipsis_idx = index.index(Ellipsis)
+        num_slices = self.rank - len(index) + 1
+        index[ellipsis_idx : ellipsis_idx + 1] = [slice(None)] * num_slices
```
pranavm-nvidia, line 109: Could we also briefly mention ellipsis in the docstring? A short code example would be good too.

tripy/nvtripy/frontend/ops/slice.py
yizhuoz004: Seeing this when building docs: `/tripy/nvtripy/frontend/tensor.py:docstring of nvtripy.frontend.ops.slice.__getitem__:1:py:class reference target not found: ellipsis`, is there a good way to make it work? We probably don't need to document `EllipsisType`

pranavm-nvidia: In `conf.py` there's an exception list:
```py
nitpick_ignore = {
    ("py:class", "nvtripy.types.ShapeLike"),
    ("py:class", "nvtripy.types.TensorLike"),
    ("py:class", "Tensor"),
}
```
You can add `EllipsisType` there. I guess it would probably be something like `nvtripy.frontend.ops.slice.EllipsisType`

pranavm-nvidia: Seems like it still fails. Maybe it's just "ellipsis"?

yizhuoz004: Yes, just `ellipsis` works.

PR #633, opened 2025-05-14
Add InstanceNorm

akhilg-nv:


tripy/nvtripy/frontend/module/instancenorm.py
```diff
@@ -0,0 +1,158 @@
...
+    num_channels: int,
+    eps: float = 1e-5,
+) -> "nvtripy.Tensor":
+
+    input_rank = input.rank
+
+    if input_rank < 3:
+        raise_error(
+            "Input is expected to have shape (N, C, D1, ...) where N is the batch size, C is the number of channels, and D1, ... are the spatial dimensions",
+            details=[f"Got {input.shape} which has rank {input_rank} < 3."],
```
pranavm-nvidia, line 49: nit:
```suggestion
            details=[f"Got {input.shape} which has rank {input_rank}."],
```

tripy/tests/integration/test_instancenorm.py
```diff
@@ -0,0 +1,97 @@
...
+class TestInstanceNorm:
+
+    @pytest.mark.parametrize("torch_dtype, tp_dtype", DTYPES)
+    @pytest.mark.parametrize("input_shape", INPUT_SHAPES)
+    def test_instancenorm_accuracy(self, torch_dtype, tp_dtype, input_shape, eager_or_compiled):
+        eps = 1e-5
+        num_channels = input_shape[1]
+
+        # Choose appropriate PyTorch InstanceNorm based on input dimensions
+        if len(input_shape) == 3:  # 1D data (NCL)
```
pranavm-nvidia, line 39: Looks like the arguments to each module are the same. We can do something like:
```py
TorchInstanceNorm = {3: torch.nn.InstanceNorm1d, 4: torch.nn.InstanceNorm2d, 5: torch.nn.InstanceNorm3d}[len(input_shape)]

instancenorm = TorchInstantNorm(num_features=num_channels, ...)
```

tripy/nvtripy/frontend/module/instancenorm.py
```diff
@@ -0,0 +1,158 @@
...
+        )
+
+    if input.shape[1] != num_channels:
```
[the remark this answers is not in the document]
akhilg-nv: ah no, should it just be taking from `input.trace_tensor.shape`?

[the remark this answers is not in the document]
pranavm-nvidia: But the shape might be unknown there. We might need to just defer this check to TRT.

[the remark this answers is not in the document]
akhilg-nv: Hm I see. Without this check the error message is a bit less clear e.g. 
```
E       MTRTException: failed to run pass pipeline                                                                                      
E           (%t56) error: 'tensorrt.slice' op inferred type(s) 'tensor<2x6x4x4xf32>' are incompatible with return type(s) of operation '
tensor<?x3x?x?xf32>
```
This is because we explicitly set the channel dimension of the input shape based on the module `num_channels.` But yeah I guess in case of dynamic shapes I don't see another solution than to defer.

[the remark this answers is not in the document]
pranavm-nvidia: Why is it failing in `slice`? I think the error would be clearer if it were failing in `normalization` instead. I'm wondering if we should check the shape in the case of static channel dimensions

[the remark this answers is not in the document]
akhilg-nv: updated

tripy/nvtripy/frontend/module/instancenorm.py
```diff
@@ -0,0 +1,158 @@
...
+
+    input_rank = input.rank
+
+    if input_rank < 3:
+        raise_error(
+            "Input is expected to have shape (N, C, D1, ...) where N is the batch size, C is the number of channels, and D1, ... are the spatial dimensions",
+            details=[f"Got {input.shape} which has rank {input_rank}."],
+        )
+
+    if input.trace_tensor.shape[1] != -1 and input.trace_tensor.shape[1] != num_channels:
```
pranavm-nvidia, line 52:
```suggestion
    if input.trace_tensor.shape[1] != constants.DYNAMIC_DIM and input.trace_tensor.shape[1] != num_channels:
```

tripy/nvtripy/trace/ops/instancenorm.py
```diff
@@ -0,0 +1,43 @@
...
+
+from dataclasses import dataclass
+import nvtripy.trace.ops.utils as op_utils
+from nvtripy.trace.ops.base import TraceOp
+from mlir_tensorrt.compiler.dialects import tensorrt
+from mlir_tensorrt.compiler import ir
+
+
+@dataclass(repr=False)
+class InstanceNorm(TraceOp):
```
yizhuoz004, line 26: Is my understanding correct that `InstanceNorm` is the base class and other ops like `GroupNorm`, `LayerNorm` are just variants with different parameters?

yizhuoz004: I also wonder if it's simpler to call it `Normalization`.

akhilg-nv: They are all calling the tensorrt.Normalization operation with different parameters (expect groupnorm, which requires a WAR to use instancenorm). We could make it a single generic Normalization TraceOp and move the logic of providing the correct args (axis, num_groups, etc.) to the frontend op that is called by the Module. I'm not sure if it'll be easier to understand in code / IR logging this way, but it would mean we need one less trace op (eventually 2-3 less once groupnorm can use the API directly and if/when we add batchnorm).

pranavm-nvidia: I'd lean towards more trace ops since that makes it easier to understand the trace. e.g. I'd rather see a `GroupNorm` than just a `Normalization` when I'm looking at the trace IR.

yizhuoz004: Good point. My only suggestion is try to avoid code duplication, if layernorm / batchnorm etc are very similer, maybe a base class is helpful to reuse code.

akhilg-nv: Right now I don't think much is duplicated, but I expect that I might to do a small cleanup PR of the frontend verification and the tests to make it consistent across all the normalization ops, once I've completed all the core implementations. Perhaps the verification can be reused, and for the tests we sometimes missed accuracy issues due inputs and/or the scale and bias being too simple (e.g. ones and zeros).

tripy/nvtripy/frontend/module/instancenorm.py
```diff
@@ -0,0 +1,158 @@
...
+        raise_error(
+            "Input is expected to have shape (N, C, D1, ...) where N is the batch size, C is the number of channels, and D1, ... are the spatial dimensions",
+            details=[f"Got {input.shape} which has rank {input_rank}."],
+        )
+
+    if input.trace_tensor.shape[1] != constants.DYNAMIC_DIM and input.trace_tensor.shape[1] != num_channels:
+        raise_error(
+            "The input channel dimension does not match the number of channels specified in the InstanceNorm constructor",
+            details=[f"Got {input.shape[1]} channels in the input, but expected {num_channels} channels"],
+        )
```
pranavm-nvidia, line 56: Same here, would be good to invert so the message reads:
```
Expected X channels in the input, but got Y channels.
    The input channel dimension must match the number of channels specified to the InstanceNorm
```

tripy/nvtripy/frontend/module/instancenorm.py
```diff
@@ -0,0 +1,158 @@
...
+    from nvtripy.frontend.ops.reshape import reshape
+
+    broadcast_shape = (1, num_channels) + (1,) * (input_rank - 2)
+    weight = reshape(weight, broadcast_shape)
+    bias = reshape(bias, broadcast_shape)
+
+    # The MLIR Graph may have dynamic dimensions, so we explicitly set static dimensions in the trace shape tensors
+    weight.trace_tensor.shape = broadcast_shape
+    bias.trace_tensor.shape = broadcast_shape
+    input.trace_tensor.shape = input.trace_tensor.shape[:1] + (num_channels,) + input.trace_tensor.shape[2:]
```
pranavm-nvidia, line 68: More concisely:
```suggestion
    input.trace_tensor.shape[1] = num_channels
```

tripy/nvtripy/frontend/module/instancenorm.py
```diff
@@ -0,0 +1,158 @@
...
+@wrappers.interface(
+    dtype_constraints={"input": "T1", "weight": "T1", "bias": "T1", wrappers.RETURN_VALUE: "T1"},
+    dtype_variables={"T1": ["float32", "float16", "bfloat16"]},
+)
+def instancenorm(
+    input: "nvtripy.Tensor",
+    weight: "nvtripy.Tensor",
+    bias: "nvtripy.Tensor",
+    num_channels: int,
+    eps: float = 1e-5,
```
pranavm-nvidia, line 41: Same rationale as for the trace op:
```suggestion
    eps: float,
```

tripy/tests/integration/test_instancenorm.py
```diff
@@ -0,0 +1,71 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import nvtripy as tp
+import pytest
+import torch
+
+DTYPES = [(torch.float16, tp.float16), (torch.float32, tp.float32)]
```
pranavm-nvidia, line 22: I think we have this mapping under `tests.helper`

tripy/nvtripy/frontend/module/instancenorm.py
akhilg-nv: This doesn't work because the shape is a tuple and items aren't assignable

pranavm-nvidia: Ah right

tripy/nvtripy/frontend/module/instancenorm.py
```diff
@@ -0,0 +1,158 @@
...
+        InstanceNormOp,
+        [input, weight, bias],
+        num_channels=num_channels,
+        eps=eps,
+    )
+
+
+@export.public_api(document_under="operations/modules")
+@dataclass
+@utils.wrappers.constant_fields(["num_channels", "dtype"])
```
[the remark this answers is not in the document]
akhilg-nv: I'll make it a constant field

PR #637, opened 2025-05-16
Implement GroupNorm with TRT APIs

akhilg-nv:
Due to an API incompatibility issue with Torch GroupNorm and TensorRT GroupNorm, the implementation uses InstanceNorm instead as a workaround (following the same WAR used by the ONNX parser and Torch -> ONNX converter). 

The latest opset of ONNX supports scale and bias with shape `(num_channels,)`, so it is likely that the TRT API will see this eventually supported, at which point we can switch to the most direct implementation. 

Running 1k iterations in a loop shows the new implementation is roughly 17% faster on average. The nsys trace shows that the new implementation is significantly better fused (only two computation kernels, for the instancenorm and the affine transform), and the performance for a single iteration of the module is up to 40% faster (30µs vs 50µs).


tripy/nvtripy/frontend/module/groupnorm.py
```diff
@@ -35,6 +37,10 @@ class GroupNorm(Module):
     :math:`\text{GroupNorm}(x) = \Large \frac{x - \bar{x}}{ \sqrt{\sigma^2 + \epsilon}} \normalsize * \gamma + \beta`
 
     where :math:`\bar{x}` is the mean and :math:`\sigma^2` is the variance.
+
+    The input should have shape :math:`[N, C, D1, ...]` where :math:`N` is the batch size, :math:`C` is the number of channels, and :math:`D1, ...` are the feature dimensions.
+
+    Note that the current implementation uses InstanceNorm as a workaround to implement this operation due to TRT API compatibility issues.
```
pranavm-nvidia, line 43: Unless this is affects the functionality, I would consider this an implementation detail and not include it in the docstring. 
```suggestion
```

tripy/nvtripy/frontend/module/groupnorm.py
```diff
@@ -104,6 +113,10 @@ def __init__(
         self.bias = DefaultParameter((num_channels,), dtype=dtype)
         self.eps = eps
 
+        self._instance_norm = InstanceNorm(num_groups, dtype=dtype, eps=eps)
+        self._instance_norm.weight = ones((num_groups,), dtype=dtype)
+        self._instance_norm.bias = zeros((num_groups,), dtype=dtype)
```
pranavm-nvidia, line 118: Can we just call the `instancenorm` function? I don't think this way will work out well with `state_dict`s because there would be extra keys.

tripy/nvtripy/frontend/module/layernorm.py
```diff
@@ -44,13 +45,19 @@ def layernorm(
...
+        )
 
-        broadcast_shape = (1,) * (input_rank - D) + normalized_shape
-        weight = reshape(weight, broadcast_shape)
-        bias = reshape(bias, broadcast_shape)
+    from nvtripy.frontend.ops.reshape import reshape
+
+    broadcast_shape = (1,) * (input_rank - D) + normalized_shape
+    weight = reshape(weight, broadcast_shape)
+    bias = reshape(bias, broadcast_shape)
```
pranavm-nvidia, line 60: I actually think keeping the `if input_rank > D` check would be good here, since `reshape` isn't smart enough to know when it's actually doing a no-op in the frontend.

tripy/tests/frontend/module/test_instancenorm.py
```diff
@@ -24,28 +24,26 @@ def test_instancenorm_improper_rank(self):
...
-        tp_instancenorm.weight = tp.ones((3,))
-        tp_instancenorm.bias = tp.ones((3,))
+        tp_instancenorm.initialize_dummy_parameters()
 
         # dynamic shape
         x = tp.ones((2, 6, 4, 4))
         with helper.raises(
             tp.TripyException,
-            match="MTRTException: failed to run pass pipeline",
+            match=r"'tensorrt.slice' op inferred type\(s\) 'tensor\<2x6x4x4xf32\>' are incompatible with return type\(s\) of operation 'tensor\<\?x3x\?x\?xf32\>'",
```
pranavm-nvidia, line 46: This is not an ideal error :slightly_frowning_face: 
Do you know why there's a `tensorrt.slice` instead of a `tensorrt.normalization`? I think the latter would have had a clearer error message.

tripy/tests/integration/test_groupnorm.py
```diff
@@ -42,18 +44,58 @@ def test_groupnorm_accuracy(self, torch_dtype, tp_dtype, input_shape, num_groups
...
+        means = reshaped.mean(dim=(2, 3))
+        vars = reshaped.var(dim=(2, 3), unbiased=False)
+
+        mean_abs = means.abs().mean().item()
+        var_diff = (vars - 1).abs().mean().item()
+
+        assert mean_abs < 2e-4, f"Group mean should be close to 0, got {mean_abs}"
+        assert var_diff < 1e-3, f"Group variance should be close to 1, got {var_diff}"
+
+        # Comparison test with the affine transformation included
```
pranavm-nvidia, line 73: Seems like this should be a separate test. We can use a fixture to reuse the modules/inputs.

tripy/tests/integration/test_instancenorm.py
```diff
@@ -50,25 +50,39 @@ def test_instancenorm_accuracy(self, dtype, input_shape, eager_or_compiled):
             dtype=dtype,
         )
 
-        torch.nn.init.uniform_(instancenorm.weight)
-        torch.nn.init.uniform_(instancenorm.bias)
+        input = torch.empty(*input_shape, dtype=torch_dtype, device="cuda").uniform_(0, 10)
+        tp_input = tp.Tensor(input, dtype=dtype)
+
+        # Verify normalized output has approximately mean=0, std=1
+        tp_instancenorm.weight = tp.ones((num_channels,), dtype=dtype)
```
pranavm-nvidia, line 57: Same here, we can just parametrize over the weight/bias

tripy/tests/integration/test_layernorm.py
```diff
@@ -38,18 +47,43 @@ def test_layernorm_accuracy(self, torch_dtype, tp_dtype, input_shape, normalized
...
             eps=eps,
-            dtype=tp_dtype,
+            dtype=dtype,
         )
 
+        input = torch.empty(*input_shape, dtype=torch_dtype, device="cuda").uniform_(0, 10)
+        tp_input = tp.Tensor(input, dtype=dtype)
+
+        # Verify normalized output has approximately mean=0, std=1
+        tp_layernorm.weight = tp.ones(normalized_shape, dtype=dtype)
```
pranavm-nvidia, line 57: Here too, can parametrize over weight/bias.

tripy/nvtripy/frontend/module/layernorm.py
akhilg-nv: Ah I thought the rank check would cover this, but I realized there is a case where `normalized_shape` could be the full tensor shape, so I added it back.

tripy/tests/frontend/module/test_instancenorm.py
akhilg-nv: The stack trace isn't particularly clear, it says the error originates from the tp.ones -> tp.full call. For the static channel shape case we can catch it in the frontend but otherwise I'm not sure how to control where the error comes from in tensorrt.

pranavm-nvidia: Is the error thrown immediately when we add the `tensorrt.normalization` op? Because in that case we could just wrap it in a `try-except` and look for this type of error

akhilg-nv: discussed offline that the error occurs when compiling the graph, leaving this case as-is for now

pranavm-nvidia (approved):

PR #639, opened 2025-05-20
Adds an API to get the stream pointer

pranavm-nvidia:
Adds an API to the stream wrapper to get a pointer to the underlying CUDA stream. This is useful for interoperating with other CUDA libraries.


christopherbate (approved):
LGTM except needs some basic test. Thanks!

pranavm-nvidia:
> LGTM except needs some basic test. Thanks!

@christopherbate I've added a very basic test that checks if `stream.ptr` is an integer. If we also want to validate that it is pointing to a CUDA stream, then I'd probably need to use cuda-python to try and enqueue work on the stream (maybe just an allocation). Let me know if you want me to add something like that.

PR #644, opened 2025-05-21
Support None in slice for a new dimension

yizhuoz004:


tripy/nvtripy/frontend/ops/slice.py
```diff
@@ -105,23 +117,43 @@ def __getitem__(
...
+    new_dim_count = index.count(None)
     if ellipsis_count > 1:
         raise_error("Slicing index can only have a single ellipsis ('...')")
-    if len(index) > self.rank:
+    if len(index) > self.rank + new_dim_count:
         raise_error(f"Input tensor has a rank of {self.rank} but was attempted to be sliced with {len(index)} indices")
     if ellipsis_count:
         ellipsis_idx = index.index(Ellipsis)
-        num_slices = self.rank - len(index) + 1
+        num_slices = self.rank + new_dim_count - len(index) + 1
```
yizhuoz004, line 129: @pranavm-nvidia Some more details on AI coding's performance: auto pilot was trying to slice first and unsqueeze after, it was more complicated and buggy. Cursor switched to reshape first on the 2nd try, but it missed some details like this line. Also Cursor was able to modify the doc string and generate tests automatically.

pranavm-nvidia: Wouldn't it actually be easier to unsqueeze after? You could just remove all the `None`s and record their indices, and then do the slice as normal before unsqueezing.

yizhuoz004: unsqueeze only takes 1 dim at a time, so `reshape` is better as there might be multiple `None`s. Using `reshape` is easier before `slice` because the input shape is still unchanged. Maybe it's not complicated but AI struggled a lot.

tripy/tests/integration/test_slice.py
```diff
@@ -64,6 +65,10 @@ class TestSliceOp:
             ((2, 2), lambda t: t[..., 0]),
             ((2, 3, 4), lambda t: t[1, ...]),
             ((2, 3, 4, 4), lambda t: t[1, ..., 2]),
+            # None
+            ((2, 3, 4), lambda t: t[None]),
+            ((2, 3, 4), lambda t: t[None, :, :, None]),
```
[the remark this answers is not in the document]
yizhuoz004: Updated, thanks

tripy/nvtripy/frontend/ops/slice.py
pranavm-nvidia: Wouldn't it just be something like:
```py
shape = out.shape
for index in new_dims:
    shape.insert(index, 1)

out = reshape(out, shape)
```

yizhuoz004: Yes, `out.shape` is cleaner in the frontend but increases the IR. I was trying to avoid it by reusing the `sizes`.

pranavm-nvidia: Ah ok, I'm fine with whichever implementation you prefer.

pranavm-nvidia (approved):

tripy/nvtripy/frontend/ops/slice.py
```diff
@@ -105,23 +117,43 @@ def __getitem__(
...
 
     inp_shape = self.shape
+    if new_dim_count:
+        new_shape = []
+        slice_indices = []
+        current_dim = 0
+        for idx in index:
+            if idx is None:
+                new_shape.append(1)
+                slice_indices.append(slice(None))
```
pranavm-nvidia, line 140: Since we know the shape is `1` it might actually reduce the IR further to use `slice(0, 1)` instead of `slice(None)` which needs to infer bounds.

PR #646, opened 2025-05-23
Various improvements to SAMv2 demo

pranavm-nvidia:
Adds `--cap-add=SYS_PTRACE` to docker commands so that `gdb` can be used with child processes in the container.

Removes unnecessary stream synchronize in SAMv2 demo.

### DIFF
--- a/tripy/examples/segment-anything-model-v2/sam2/modeling/backbones/image_encoder.py
+++ b/tripy/examples/segment-anything-model-v2/sam2/modeling/backbones/image_encoder.py
@@ -56,7 +56,6 @@ class ImageEncoder(tp.Module):
         # Forward through backbone
         if self.compiled_executable:
             features_pos = list(self.compiled_executable(sample))
-            tp.default_stream().synchronize()
         else:
             features_pos = self.forward_impl(sample)
         for i in range(len(features_pos)):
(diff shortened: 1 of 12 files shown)


tripy/examples/segment-anything-model-v2/sam2/modeling/backbones/image_encoder.py
```diff
@@ -56,7 +56,6 @@ def __call__(self, sample: tp.Tensor):
         # Forward through backbone
         if self.compiled_executable:
             features_pos = list(self.compiled_executable(sample))
-            tp.default_stream().synchronize()
```
christopherbate, line 59: Why do you  think this is not required here? Are you synchronizing in another location?

We currently don't do anything in our `__dlpack__` method to ensure that the tensors are ready for the external consumer. We should be synchronizing there or inserting the proper CUDA event-based sychronizations to sychronize the MLIR-TRT runtime stream (which I'm guess what `tp.default_stream()`  points to) with the Torch default stream.

pranavm-nvidia: Good point. I've updated the demo to use the Tripy default stream for PyTorch operations as well. The synchronization happens implicitly when we call `(out_mask_logits[i] > 0.0).cpu()` at the end of each loop iteration.

pranavm-nvidia:
Pulled out the package index changes into a separate PR: https://github.com/NVIDIA/TensorRT-Incubator/pull/647

PR #649, opened 2025-05-30
Adds support for a dimension names attribute for block arguments

pranavm-nvidia:
Adds support for a dimension names attribute which is used to set dimension names in TensorRT. This conveys that the dimensions are equal at runtime.

### DIFF
--- a/mlir-tensorrt/tensorrt/lib/Target/TensorRTEncodingOpInterface/NetworkEncoder.cpp
+++ b/mlir-tensorrt/tensorrt/lib/Target/TensorRTEncodingOpInterface/NetworkEncoder.cpp
@@ -922,6 +922,34 @@ LogicalResult NvInferNetworkEncoder::encodeFunc(FunctionOpInterface func) {
       return failure();
     nvinfer1::ITensor *inputTensor =
         getNetworkDefinition()->addInput(name.c_str(), *dtype, trtShape);
+
+    // setDimensionName must be called immediately after addInput, or TensorRT
+    // will not deduplicate equal dimensions, which leads to perf gaps.
+    auto dimNamesAttr = func.getArgAttrOfType<DictionaryAttr>(
+        arg.getArgNumber(), TensorRTDialect::getDimensionNamesArgAttrName());
+    if (dimNamesAttr) {
+      for (NamedAttribute namedAttr : dimNamesAttr) {
+        int32_t key;
+        if (namedAttr.getName().getValue().getAsInteger(10, key))
+          return func->emitOpError()
+                 << "dimension name key '" << namedAttr.getName()
+                 << "' is not an integer";
+
+        if (key < 0 || key >= argType.getRank())
+          return func->emitOpError()
+                 << "dimension name key '" << key
+                 << "' is out of bounds for rank " << argType.getRank();
+
+        StringAttr strAttr = dyn_cast<StringAttr>(namedAttr.getValue());
+        if (!strAttr)
+          return func->emitOpError()
+                 << "dimension name value '" << namedAttr.getValue()
+                 << "' is not a string";
+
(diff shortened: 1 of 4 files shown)


shelkesagar29 (approved):
LGTM except few comments.

mlir-tensorrt/tensorrt/lib/Target/TensorRTEncodingOpInterface/NetworkEncoder.cpp
```diff
@@ -922,6 +922,29 @@ LogicalResult NvInferNetworkEncoder::encodeFunc(FunctionOpInterface func) {
...
+
+      for (NamedAttribute namedAttr : dimNamesAttr) {
+        int32_t key;
+        if (namedAttr.getName().getValue().getAsInteger(10, key)) {
+          return func->emitOpError()
+                 << "dimension name key '" << namedAttr.getName()
+                 << "' is not an integer";
+        }
+
+        if (StringAttr strAttr = namedAttr.getValue().dyn_cast<StringAttr>()) {
```
christopherbate, line 940: Failure to cast to `StringAttr` should be an error, no?

Also, we should be checking for `key` being out-of-bounds here.

Technically, the correct way to verify all these things are correct is to use  a Discardable Attribute Verifier (https://mlir.llvm.org/docs/DefiningDialects/#discardable-attribute-verification) on the TensorRT dialect. That would ensure any appearance of `tensorrt.dimension_names` could be verified to ensure it conforms to the correct requirements. But since we don't already have that for the other func arg/result attributes we have been associating with the TensorRT dialect, I don't think you have to introduce it in this MR. Until we have that verifier, however, we should do our best to emit errors for out-of-spec uses of the arg attr here.

mlir-tensorrt/tensorrt/lib/Target/TensorRTEncodingOpInterface/NetworkEncoder.cpp
```diff
@@ -922,6 +922,29 @@ LogicalResult NvInferNetworkEncoder::encodeFunc(FunctionOpInterface func) {
...
+          StringRef value = strAttr.getValue();
+          inputTensor->setDimensionName(static_cast<int32_t>(key),
```
christopherbate, line 942: `key` is already `int32_t`

mlir-tensorrt/tensorrt/lib/Target/TensorRTEncodingOpInterface/NetworkEncoder.cpp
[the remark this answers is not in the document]
pranavm-nvidia: Updated

pranavm-nvidia: Yes, changed it to emit an error and also added checks for the key being in-bounds.

mlir-tensorrt/tensorrt/lib/Target/TensorRTEncodingOpInterface/NetworkEncoder.cpp
```diff
@@ -27,6 +27,7 @@
 #include "mlir-tensorrt-dialect/TensorRT/Utils/Utils.h"
 #include "mlir-tensorrt-dialect/Utils/NvInferAdaptor.h"
 #include "mlir-tensorrt-dialect/Utils/StaticValueUtils.h"
+#include "mlir/IR/BuiltinAttributes.h"
```
shelkesagar29, line 30: excess header?

pranavm-nvidia: Thought it was needed for DictionaryAttr, but looks like it's not. Removed.

PR #650, opened 2025-06-04
Update LayerNorm2D to use TRT normalization API

akhilg-nv:
Naive performance test results indicate the new implementation is about 50% faster

```
LayerNorm2d_TP1 (Direct Implementation) took 0.0059s for 100 iterations
LayerNorm2d_TP2 (LayerNorm-based) took 0.0038s for 100 iterations
```

### DIFF
--- a/tripy/examples/segment-anything-model-v2/sam2/modeling/sam2_utils.py
+++ b/tripy/examples/segment-anything-model-v2/sam2/modeling/sam2_utils.py
@@ -180,26 +180,14 @@ class MLP(tp.Module):
         return x
 
 
-class LayerNorm2d(tp.Module):
+class LayerNorm2d(tp.LayerNorm):
     def __init__(self, num_channels: int, eps: float = 1e-6) -> None:
-        super().__init__()
-        from nvtripy.frontend.module.parameter import DefaultParameter
-
-        self.weight = DefaultParameter((num_channels,), tp.float32)
-        self.bias = DefaultParameter((num_channels,), tp.float32)
-        self.eps = eps
+        super().__init__(num_channels, dtype=tp.float32, eps=eps)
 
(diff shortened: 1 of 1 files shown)


tripy/examples/segment-anything-model-v2/sam2/modeling/sam2_utils.py
```diff
@@ -188,18 +188,14 @@ def __init__(self, num_channels: int, eps: float = 1e-6) -> None:
...
-        x = (x - u) / tp.sqrt(s + self.eps)
-        w = tp.unsqueeze(tp.unsqueeze(self.weight, 1), 2)
-        b = tp.unsqueeze(tp.unsqueeze(self.bias, 1), 2)
-        x = w * x + b
-        x = tp.cast(x, original_dtype)
-        return x
+        x_perm = tp.permute(x, (0, 2, 3, 1))
+        ln = tp.LayerNorm(self.num_channels, dtype=x.dtype, eps=self.eps)
+        ln.weight = self.weight
+        ln.bias = self.bias
```
pranavm-nvidia, line 197: I wonder if we should just have `LayerNorm2d` inherit from `tp.LayerNorm`? Something like this:
```py
class LayerNorm2d(tp.LayerNorm):
    def __init__(self, num_channels: int, eps: float = 1e-6) -> None:
        super().__init__(num_channels, dtype=tp.float32, eps=eps)

    def forward(self, x: tp.Tensor) -> tp.Tensor:
        x = tp.permute(x, (0, 2, 3, 1))
        x = super().forward(x)
        return tp.permute(x, (0, 3, 1, 2))
```

pranavm-nvidia: We might even be able to omit the constructor.

PR #668, opened 2025-07-03
Add text option to sphinx-build and update pyproject.toml dependencies

Mgluhovskoi:
Documentation building with text mode before this resulted in an error due to `CollapseNodes`. This changes should add support for `CollapseNodes` with text build option while not interfering with the html build option. Part of this change also updates some of the dependencies, particularly sphinx-build and sphinx-toolboox which had some bugs when building documentation with different build options.

### DIFF
--- a/tripy/docs/conf.py
+++ b/tripy/docs/conf.py
@@ -19,6 +19,7 @@ import os
 import sys
 
 ROOT_DIR = os.path.join(os.path.abspath(os.path.dirname(__file__)), os.path.pardir)
+import sphinx_toolbox
 sys.path.insert(0, ROOT_DIR)
 import inspect
 import re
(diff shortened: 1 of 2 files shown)


tripy/docs/conf.py
```diff
@@ -19,6 +19,7 @@
 import sys
 
 ROOT_DIR = os.path.join(os.path.abspath(os.path.dirname(__file__)), os.path.pardir)
+import sphinx_toolbox
```
pranavm-nvidia, line 22: Could we group this with the other imports?

tripy/pyproject.toml
```diff
@@ -45,38 +45,38 @@ doc_test_common = [
...
+  "sphinx-toolbox==4.0.0",
+  "docutils==0.21.2",
+  "myst-parser==3.0.1",
+  "sphinxcontrib-mermaid==1.0.0",
   "nvtripy[doc_test_common]",
   # Needed for guides:
-  "nvidia-modelopt==0.11.1",
-  "transformers==4.44.2",
-  "datasets==2.21.0",
+  "nvidia-modelopt==0.29.0",
```
pranavm-nvidia, line 57: Could you make sure L1 passes with this change?

PR #669, opened 2025-07-07
Updates MLIR to include layer metadata

pranavm-nvidia:
Updates our Trace->MLIR translation to include metadata for each layer that includes stack information and details about the Trace operation.

### DIFF
--- a/tripy/nvtripy/common/exception.py
+++ b/tripy/nvtripy/common/exception.py
@@ -15,8 +15,6 @@
 # limitations under the License.
 #
 
-import inspect
-from dataclasses import dataclass
 from textwrap import indent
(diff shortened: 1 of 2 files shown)


akhilg-nv:
Does fetching the source code affect the performance when compiling or running in eager mode?

yizhuoz004 (approved):

yizhuoz004:
> > Does fetching the source code affect the performance when compiling or running in eager mode?
> 
> @akhilg-nv yes, fetching source code is quite slow. It's also not readable at all because TRT/Nsight/NVTX (not sure which) doesn't seem to support newlines, so everything is just mashed together on a single line.

If this is the case, can we not enable this by default? e.g. have a flag related to perf debugging

PR #674, opened 2025-07-17
[TensorRT] Sets TRT layer metadata and nvtx profiling verbosity

yizhuoz004:
For now we query an environment variable `MTRT_TENSORRT_NVTX` to set the nvtx profiling verbosity. This is not ideal because it cannot support per-engine profiling verbosity. We will change that with a runtime option for TRT module.

### DIFF
--- a/mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
+++ b/mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
@@ -42,6 +42,23 @@
 using namespace mlirtrt;
 using namespace mlirtrt::runtime;
 
+static constexpr std::string_view kNvtxVerbosityEnvVariable =
+    "MTRT_TENSORRT_NVTX";
+
+/// Helper method that gets nvtx verbosity from environment value
+static nvinfer1::ProfilingVerbosity getNvtxVerbosity() {
+  const char *verbosity_str = std::getenv(kNvtxVerbosityEnvVariable.data());
+  if (!verbosity_str)
+    return nvinfer1::ProfilingVerbosity::kLAYER_NAMES_ONLY;
+  if (std::string_view(verbosity_str) == "NONE")
+    return nvinfer1::ProfilingVerbosity::kNONE;
+  if (std::string_view(verbosity_str) == "DETAILED")
+    return nvinfer1::ProfilingVerbosity::kDETAILED;
+  return nvinfer1::ProfilingVerbosity::kLAYER_NAMES_ONLY;
+}
+
+static const nvinfer1::ProfilingVerbosity gNvtxVerbosity = getNvtxVerbosity();
+
 namespace {
 /// A simple logger that implements TensorRT's logging interface. Errors and
 /// warnings are reported through TensorRT's diagnostic system, everything else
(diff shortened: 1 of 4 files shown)


mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -42,6 +42,22 @@
...
+  if (!verbosity_str)
+    return nvinfer1::ProfilingVerbosity::kLAYER_NAMES_ONLY;
+  if (std::string_view(verbosity_str) == "NONE")
+    return nvinfer1::ProfilingVerbosity::kNONE;
+  if (std::string_view(verbosity_str) == "DETAILED")
+    return nvinfer1::ProfilingVerbosity::kDETAILED;
+  return nvinfer1::ProfilingVerbosity::kLAYER_NAMES_ONLY;
+}
+
+static const nvinfer1::ProfilingVerbosity gNvtxVerbosity = getNvtxVerbosity();
```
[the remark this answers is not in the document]
yizhuoz004: Either way works for me as a WAR. I'm following the convensions in MLIR-TRT to get environment variables as a global static. @christopherbate do you have a preference here?

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -42,6 +42,22 @@
 using namespace mlirtrt;
 using namespace mlirtrt::runtime;
 
+static const char *kNvtxVerbosityEnvVariable = "MTRT_TENSORRT_NVTX";
+
+/// Helper method that gets nvtx verbosity from environment value
+nvinfer1::ProfilingVerbosity getNvtxVerbosity() {
```
christopherbate, line 48: Add `static` qualifier

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
```diff
@@ -42,6 +42,22 @@
 using namespace mlirtrt;
 using namespace mlirtrt::runtime;
 
+static const char *kNvtxVerbosityEnvVariable = "MTRT_TENSORRT_NVTX";
```
christopherbate, line 45: Use `static constexpr std::string_view` or `static constexpr llvm::StringRef`

mlir-tensorrt/executor/lib/Runtime/Backend/Lua/Modules/TensorRT/TensorRTModule.cpp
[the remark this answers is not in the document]
christopherbate: Either is fine with me for now if you have an actual use-case that is needed. I don't think we want to make a pattern of mutating the state of the runtime modules after load, however.

mlir-tensorrt/tensorrt/lib/Target/TensorRTEncodingOpInterface/NetworkEncoder.cpp
```diff
@@ -278,6 +278,10 @@ void NvInferNetworkEncoder::setMetadata(nvinfer1::ILayer *layer,
                                         Operation *sourceOp) {
   std::string name = createName(namesSet, sourceOp);
   layer->setName(name.c_str());
+
+  if (auto metadataAttr = sourceOp->getAttrOfType<StringAttr>("metadata")) {
```
christopherbate, line 282: This string `"metadata"` should be defined somewhere stable. Usually we put such string definitions in the dialect class as a constexpr stringref.

christopherbate (approved):

mlir-tensorrt/tensorrt/lib/Target/TensorRTEncodingOpInterface/NetworkEncoder.cpp
yizhuoz004: Updated, thanks!

PR #693, opened 2025-08-12
Allow DimensionSize input to conv params

akhilg-nv:


tripy/nvtripy/frontend/module/conv/base.py
```diff
@@ -39,12 +40,12 @@ class ConvBase(Module):
 
     def __init__(
         self,
-        in_channels: int,
-        out_channels: int,
-        kernel_dims: Sequence[int],
+        in_channels: Union[int, DimensionSize],
```
pranavm-nvidia, line 43: How would this work? TRT expects the channel dimension to be known at build-time.

yizhuoz004: I think it has to be a known constant because the in_channels decides the weight's shape.

